PR #7, opened 2022-04-08
Resource control based on cgroup

SongGuyang:
See the doc for more details.


reps/2022-04-01-resource-control.md
```diff
@@ -0,0 +1,167 @@
...
+@ericl
+
+## Design and Architecture
+
+### Cluster level API
+We should add some new system configs for the resource control.
+- worker_resource_control_method: Set to "cgroup" by default.
+- cgroup_manager: Set to "cgroupfs" by default. We should also support `systemd`.
+- cgroup_mount_path: Set to `/sys/fs/cgroup/` by default.
+- cgroup_unified_hierarchy: Whether to use cgroup v2 or not. Set to `False` by default.
```
simon-mo, line 28: can we only support cgroup v2 to reduce the scope?

reps/2022-04-01-resource-control.md
```diff
@@ -0,0 +1,167 @@
...
+@ericl, @edoakes, @simon-mo, @chenk008 @raulchen 
+
+### Shepherd of the Proposal (should be a senior committer)
+@ericl
+
+## Design and Architecture
+
+### Cluster level API
+We should add some new system configs for the resource control.
+- worker_resource_control_method: Set to "cgroup" by default.
```
simon-mo, line 25: cgroup is the only method at the moment right?

reps/2022-04-01-resource-control.md
```diff
@@ -0,0 +1,167 @@
...
+mkdir /sys/fs/cgroup/{worker_id}
+echo "200000 1000000" > /sys/fs/cgroup/foo/cpu.max
+echo {pid} > /sys/fs/cgroup/foo/cgroup.procs
+```
+NOTE: This is an example based on cgroup v2. The commmand lines in cgroup v1 is defferent and incompatible.
+
+We also could use the [libcgroup](https://github.com/libcgroup/libcgroup/blob/main/README) to simplify the implementation. This library support both cgroup v1 and cgroup v2. 
+
+##### systemd
+Using systemd to manage cgroups is like:
```
simon-mo, line 101: what's the advantage of using systemd here?

reps/2022-04-01-resource-control.md
```diff
@@ -0,0 +1,167 @@
...
+And you can try to enable cgroup v2:
+```
+grubby --update-kernel=ALL --args="systemd.unified_cgroup_hierarchy=1"
+reboot
+```
+
+#### How to delete cgroups
+When worker processes die, we should delete the cgroup which is created for the processes. This work is only needed for cgroupfs because systemd could handle it.
+
+So, we should add a new RPC named `CleanForDeadWorker` in `RuntimeEnvService`. The Raylet should send this PRC to the Agent and the Agent will delete the cgroup.
```
simon-mo, line 129: can this be part of RuntimeEnv's refcounting and GC RPCs?

ericl:
@scv119 want to shepherd this one?

reps/2022-04-01-resource-control.md
```diff
@@ -0,0 +1,167 @@
...
+- cgroup_manager: Set to "cgroupfs" by default. We should also support `systemd`.
+- cgroup_mount_path: Set to `/sys/fs/cgroup/` by default.
+- cgroup_unified_hierarchy: Whether to use cgroup v2 or not. Set to `False` by default.
+- cgroup_use_cpuset: Whether to use cpuset. Set to `False` by default.
+
+### User level API
+#### Simple API using a flag
+```python
+    runtime_env = {
+        "enable_resource_control": True,
```
ericl, line 35: Is there any reason this cannot be implemented using runtime_env plugins? We should consider whether to extend the plugin mechanism to support this, if this is not possible, rather than special casing cgroups.

The goal of runtime env plugins is exactly to enable supporting these kind of complex use cases in a customizable way.

reps/2022-04-01-resource-control.md
```diff
@@ -0,0 +1,167 @@
...
+
+Currently, we don't control and limit the resource usage of worker processes, except running the worker in a container (see the `container` part in the [doc](https://docs.ray.io/en/latest/ray-core/handling-dependencies.html#api-reference)). In most of the scenarios, the container is unnecessary, but resource control is necessary for isolation.
+
+[Control groups](https://man7.org/linux/man-pages/man7/cgroups.7.html), usually referred to as cgroups, are a Linux kernel feature which allow processes to be organized into hierarchical groups whose usage of various types of resources can then be limited and monitored.
+
+So, the goal of this proposal is to achieve resource control for worker processes by cgroup in Linux.
+
+### Should this change be within `ray` or outside?
+
+These changes would be within Ray core.
```
ericl, line 12: I actually think this is a prime candidate for being a third-party runtime_env plugin (which we can either keep in a separate repo, or as a "contrib" module to Ray.

reps/2022-04-01-resource-control.md
SongGuyang: I think we should also support cgroup v1 because it has been used widely and written into standard. I have described here https://github.com/ray-project/enhancements/pull/7/files#diff-98ccfc9582e95581aae234797bc273b2fb68cb9e4dcc3030c8e94ba447daef7dR112-R113.
And in Ant, we only support v1 in production. Can you check your production environments and CI environments?

reps/2022-04-01-resource-control.md
SongGuyang: Right. But we can also support others, e.g. a custom monitor which could kill worker processes when the resources exceeded.

reps/2022-04-01-resource-control.md
SongGuyang: If we use systemfs in a systemd based system,  there will be more than one component which manage the cgroup tree. And from https://www.freedesktop.org/wiki/Software/systemd/ControlGroupInterface/, we could know that:

*In the short-term future writing directly to the control group tree from applications should still be OK, as long as the [Pax Control Groups](http://www.freedesktop.org/wiki/Software/systemd/PaxControlGroups/) document is followed. In the medium-term future it will still be supported to alter/read individual attributes of cgroups directly, but no longer to create/delete cgroups without using the systemd API. In the longer-term future altering/reading attributes will also be unavailable to userspace applications, unless done via systemd's APIs (either D-Bus based IPC APIs or shared library APIs for passive operations).*

In addition, systemd is highly recommended by `runc` for cgroup v2 https://github.com/opencontainers/runc/blob/main/docs/cgroup-v2.md.

reps/2022-04-01-resource-control.md
SongGuyang: Good question. I have thought about this before. Currently, the URIs could be shared and reused between different worker processes. But for cgroups, it could only be used for one process and it is one-off. So, the existing logic couldn't be used directly.

reps/2022-04-01-resource-control.md
raulchen: In the current runtime env plugin mechanism design, a plugin must implement a whole runtime env type (e.g., pip, conda, etc).
Cgroup is different here. Because it is more like a config of runtime env, rather than a new runtime env type. Because it can be used together with other types. 
So in order to pluginize cgroup, we may need a new plugin mechanism that can extend the the current plugin mechanism. This seems weird. and I'm not sure if such a new plugin mechanism can be re-used for other features. 
So I think maybe we can consider pluginizing it later if we have a second similar requirement.

SongGuyang: How to achieve the logic of deleting cgroups in plugin?

reps/2022-04-01-resource-control.md
```diff
@@ -0,0 +1,199 @@
...
+The cgroupfs and the systemd are two mainstream ways to manage cgroups. In [container technology](https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver/), this two ways are also supported. We should also support both of them.
+
+The reason of using cgroupfs:
+- Cgroupfs is a traditional and common way to manage cgroups. If your login user has the privilege of cgroup file system, you can create the cgroups.
+- In some environments, you couldn't use systemd, e.g. in a non-systemd based system or in a container which couldn't access the systemd service.
+
+The reason of using systemd:
+- Systemd is highly recommended by [runc](https://github.com/opencontainers/runc/blob/main/docs/cgroup-v2.md.) for cgroup v2.
+- If we use cgroupfs in a systemd based system, there will be more than one component which manage the cgroup tree.
+- In the systemd [docs](https://www.freedesktop.org/wiki/Software/systemd/ControlGroupInterface/), we can know that  create/delete cgroups will be unavailable to userspace applications, unless done via systemd's APIs.
```
raulchen, line 115: what does this mean? what problem will occur if we use cgroupfs?

reps/2022-04-01-resource-control.md
```diff
@@ -0,0 +1,199 @@
...
+- The systemd has a good abstract API of cgroups. 
+
```
raulchen, line 117: So the conclusion is that, we support both cgroupfs and systemd. And systemd will be used when it is available, otherwise we fallback to cgroupfs. Right? can you also add the conclusion to the doc?

reps/2022-04-01-resource-control.md
```diff
@@ -0,0 +1,199 @@
...
+### Support using cgroups in Ray
```
raulchen, line 118:
```suggestion
### Changes needed in Ray
```

reps/2022-04-01-resource-control.md
```diff
@@ -0,0 +1,199 @@
...
+
+1. Cgroup manager in the Agent
+
+The Cgroup Manager is used to create or delete cgroups, and bind worker processes to cgroups. We plan to integrate the Cgroup Manager in the Agent. 
+
+We should abstract the Cgroup Manager because there are more than one way to manager cgroups in linux: cgroupfs and systemd.
```
raulchen, line 124:
```suggestion
Cgroup Manager should expose an abstract interface that can hide the differences between cgroupfs/systemd and cgroup v1/v2.
```

reps/2022-04-01-resource-control.md
```diff
@@ -0,0 +1,199 @@
...
+#### Why we should support both cgroupfs and systemd
+The cgroupfs and the systemd are two mainstream ways to manage cgroups. In [container technology](https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver/), this two ways are also supported. We should also support both of them.
+
+The reason of using cgroupfs:
+- Cgroupfs is a traditional and common way to manage cgroups. If your login user has the privilege of cgroup file system, you can create the cgroups.
+- In some environments, you couldn't use systemd, e.g. in a non-systemd based system or in a container which couldn't access the systemd service.
+
+The reason of using systemd:
+- Systemd is highly recommended by [runc](https://github.com/opencontainers/runc/blob/main/docs/cgroup-v2.md.) for cgroup v2.
+- If we use cgroupfs in a systemd based system, there will be more than one component which manage the cgroup tree.
```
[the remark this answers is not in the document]
SongGuyang: I think so in any cases.

reps/2022-04-01-resource-control.md
SongGuyang: This is a planing of systemd. It will occur in future.

reps/2022-04-01-resource-control.md
SongGuyang: Maybe we should make it to be a config?

reps/2022-04-01-resource-control.md
SongGuyang: Sorry, let me explain what I said above.

with the current plugin mechanism, it's hard to implement cgroup deletion. As I said here https://github.com/ray-project/enhancements/pull/7#discussion_r847002806, the cgroups are one-off. It couldn't be shared and reused. So, the logic should be that: Create cgroup -> start worker -> worker dies -> delete cgroup. For example, if the cgroup dir is `/sys/fs/cgroup/{worker_id}`, we can't reuse the reference counting logic to delete this dir because we don't provide the `worker_id` when we decrease the reference by the PRC from raylet to agent.

SongGuyang:
@ericl @simon-mo We've refined the structure of the whole doc.

ericl:
> In the current runtime env plugin mechanism design, a plugin must implement a whole runtime env type (e.g., pip, conda, etc).  Cgroup is different here. Because it is more like a config of runtime env, rather than a new runtime env type. Because it can be used together with other types.  So in order to pluginize cgroup, we may need a new plugin mechanism that can extend the the current plugin mechanism. This seems weird. and I'm not sure if such a new plugin mechanism can be re-used for other features.  So I think maybe we can consider pluginizing it later if we have a second similar requirement.

I'm not sure I see this as weird. This was one of the original use cases we envisioned for runtime_env plugins when we designed it (arbitrary modifications to the environment, such as attaching debuggers, resource limits, etc.).

> with the current plugin mechanism, it's hard to implement cgroup deletion. As I said here #7 (comment), the cgroups are one-off. It couldn't be shared and reused. So, the logic should be that: Create cgroup -> start worker -> worker dies -> delete cgroup. For example, if the cgroup dir is /sys/fs/cgroup/{worker_id}, we can't reuse the reference counting logic to delete this dir because we don't provide the worker_id when we decrease the reference by the PRC from raylet to agent.

This doesn't sound like a fundamental issue, but something we can add right? Here's my concern: we're adding a ton of stuff into core, which will not be used by most users. If we can do this with the existing plugin infra, with some minor improvements, then longer term that leads to a better architecture than handling it directly in core. I can understand it makes this project a little more involved, but from the overall Ray project perspective this is cleaner long term as we might have other types of plugins of this sort.

SongGuyang:
@ericl I agree that the pluggable architecture could make the core cleaner in long term. And I could try to make it to be a `resource_control` plugin. But I think the current abstraction of runtime env plugin is not clear and couldn't meet my requirements. We should redefine it first. Not sure if we need to create a separate REP for this. And we need to do more discussion about this with @edoakes and @architkulkarni.

btw, I don't think the resource control is a minor improvement. Maybe it will be a basic ability of Ray Core and widely used by users in large scale cluster in future. We already have a lot issues in Ant and we plan to put it into production in few month.

ericl:
I'd like to see the discussion outcome then. If @edoakes and @architkulkarni think it's not a good fit for runtime env plugin, then I'm fine with that decision.

reps/2022-04-01-resource-control.md
simon-mo: Can you clarify the design here? Is it fallback based or config based? If config based, what is the documentation for user here?

reps/2022-04-01-resource-control.md
```diff
@@ -0,0 +1,202 @@
...
+
+#### Entire APIs
+```python
+    runtime_env = {
+        "enable_resource_control": True,
+        "resource_control_config": {
+            "cpu_enabled": True,
+            "memory_enabled": True,
+            "cpu_strict_usage": False,
+            "memory_strict_usage": True,
```
architkulkarni, line 55: What does each of these options do?  Where would the memory limit and CPU limit numbers be specified in this API?  Sorry if these are well-known options, I'm still learning about cgroups.

architkulkarni:
> @ericl I agree that the pluggable architecture could make the core cleaner in long term. And I could try to make it to be a `resource_control` plugin. But I think the current abstraction of runtime env plugin is not clear and couldn't meet my requirements. We should redefine it first. Not sure if we need to create a separate REP for this. And we need to do more discussion about this with @edoakes and @architkulkarni.

The issue with deletion makes sense, I agree it seems tricky to support it with the current abstraction.  What would the ideal abstraction look like?  One naive idea is to have each runtime env plugin have a method `OnWorkerDead(worker_id: int)` which is called whenever the worker dies, but that might be too specific to this use case.

reps/2022-04-01-resource-control.md
SongGuyang: @simon-mo Yep, I will add this details.

reps/2022-04-01-resource-control.md
SongGuyang: I will add some description for each of the options. We already support setting CPU number in the remote call option, for example:
```
@ray.remote(num_cpus=1)
class actor:
    ...
```

SongGuyang:
> The issue with deletion makes sense, I agree it seems tricky to support it with the current abstraction. What would the ideal abstraction look like? One naive idea is to have each runtime env plugin have a method `OnWorkerDead(worker_id: int)` which is called whenever the worker dies, but that might be too specific to this use case.

Yep, we should add a method like what you said. And we should do more things to make it ideal. For example, we need to support configuration for each plugin, because I will add a lot cluster level config here https://github.com/ray-project/enhancements/pull/7/files#diff-98ccfc9582e95581aae234797bc273b2fb68cb9e4dcc3030c8e94ba447daef7dR24.

@architkulkarni btw, do we plan to make current runtime env modules, like pip and conda, to be plugins? Or we will keep it in long term?

Another thing is what @raulchen said here https://github.com/ray-project/enhancements/pull/7#discussion_r847047753. What do you think about this? In addition, do we need to support applying multiple plugins? If we do, how can we guarantee the sort and avoid the conflicts between plugins? We should consider any more for those issues in the abstraction of plugin, right?

ericl:
I made some edits to https://docs.google.com/document/d/1x1JAHg7c0ewcOYwhhclbuW0B0UC7l92WFkF4Su0T-dk/edit# that I think will enable cgroups to be supported (namely, adding priority, create_for_worker, delete_for_worker). Cgroups might be implementable out of the box as a plugin after these hooks are added, wdyt?

SongGuyang:
> I made some edits to https://docs.google.com/document/d/1x1JAHg7c0ewcOYwhhclbuW0B0UC7l92WFkF4Su0T-dk/edit# that I think will enable cgroups to be supported (namely, adding priority, create_for_worker, delete_for_worker). Cgroups might be implementable out of the box as a plugin after these hooks are added, wdyt?

I got your thought. But I think this doc is still not clear.
- How these hooks cooperate with the Raylet and the Agent? How to refine the RPC ?
- How the plugin affects the runtime env cache? Currently, if the runtime env json is same, we will return the cached runtime env context. But in the case of cgroup, maybe the context is different?
- How to implement configuration of each plugin? 

btw, @ericl @architkulkarni I think we should also make the google doc to be a REP, wdyt? Who can take this work? I'm also available for this if you don't have bandwidth.

raulchen:
I think we are on the same page that resource control is better implemented as a plugin. Also, it seems that the plugin system still needs an overhaul to meet different kinds of runtime env requirements. 
So my suggestion is that we can do 2 things simultaneously:
1) summarize the minimal needed changes of the plugin system to support cgroup. And make these changes as part of the this REP. 
2) discuss what other changes are needed to support other plugins, and potentially file another REP for the plugin system refinements. 

What do you think? If we agree, @SongGuyang can add the first part to this REP first.

ericl:
+1

edoakes:
@raulchen +1. Please keep @architkulkarni in the loop on the plugin changes, he can also help scope out the design offline first.

reps/2022-04-01-resource-control.md
rkooo567: Why not just making URI not shareable? I believe it should be easy to implement? (e.g., add worker id to each group URI key? )

reps/2022-04-01-resource-control.md
```diff
@@ -0,0 +1,202 @@
...
+    runtime_env = ray.runtime_env.RuntimeEnv(
+        enable_resource_control=True
+    )
+```
+
+#### Entire APIs
+```python
+    runtime_env = {
+        "enable_resource_control": True,
+        "resource_control_config": {
```
rkooo567, line 51: Is cgroup version automatically detected?

SongGuyang:
I will continue to work on current REP this week and back to discuss with you about the left comments.

reps/2022-04-01-resource-control.md
```diff
@@ -0,0 +1,202 @@
...
+
+Currently, we don't control and limit the resource usage of worker processes, except running the worker in a container (see the `container` part in the [doc](https://docs.ray.io/en/latest/ray-core/handling-dependencies.html#api-reference)). In most of the scenarios, the container is unnecessary, but resource control is necessary for isolation.
+
+[Control groups](https://man7.org/linux/man-pages/man7/cgroups.7.html), usually referred to as cgroups, are a Linux kernel feature which allow processes to be organized into hierarchical groups whose usage of various types of resources can then be limited and monitored.
+
+So, the goal of this proposal is to achieve resource control for worker processes by cgroup in Linux.
+
+### Should this change be within `ray` or outside?
+
+These changes would be within Ray Core.
```
ericl, line 12: The changes should no longer be within Ray core, but scoped to `ray.experimental.cgroup_env_plugin` (moved to util eventually).

reps/2022-04-01-resource-control.md
```diff
@@ -0,0 +1,202 @@
...
+### Required Reviewers
+@ericl, @edoakes, @simon-mo, @chenk008 @raulchen 
+
+### Shepherd of the Proposal (should be a senior committer)
+@ericl
+
+## Design and Architecture
+
+### Cluster level API
+We should add some new system configs for the resource control.
```
ericl, line 24: There should be no system level configs for a runtime_env plugin.

reps/2022-04-01-resource-control.md
```diff
@@ -0,0 +1,202 @@
...
+            "cpu_enabled": True,
+            "memory_enabled": True,
+            "cpu_strict_usage": False,
+            "memory_strict_usage": True,
+        }
+    }
+```
+
+```python
+    from ray.runtime_env import RuntimeEnv, ResourceControlConfig
```
ericl, line 61: The config example should be rewritten to show what it looks like as a runtime_env plugin.

reps/2022-04-01-resource-control.md
```diff
@@ -0,0 +1,202 @@
...
+
+The reason of using systemd:
+- Systemd is highly recommended by [runc](https://github.com/opencontainers/runc/blob/main/docs/cgroup-v2.md.) for cgroup v2.
+- If we use cgroupfs in a systemd based system, there will be more than one components which simultaneously manage the cgroup tree.
+- In the systemd [docs](https://www.freedesktop.org/wiki/Software/systemd/ControlGroupInterface/), we can know that, in future, create/delete cgroups will be unavailable to userspace applications, unless done via systemd's APIs.
+- The systemd has a good abstract API of cgroups. 
+
+So, we should support both cgroupfs and systemd. And we can provide a config which is used to change the cgroup manager.
+
+### Changes needed in Ray
```
ericl, line 121: Please rewrite this to reflect the plugin architecture.

reps/2022-04-01-resource-control.md
SongGuyang: I found a util function in runc https://github.com/opencontainers/runc/blob/main/libcontainer/cgroups/utils.go#L34-L50. But cgroup v1 and v2 could both be enabled in some systems. And it also depends on which sub systems(e.g. cpu, memory) has been enable in the cgroup. So, I'd like to add a config in first step. We can also change this in future.

SongGuyang:
@ericl @architkulkarni @simon-mo @rkooo567 Take a look again?

reps/2022-04-01-resource-control.md
```diff
@@ -0,0 +1,198 @@
...
+When worker processes die, we should delete the cgroup which is created for the processes. This work is only needed for cgroupfs because systemd can automatically delete the cgroup when the process dies.
+
+So, we should make some changes to allow each runtime environment plugin to do some work when worker processes die.
+- Add `worker_id` field to the existing PRC request `GetOrCreateRuntimeEnvRequest` and `DeleteRuntimeEnvIfPossibleRequest`. These two RPCs are used to create and delete runtime environments from the Raylet to the Agent.
+- Add two methods, named `create_for_worker` and `delete_for_worker`, in the plugin interface. We already have two methods named `create` and `delete` in the interface. The difference is that, `delete` is called only when there is no worker using the runtime environment(The reference counting to be zero), but `delete_for_worker` is called every time when a process dies.
+
+Another change is that we should support loading runtime env plugins when we start the Ray nodes. For example:
+```
+ray start --head --runtime-env-plugins="ray.experimental.cgroup_env_plugin:ray.experimental.test_env_plugin"
+```
```
architkulkarni, line 155: Is this different from `eager_install`, which already exists in Ray?

SongGuyang: Two things absolutely. This option is used to register plugins into the dashboard agent. For example, you can register a plugin named "conda" when start the ray node. So, when a job uses "conda" field  in `runtime_env` after the ray cluster starts, agent could call the "conda"  plugin directly to set up the environment.

architkulkarni: Makes sense!  In the [pluggability design](https://docs.google.com/document/d/1x1JAHg7c0ewcOYwhhclbuW0B0UC7l92WFkF4Su0T-dk/edit?usp=sharing) there's a related idea which is to use environment variables, e.g. `RAY_RUNTIME_ENV_PLUGIN_CGROUP=ray.experimental.test_cgroup_plugin ray start --head`.  These both seem straightforward to implement but I'm not sure which approach is the best, do you have any thoughts about this?

reps/2022-04-01-resource-control.md
```diff
@@ -0,0 +1,198 @@
...
+### Should this change be within `ray` or outside?
+
+The resource control ability will be implemented in a runtime environments plugin. So the changes should no longer be within Ray core, but scoped to ray.experimental.cgroup_env_plugin (moved to util eventually).
+
+## Stewardship
+### Required Reviewers
+@ericl, @edoakes, @simon-mo, @chenk008 @raulchen 
+
+### Shepherd of the Proposal (should be a senior committer)
+@ericl
```
edoakes, line 19:
```suggestion
@edoakes
```

edoakes (commented):
Hey @SongGuyang we did a small group review of this and this was the conclusion/feedback:

1. The requisite changes to runtime_env plugin architecture make sense and would be an overall improvement.
2. We would prefer that the functionality does not live in the main Ray repo to begin with, but we can reevaluate this later as it matures and we get more user feedback.
3. We aren't quite sure what the gap between this and the container field in runtime_env is or how we should be thinking about them together. Should we support both? Can we support the cgroup functionality by extending container support? If we are providing cgroup, should we remove container support? Additionally, the existing container support has not been publicly documented/adopted nor maintained. Given these concerns, we think it would sense to evaluate both the container and cgroup options holistically as part of this REP -- a strawman proposal would be to move the existing container support out of the codebase and into a plugin as well if we think both are needed.
4. Could you please add more details about how this will be tested to the REP? Specifically, where will the tests live, what types of unit & integration tests we will need, etc.? Especially if this is not living in the main Ray repo we will need to make sure that the testing infrastructure is set up to avoid breakages.

Also, @ericl asked me to shepherd this change given that I have more context on runtime_env, so could you please update the text accordingly?

Please let me know if you have any questions!

SongGuyang:
> 2. We would prefer that the functionality does not live in the main Ray repo to begin with, but we can reevaluate this later as it matures and we get more user feedback.

Why it couldn't to be an experimental module in the main Ray repo? I didn't get a strong reason for this in your description.

SongGuyang:
> 3. We aren't quite sure what the gap between this and the container field in runtime_env is or how we should be thinking about them together. Should we support both? Can we support the cgroup functionality by extending container support? If we are providing cgroup, should we remove container support? Additionally, the existing container support has not been publicly documented/adopted nor maintained. Given these concerns, we think it would sense to evaluate both the container and cgroup options holistically as part of this REP -- a strawman proposal would be to move the existing container support out of the codebase and into a plugin as well if we think both are needed.

I think we need both. 

Cgroup is only used to control the physics resources, e.g. cpus and memory, but the container is used for a lot of purposes, e.g. custom linux kernel, custom system packages(yum install ...), custom file system, and resource control certainly. We cannot use container instead of cgroup in some scenes because cgroup is more lightweight which doesn't rely on image and image center. When you create a directory in the file system, a cgroup is obtained.

Can we only add some description like what I say above? Container is another big issue. I ensure that it could also be a plugin. But, please don't let it block the pluggability enhancement and cgroup. We can discuss it in a separate issue or REP.

raulchen:
@edoakes here are my thoughts to your questions.

1. Why need both cgroup and container? 

As @SongGuyang mentioned above, sometimes users only want resource control + pip requirements. In this case, containers are too heavy-weighted, as they would have to manage their docker files and images. 

2. Where should the code live?

I'd prefer to putting it in the main repo (at least to begin with). For one, I think the resource control requirement is as common as other existing ones. Putting it in the main repo would benefit more users. For another, we need to keep refining the plugin system along with this feature. Putting everything in the same repo would facilitate testing. 

I think we can put it in a new "runtime_env_plugins" directory, and later migrate existing runtime env to this directory as well. The core shouldn't depend on anything in this directory. Also, runtime envs can be shipped separately as well. E.g., "pip install ray[runtime_env_pip]" "pip install ray[runtime_env_container]".

edoakes:
@SongGuyang @raulchen

Re: cgroups + containers, I understand that you cannot use cgroup as a replacement for containers, but containers are essentially a superset of cgroup, so could we leverage the existing container support for this purpose as well _without_ requiring users to specify an image or worry about other container details? For example, make the image and cgroup options both configurable in the `container` command and the image is optional.

Re: putting it in the repo vs. outside, the reasoning is we do not yet have evidence of user demand for this feature and adding it to the core repo increases maintenance burden. A relevant example is the container support in runtime_env: this is not something used by the community because it's not polished or well-documented, but it has increased maintenance burden for runtime_env and caused test failures without anyone taking clear ownership ([the relevant test is now disabled](https://github.com/ray-project/ray/commit/3e7511e84f5f8a4705ee1e9eab1d0b2f369a188f)). We want to ensure the same thing doesn't happen w/ cgroup support. Once the feature is fully-fledged and we see user demand, we can reconsider promoting it to put it in the repo.

edoakes:

PR #9, opened 2022-04-25
[Ray enhancement] defer rep4 for future consideration.

scv119:
Hey @septicmk. Thanks for submitting the rep and contributing to the Ray community. After voting in Ray committers, we concluded that we **won't accept this REP this time**. Concretely:

1. The ray committers are generally in favor of the direction of having a clean interface for plasma.
2. The main concern is the feasibility and maintainability of the proposed interface; this requires a fully working prototype to validate. 
3. Also, there are questions about if we have fully explored the design spaces and if there exists a better alternative.

Again, thanks for putting up the efforts to come up with the REP. As we discussed offline, Vineyard and Ray team will collaborate in the following mode:
1. Ray will not support an alternative plasma store at the moment, and there is no guarantee that the interface won't change.
2. Instead of proposed REP, Vineyard team will prototype and maintain the integration code as addon/plug-in/patch in one of their repos. As they prototype, they might work with Ray team and submit PRs to clean up Ray's plasma interface that brings net benefit to Ray.
3. If the community accepts the API changes, and a net benefit to Ray can be demonstrated, and the introduction of Vineyard does not complicate the Ray development process, we may reconsider the plasma alternative REP.


scv119:
Original REP https://github.com/ray-project/enhancements/pull/4

PR #13, opened 2022-09-02
Add labels mechanism and actor-affinity-feature proposal

larrylian:
# 1. Labels mechanism.
Introduce the Labels mechanism. Give Labels to Actors/Nodes. 
Affinity features such as ActorAffinity/NodeAffinity can be realized through Labels.

Read reps/2022-11-23-labels-mechanish-and-affinity-schedule-feature.md first.

# 2.  Actor-Affinity Feature
Provides a set of lightweight actor affinity and anti-affinity scheduling interfaces.
Replacing the heavy PG interface to implement collocate or spread actors.

* Affinity
  * Co-locate the actors in the same node.
  * Co-locate the actors in the same batch of nodes, like nodes in the same zones
* Anti-affinity
  * Spread the actors of a service across nodes and/or availability zones, e.g. to reduce correlated failures.
  * Give a actor "exclusive" access to a node to guarantee resource isolation 
  * Spread the actors of different services that will affect each other on different nodes.

ActorAffinity调度

### DIFF
--- a/reps/2022-08-31-actor-affinity-apis.md
+++ b/reps/2022-08-31-actor-affinity-apis.md
@@ -36,8 +36,8 @@ Yes, this will be a complement to ray core's ability to flexibly schedule actors
 
 Actor affinity/anti-affinity schedule API Design
 1. Scheduling Strategy adds an ActorAffinitySchedulingStrategy. 
-2. This strategy consists of several ActorAffinityMatchExpressions.
-3. ActorAffinityMatchExpression has 4 match types which are IN/NOT_IN/EXISTS/DOES_NOT_EXIST
+2. This strategy consists of several LabelMatchExpressions.
+3. LabelMatchExpression has 4 match types which are IN/NOT_IN/EXISTS/DOES_NOT_EXIST
 
 Use Case | ActorAffinityOperator
 -- | --
@@ -72,10 +72,10 @@ SchedulingStrategyT = Union[None, str,
                             ActorAffinitySchedulingStrategy]
 
 class ActorAffinitySchedulingStrategy:
-    def __init__(self, match_expressions: List[ActorAffinityMatchExpression]):
+    def __init__(self, match_expressions: List[LabelMatchExpression]):
         self.match_expressions = match_expressions
 
-class ActorAffinityMatchExpression:
+class LabelMatchExpression:
     """An expression used to represent an actor's affinity.
     Attributes:
         key: the key of label
@@ -84,7 +84,7 @@ class ActorAffinityMatchExpression:
         values: a list of label value
         soft: ...
     """
-    def __init__(self, key: str, operator: ActorAffinityOperator,
+    def __init__(self, key: str, operator: LabelMatchOperator,
                  values: List[str], soft: bool):
         self.key = key
         self.operator = operator
@@ -106,7 +106,7 @@ Step 2: Set actor affinity strategy.
 1. The target actor is expected to be scheduled with the actors whose label key is "location" and value in ["dc-1"].
 ```python
 match_expressions = [
-    ActorAffinityMatchExpression("location", ActorAffinityOperator.IN, ["dc_1"], False)
+    LabelMatchExpression("location", LabelMatchOperator.IN, ["dc_1"], False)
 ]
 actor_affinity_strategy = ActorAffinitySchedulingStrategy(match_expressions)
 actor = Actor.options(scheduling_strategy = actor_affinity_strategy).remote()
@@ -116,7 +116,7 @@ actor = Actor.options(scheduling_strategy = actor_affinity_strategy).remote()
 with the actors whose label key is "location" and value in ["dc-1"].
 ```python
 match_expressions = [
- ActorAffinityMatchExpression("location", ActorAffinityOperator.NOT_IN, ["dc_1"], False)
+ LabelMatchExpression("location", LabelMatchOperator.NOT_IN, ["dc_1"], False)
 ]
 actor_affinity_strategy = ActorAffinitySchedulingStrategy(match_expressions)
 actor = Actor.options(scheduling_strategy = actor_affinity_strategy).remote()
@@ -125,7 +125,7 @@ actor = Actor.options(scheduling_strategy = actor_affinity_strategy).remote()
 3. The target actor is expected to be scheduled with the actors whose label key exists "location".
 ```python
 match_expressions = [
- ActorAffinityMatchExpression("location", ActorAffinityOperator.EXISTS, [], False)
+ LabelMatchExpression("location", LabelMatchOperator.EXISTS, [], False)
 ]
 actor_affinity_strategy = ActorAffinitySchedulingStrategy(match_expressions)
 actor = Actor.options(scheduling_strategy = actor_affinity_strategy).remote()
@@ -134,7 +134,7 @@ actor = Actor.options(scheduling_strategy = actor_affinity_strategy).remote()
 4. The target actor is not expected to be scheduled with the actors whose label key exists "location".
 ```python
 match_expressions = [
- ActorAffinityMatchExpression("location", ActorAffinityOperator.DOES_NOT_EXIST, [], False)
+ LabelMatchExpression("location", ActorAffinityOperator.DOES_NOT_EXIST, [], False)
 ]
 actor_affinity_strategy = ActorAffinitySchedulingStrategy(match_expressions)
 actor = Actor.options(scheduling_strategy = actor_affinity_strategy).remote()
@@ -143,8 +143,8 @@ actor = Actor.options(scheduling_strategy = actor_affinity_strategy).remote()
 5. You can also set multiple expressions at the same time, and multiple expressions need to be satisfied when scheduling.
 ```python
 match_expressions = [
- ActorAffinityMatchExpression("location", ActorAffinityOperator.DOES_NOT_EXIST, [], False),
- ActorAffinityMatchExpression("version", ActorAffinityOperator.EXISTS, [], False)
+ LabelMatchExpression("location", LabelMatchOperator.DOES_NOT_EXIST, [], False),
+ LabelMatchExpression("version", LabelMatchOperator.EXISTS, [], False)
 ]
 actor_affinity_strategy = ActorAffinitySchedulingStrategy(match_expressions)
 actor = Actor.options(scheduling_strategy = actor_affinity_strategy).remote()
@@ -181,70 +181,70 @@ Set the labels for this actor API
 Actor affinity scheduling strategy API
 ```java
 public class ActorAffinitySchedulingStrategy implements SchedulingStrategy {
-  private ActorAffinitySchedulingStrategy(List<ActorAffinityMatchExpression> expressions) {
+  private ActorAffinitySchedulingStrategy(List<LabelMatchExpression> expressions) {
 }
 
-public class ActorAffinityMatchExpression {
+public class LabelMatchExpression {
   private String key;
-  private ActorAffinityOperator operator;
+  private LabelMatchOperator operator;
   private List<String> values;
   private boolean isSoft;
 
   /**
    * Returns an affinity expression to indicate that the target actor is expected to be scheduled
    * with the actors whose label meets one of the composed key and values. eg:
-   * ActorAffinityMatchExpression.in("location", new ArrayList<>() {{ add("dc-1");}}, false).
+   * LabelMatchExpression.in("location", new ArrayList<>() {{ add("dc-1");}}, false).
    *
    * @param key The key of label.
    * @param values A list of label values.
    * @param isSoft If true, the actor will be scheduled even there's no matched actor.
-   * @return ActorAffinityMatchExpression.
+   * @return LabelMatchExpression.
    */
-  public static ActorAffinityMatchExpression in(String key, List<String> values, boolean isSoft) {
-    return new ActorAffinityMatchExpression(key, ActorAffinityOperator.IN, values, isSoft);
+  public static LabelMatchExpression in(String key, List<String> values, boolean isSoft) {
+    return new LabelMatchExpression(key, LabelMatchOperator.IN, values, isSoft);
   }
 
   /**
    * Returns an affinity expression to indicate that the target actor is not expected to be
    * scheduled with the actors whose label meets one of the composed key and values. eg:
-   * ActorAffinityMatchExpression.notIn( "location", new ArrayList<>() {{ add("dc-1");}}, false).
+   * LabelMatchExpression.notIn( "location", new ArrayList<>() {{ add("dc-1");}}, false).
    *
    * @param key The key of label.
    * @param values A list of label values.
    * @param isSoft If true, the actor will be scheduled even there's no matched actor.
-   * @return ActorAffinityMatchExpression.
+   * @return LabelMatchExpression.
    */
-  public static ActorAffinityMatchExpression notIn(
+  public static LabelMatchExpression notIn(
       String key, List<String> values, boolean isSoft) {
-    return new ActorAffinityMatchExpression(key, ActorAffinityOperator.NOT_IN, values, isSoft);
+    return new LabelMatchExpression(key, LabelMatchOperator.NOT_IN, values, isSoft);
   }
 
   /**
    * Returns an affinity expression to indicate that the target actor is expected to be scheduled
    * with the actors whose labels exists the specified key. eg:
-   * ActorAffinityMatchExpression.exists("location", false).
+   * LabelMatchExpression.exists("location", false).
    *
    * @param key The key of label.
    * @param isSoft If true, the actor will be scheduled even there's no matched actor.
-   * @return ActorAffinityMatchExpression.
+   * @return LabelMatchExpression.
    */
-  public static ActorAffinityMatchExpression exists(String key, boolean isSoft) {
-    return new ActorAffinityMatchExpression(
-        key, ActorAffinityOperator.EXISTS, new ArrayList<String>(), isSoft);
+  public static LabelMatchExpression exists(String key, boolean isSoft) {
+    return new LabelMatchExpression(
+        key, LabelMatchOperator.EXISTS, new ArrayList<String>(), isSoft);
   }
 
   /**
    * Returns an affinity expression to indicate that the target actor is not expected to be
    * scheduled with the actors whose labels exists the specified key. eg:
-   * ActorAffinityMatchExpression.doesNotExist("location", false).
+   * LabelMatchExpression.doesNotExist("location", false).
    *
    * @param key The key of label.
    * @param isSoft If true, the actor will be scheduled even there's no matched actor.
-   * @return ActorAffinityMatchExpression.
+   * @return LabelMatchExpression.
    */
-  public static ActorAffinityMatchExpression doesNotExist(String key, boolean isSoft) {
-    return new ActorAffinityMatchExpression(
-        key, ActorAffinityOperator.DOES_NOT_EXIST, new ArrayList<String>(), isSoft);
+  public static LabelMatchExpression doesNotExist(String key, boolean isSoft) {
+    return new LabelMatchExpression(
+        key, LabelMatchOperator.DOES_NOT_EXIST, new ArrayList<String>(), isSoft);
   }
 
 }
@@ -282,7 +282,7 @@ locationValues.add("dc_1");
 locationValues.add("dc_2");
 ActorAffinitySchedulingStrategy schedulingStrategy =
     new ActorAffinitySchedulingStrategy.Builder()
-        .addExpression(ActorAffinityMatchExpression.in("location", locationValues, false))
+        .addExpression(LabelMatchExpression.in("location", locationValues, false))
         .build();
 ActorHandle<Counter> actor2 =
     Ray.actor(Counter::new, 1).setSchedulingStrategy(schedulingStrategy).remote();
@@ -294,7 +294,7 @@ List<String> values = new ArrayList<>();
 values.add("dc-1");
 ActorAffinitySchedulingStrategy schedulingStrategyNotIn =
     new ActorAffinitySchedulingStrategy.Builder()
-        .addExpression(ActorAffinityMatchExpression.notIn("location", values, false))
+        .addExpression(LabelMatchExpression.notIn("location", values, false))
         .build();
 ActorHandle<Counter> actor3 =
     Ray.actor(Counter::new, 1).setSchedulingStrategy(schedulingStrategyNotIn).remote();
@@ -304,7 +304,7 @@ ActorHandle<Counter> actor3 =
 ```java
 ActorAffinitySchedulingStrategy schedulingStrategyExists =
     new ActorAffinitySchedulingStrategy.Builder()
-        .addExpression(ActorAffinityMatchExpression.exists("version", false))
+        .addExpression(LabelMatchExpression.exists("version", false))
         .build();
 ActorHandle<Counter> actor4 =
     Ray.actor(Counter::new, 1).setSchedulingStrategy(schedulingStrategyExists).remote();
@@ -315,7 +315,7 @@ Assert.assertEquals(actor4.task(Counter::getValue).remote().get(10000), Integer.
 ```java
 ActorAffinitySchedulingStrategy schedulingStrategyDoesNotExist =
     new ActorAffinitySchedulingStrategy.Builder()
-        .addExpression(ActorAffinityMatchExpression.doesNotExist("version", false))
+        .addExpression(LabelMatchExpression.doesNotExist("version", false))
         .build();
 ActorHandle<Counter> actor5 =
     Ray.actor(Counter::new, 1).setSchedulingStrategy(schedulingStrategyDoesNotExist).remote();
@@ -325,8 +325,8 @@ ActorHandle<Counter> actor5 =
 ```java
 ActorAffinitySchedulingStrategy schedulingStrategy =
     new ActorAffinitySchedulingStrategy.Builder()
-        .addExpression(ActorAffinityMatchExpression.doesNotExist("version", false))
-        .addExpression(ActorAffinityMatchExpression.Exists("location", false))
+        .addExpression(LabelMatchExpression.doesNotExist("version", false))
+        .addExpression(LabelMatchExpression.Exists("location", false))
         .build();
 ActorHandle<Counter> actor6 =
     Ray.actor(Counter::new, 1).setSchedulingStrategy(schedulingStrategy).remote();
@@ -412,10 +412,10 @@ message ResourcesData {
   // heartbeat enabled.
   bool resources_available_changed = 3;
 
-  // Map<key, Map<value, reference_count>> Actors scheduled to this node and actor labels information
-  repeat Map<string, Map<string, int>> actor_labels = 15
+  // Map<label_type, Map<namespace, Map<label_key, label_value>>> Actors/Tasks/Nodes labels information
+  repeat Map<string, Map<string, Map<string, string>>> labels = 15
   // Whether the actors of this node is changed.
-  bool actor_labels_changed = 16,
+  bool labels_changed = 16,
 }
 
 
@@ -426,8 +426,8 @@ NodeResources {
   ResourceRequest load;
   /// Resources owned by normal tasks.
   ResourceRequest normal_task_resources
-  /// Actors scheduled to this node and actor labels information
-  absl::flat_hash_map<string, absl::flat_hash_map<string, int>> actor_labels;
+  /// Map<label_type, Map<namespace, Map<label_key, label_value>>> Actors/Tasks/Nodes labels information
+  absl::flat_hash_map<string, absl::flat_hash_map<string, absl::flat_hash_map<string, string>>> labels;
 }
 ```
 
@@ -465,12 +465,12 @@ Actor scheduling flowchart：
 ![Actor scheduling flowchart](https://user-images.githubusercontent.com/11072802/202128385-f72609c5-308d-4210-84ff-bf3ba6df381c.png)
 
 Node Resources synchronization mechanism:
-![Node Resources synchronization mechanism](https://user-images.githubusercontent.com/11072802/202128406-b4745e6e-3565-41a2-bfe3-78843379bf09.png)
+![Node Resources synchronization mechanism](https://user-images.githubusercontent.com/11072802/203783157-fad67f25-b046-49ac-b201-b54942073823.png)
 
 4. Scheduling optimization through ActorLabels  
 Now any node raylet has ActorLabels information for all nodes. 
 However, when ActorAffinity schedules, if it traverses the Labels of all Actors of each node, the algorithm complexity is very large, and the performance will be poor.   
-<b> Therefore, it is necessary to generate a full-cluster ActorLabels index table to improve scheduling performance. <b>
+<b> Therefore, it is necessary to generate a full-cluster ActorLabels index table to improve scheduling performance. </b>
 
 ```
 class GcsLabelManager {
@@ -491,18 +491,12 @@ class GcsLabelManager {
 }
 ```
 
-<b> VS.  putting Labels into the custom resource solution  <b>  
-<b>Advantages:<b>
-1. Compared with the scheme of putting Labels in the custom resource. This scheme can also reuse the resource synchronization mechanism. Then it won't destroy the concept of coustrom resouce.
-2. The Label index table of all nodes can be constructed from the ActorLabels information of each node. If you use Custom Resource, you can't build.
-3. If the Custom Resouces scheme is used, the accuracy of custom resouces scheduling will be affected. The current scheme is completely independent of existing scheduling policies and resources and will not affect them. The code is also more concise.
-
-
-<b>DisAdvantages:<b>
-1. The interface for resource reporting and updating needs to be adapted to ActorLabels in ResouceData.
-
-<b>Issue<b>
+**Issue**
 1. Because there must be a delay in resource synchronization under raylet scheduling. So if actor affinity is Soft semantics, there will be inaccurate scheduling.
+
+    a. If the user selects the Soft strategy. That means that the user can accept the fact that the scheduling has a certain probability of being inaccurate.
+
+    b. Most users schedule a batch of actors on the same node. In this case we can do exactly the right thing.
 
 
 ### Failures and Special Scenarios
@@ -523,8 +517,7 @@ This solution is to learn the PodAffinity/NodeAffinity features of K8s。
 https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity
 
 ### what's the alternative to achieve the same goal?
-1、Ray 
-Now ray placement group can achieve the same goal. But PG is too heavy and complicated to be user friendly
+
 ## Compatibility, Deprecation, and Migration Plan
 
 ## Test Plan and Acceptance Criteria
@@ -536,7 +529,7 @@ All APIs will be fully unit tested. All specifications in this documentation wil
 Later, if necessary, you can extend the semantics of "OR" by adding "is_or_semantics" to ActorAffinitySchedulingStrategy.
 ```
 class ActorAffinitySchedulingStrategy:
-    def __init__(self, match_expressions: List[ActorAffinityMatchExpression], is_or_semantics = false):
+    def __init__(self, match_expressions: List[LabelMatchExpression], is_or_semantics = false):
         self.match_expressions = match_expressions
         self.is_or_semantics = 
 ```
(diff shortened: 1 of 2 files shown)


reps/2022-08-31-actor-affinity-apis.md
```diff
@@ -0,0 +1,405 @@
...
+The implementation of Raylet scheduling mode is same as GCS scheduling above.  
+Mainly, one more Labels information needs to be synchronized to all Raylet nodes
+
+## Compatibility, Deprecation, and Migration Plan
+
+## Test Plan and Acceptance Criteria
+All APIs will be fully unit tested. All specifications in this documentation will be thoroughly tested at the unit-test level. The end-to-end flow will be tested within CI tests. Before the beta release, we will add large-scale testing to precisely understand scalability limitations and performance degradation in large clusters. 
+
+## (Optional) Follow-on Work
+
```
jjyao, line 405: Could you discuss what happens if the actor cannot be scheduled when the matching expression cannot be satisfied?

Since we asynchronously create the actor, what if the next actor wants to co-locate with the previous actor but the gcs hasn't received the label information of the previous actor. Do we just ask the application author to make sure the dependent actors are already scheduled?

jjyao, line 405: We need to discuss more about the failure modes.

For example if actor B collocates with actor A:

What if actor A fails/out of scope, what should happen to actor B?
What if actor A fails and restarts on a different node, what should happen to actor B?

jjyao (commented):
This closely matches the k8s pod affinity/anti-affinity (https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity).

We may want label based node affinity/anti-affinity as well in the future but I think it can be solved separately.

reps/2022-08-31-actor-affinity-apis.md
```diff
@@ -0,0 +1,405 @@
...
+
+## Design and Architecture
+
+### Brief idea
+
+1. Tag the Actor with key-value labels first
+2. Then affinity or anti-affinity scheduling to the actors of the specified labels.
+
+![ActorAffinity调度](https://user-images.githubusercontent.com/11072802/188054945-48d980ed-2973-46e7-bf46-d908ecf93b31.jpg)
```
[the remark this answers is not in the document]
larrylian: 可以的。  正和我意 哈哈哈

reps/2022-08-31-actor-affinity-apis.md
```diff
@@ -0,0 +1,405 @@
...
+SchedulingStrategyT = Union[None, str,
+                            PlacementGroupSchedulingStrategy,
+                            NodeAffinitySchedulingStrategy,
+                            ActorAffinitySchedulingStrategy]
+
+class ActorAffinitySchedulingStrategy:
+    def __init__(self, match_expressions: List[ActorAffinityMatchExpression]):
+        self.match_expressions = match_expressions
+
+class ActorAffinityMatchExpression:
```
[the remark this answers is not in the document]
larrylian: 这个隔离是以 ray-namespace来隔离。 不同namesapce之间是隔离的。  不同job想要一起调度，就要设置相同的namespace

reps/2022-08-31-actor-affinity-apis.md
```diff
@@ -0,0 +1,405 @@
...
+    """An expression used to represent an actor's affinity.
+    Attributes:
+        key: the key of label
+        operator: IN、NOT_IN、EXISTS、DOES_NOT_EXIST,
```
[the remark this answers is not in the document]
larrylian: 对的。 这个特性就是借鉴k8s的 Afinity and anti-affinity特性。 所以这4个关键字也是一样

reps/2022-08-31-actor-affinity-apis.md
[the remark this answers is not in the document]
larrylian: This isolation is based on ray-namespace isolation. There is isolation between different namesapce. If different jobs want to be scheduled together, they must set the same namespace

reps/2022-08-31-actor-affinity-apis.md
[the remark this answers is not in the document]
larrylian: correct. This feature is based on the Afinity and anti-affinity features of k8s. So these 4 keywords are also the same

reps/2022-08-31-actor-affinity-apis.md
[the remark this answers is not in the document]
larrylian: ok

reps/2022-08-31-actor-affinity-apis.md
```diff
@@ -0,0 +1,405 @@
...
+match_expressions = [
+ ActorAffinityMatchExpression("location", ActorAffinityOperator.DOES_NOT_EXIST, [], False)
+]
+actor_affinity_strategy = ActorAffinitySchedulingStrategy(match_expressions)
+actor = Actor.options(scheduling_strategy = actor_affinity_strategy).remote()
+```
+
+5. You can also set multiple expressions at the same time, and multiple expressions need to be satisfied when scheduling.
+```python
+match_expressions = [
```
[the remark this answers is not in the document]
larrylian: Don't support OR semantic.  Because of the AND semantics +  the operation in "IN/ NOT_IN / EXISTS/ DOES_NOT_EXIST", all scenarios can already be expressed. Adding the semantics of OR will increase the complexity of the API, making it difficult for users to get started.

reps/2022-08-31-actor-affinity-apis.md
larrylian: Yes.   If the matching expression cannot be satisfied, The actor will be add to the pending actor queue.   Util the matching expression all be statisfied.

larrylian: OK, I will add to ”Failures and Special Scenarios“ chapter.  

If actor B collocates with actor A and The actor A & B both complete schedule to Node 1.
1、If actor A failed and restarted to Node 2, the actor B will not be affected, it still working in Node 1.  
2、If actor A failed and restarted to Node 2, Then the actor B also failed, It will be scheduled to Node 2 and collocates with actor A.

reps/2022-08-31-actor-affinity-apis.md
```diff
@@ -0,0 +1,420 @@
...
+
+Yes, this will be a complement to ray core's ability to flexibly schedule actors/tasks.
+
+## Stewardship
+### Required Reviewers
+@WangTaoTheTonic
+### Shepherd of the Proposal (should be a senior committer)
+
+
+## Design and Architecture
```
scv119, line 27: can we also add a section on how other system achieve the same goal?

reps/2022-08-31-actor-affinity-apis.md
```diff
@@ -0,0 +1,420 @@
...
+
+### Brief idea
+
+1. Tag the Actor with key-value labels first
```
scv119, line 31: Could we apply this API to node as well?

reps/2022-08-31-actor-affinity-apis.md
```diff
@@ -0,0 +1,420 @@
...
+3. ActorAffinityMatchExpression has 4 match types which are IN/NOT_IN/EXISTS/DOES_NOT_EXIST
+
+Use Case | ActorAffinityOperator
+-- | --
+Affinity | IN/EXISTS
+Anti-Affinity | NOT_IN/DOES_NOT_EXIST
+
+This is to learn the Affinity & Anti-Affinity features of K8s.
+
+Advantage
```
scv119, line 48: Also, let's add concrete use cases on what this API enables?

reps/2022-08-31-actor-affinity-apis.md
larrylian: K8s uses the same method to achieve affinity/antiaffinity goals. Then other systems I estimate will take some time to investigate.

reps/2022-08-31-actor-affinity-apis.md
larrylian: Yes. Node Affinity can also be implemented this way. I have the same consideration later. But it needs to be done after ActorAffinity feature with raylet scheduling mode  completed.

reps/2022-08-31-actor-affinity-apis.md
```diff
@@ -0,0 +1,420 @@
...
+        self.match_expressions = match_expressions
+
+class ActorAffinityMatchExpression:
+    """An expression used to represent an actor's affinity.
+    Attributes:
+        key: the key of label
+        operator: IN、NOT_IN、EXISTS、DOES_NOT_EXIST,
+                  if EXISTS、DOES_NOT_EXIST, values set []
+        values: a list of label value
+        soft: ...
```
scv119, line 84: what's the semantic of soft?

jovany-wang:
For failure section, we should add more discussions on how we promise the scheduling result consensus.

I'd also like to know if it's possible to achieve the following typical goals:
distribute one actor for each node, also it could be promised that every nodes has 1 actor even a new node joined or exited.

reps/2022-08-31-actor-affinity-apis.md
larrylian: yes

reps/2022-08-31-actor-affinity-apis.md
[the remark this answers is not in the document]
jjyao: Given we are mimicking k8s apis, we should probably do some research to see what problems people have with these apis (if there is any) so we can avoid the same mistake.

larrylian:
> For failure section, we should add more discussions on how we promise the scheduling result consensus.

I'll add some more failure scenarios


> I'd also like to know if it's possible to achieve the following typical goals: distribute one actor for each node, also it could be promised that every nodes has 1 actor even a new node joined or exited.

This feature is too complicated, and needs to be discussed, so it is not implemented in this feature.

reps/2022-11-23-labels-mechanish-and-affinity-schedule-feature.md
```diff
@@ -0,0 +1,382 @@
+## Summary
+### General Motivation
+
+Introduce the Labels mechanism. Give Labels to Actors/Tasks/Nodes/Objects.
```
scv119, line 4: Introduce labels mechanism, which associate an enumerated property to Ray nodes. We can static assign labels to Ray node on start up, or dynamically change node's label through Ray scheduling APIs. 

Labels mechanism makes it easy to implement Actor and Node affinity.([Taints and Tolerations style](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/))

reps/2022-11-23-labels-mechanish-and-affinity-schedule-feature.md
```diff
@@ -0,0 +1,382 @@
...
+ 
+@wumuzi520 SenlinZhu @Chong Li  @scv119 (Chen Shen) @jjyao (Jiajun Yao) @Yi Cheng
+### Shepherd of the Proposal (should be a senior committer)
+
+
+## Design and Architecture
+
+### Brief idea
+1. Introduce the concept of Label. Add the Labels attribute to Actor/Task/Node. Labels = Map<String, String>.
+2. After Actor/Task are scheduled to a certain node, the Labels of Actor/Task will be attached to the node resource(Named: LabelsResource). Node's Labels are naturally in the node resource.
```
scv119, line 23: let's remove task for now as it might not applicable.

reps/2022-11-23-labels-mechanish-and-affinity-schedule-feature.md
```diff
@@ -0,0 +1,382 @@
...
+    """
+    def __init__(self, key: str, operator: LabelMatchOperator,
+                 values: List[str], soft: bool):
+        self.key = key
+        self.operator = operator
+        self.values = values
+        self.soft = soft
+
+# ActorAffinity use case
+actor_1 = Actor.options(scheduling_strategy=ActorAffinitySchedulingStrategy([
```
scv119, line 79: is there a better alternative?
for example:
```
actor_1 = Actor.options(
   resources={"4c8g": 1},
   in("location", ["dc_1"]),
    ....
}
``` 

or

```
actor_1 = Actor.options(
   resources={"4c8g": 1},
  NodeSelector("location in [dc_1, dc_2]")
    ....
}
```

scv119:
@ericl @stephanie-wang @rkooo567 wanna take a look and given some comments before we start voting process?

ericl:
The proposal seems reasonable to me (narrowly scoped enhancement to actor scheduling; should be orthogonal to future scheduling improvements).

Who is the shepherd here? How about @stephanie-wang ?

reps/2022-08-31-actor-affinity-apis.md
```diff
@@ -0,0 +1,538 @@
+
+## Summary
+### General Motivation
+
+Provides a set of lightweight actor affinity and anti-affinity scheduling interfaces.
+Replacing the heavy PG interface to implement collocate or spread actors.
```
ericl, line 6: To clarify, this proposal only allows scheduling of actors relative to other actors, correct?

reps/2022-08-31-actor-affinity-apis.md
```diff
@@ -0,0 +1,538 @@
...
+
+* Affinity
+  * Co-locate the actors in the same node.
+  * Co-locate the actors in the same batch of nodes, like nodes in the same zones
+* Anti-affinity
+  * Spread the actors of a service across nodes and/or availability zones, e.g. to reduce correlated failures.
```
ericl, line 12: How is it possible to spread across AZs using this proposal?

ericl:
> Give Labels to Actors/Nodes

I have concerns about adding labels to nodes, since we already have the notion of custom resources. Any reason not to use custom resources there?

stephanie-wang:
Hmm I agree with @ericl, I'm concerned the current scope of the REP is too broad. If we want to go ahead with full labels and expressions, we need deeper design review for the API.

For the use cases listed, I think we could pare down this REP to only allowing passing in other actor handles and/or custom resources to the scheduling strategy, and just use flags for affinity/anti-affinity instead of expressions. How does that sound?

reps/2022-08-31-actor-affinity-apis.md
larrylian: Yes. The current implementation of ActorAffinity is to choose affinity or anti-affinity to other actors when scheduling actors. However, it can also be extended to the relationship between task and actor in the future. 
image

reps/2022-08-31-actor-affinity-apis.md
larrylian: 1. First use custom resource or other methods to schedule an actor to the specified AZs.
2. Then ohters actors just use actor-affinity feature to affinity with the actor corresponding to AZs.

larrylian:
> > Give Labels to Actors/Nodes
> 
> I have concerns about adding labels to nodes, since we already have the notion of custom resources. Any reason not to use custom resources there?

1. NodeAffinity may look similar to custom resource. But ActorAffinity is completely different. Our REP will first implement ActorAffinity and then extend NodeAffinity.
2. Adding Labels to Node looks similar to custom resource, but there is a big difference in reality. Custom resource is a kind of resource, which is of double type and can be used for resource counting. But Labels are attributes or tags. He is not a double type and is not used for counting.
3.  Add labels to node is another simplification and supplement to custom resource.  
 Labels are more user-friendly in many scenarios and easier for users to understand.   
for example:
If the nodes of different AZs are isolated using custom resources,  
the node setting of "AZ-1" is {"AZ-1": 9999} and the node setting of "AZ-2" is {"AZ-2": 9999}.   
Then Actor set resource{"AZ-1": 1} to scheduling.   
The count in custom_resource is actually redundant when scheduling actors. Users will be confused.

But the words of user labels and NodeAffinity. 
You can directly set labels {"AZ": "AZ-1"} ，{ "AZ": "AZ-2"} to Node.   
Then when the user schedules actors, use NodeAffinity affinity {"AZ": "AZ-1"}.

larrylian:
> Hmm I agree with @ericl, I'm concerned the current scope of the REP is too broad. If we want to go ahead with full labels and expressions, we need deeper design review for the API.
> 
> For the use cases listed, I think we could pare down this REP to only allowing passing in other actor handles and/or custom resources to the scheduling strategy, and just use flags for affinity/anti-affinity instead of expressions. Of course we could always improve it later to add labels/expressions once we have more information from users. How does that sound?


This REP introduces the concept of labels, which is quite different from custom_resource. This labels will greatly improve the scheduling of ray.
At a high level, we have reached an agreement with @scv119 @iycheng  on the introduction of labels and affinity scheduling, and we think this is a good idea.

rkooo567 (commented):
I believe we need a section when scheduling cannot happen. More concretely, what's the behavior when

1. resources are enough, but affinity cannot be satisfied?
2. affinity can be satisfied, but resources are not enough? 
3. both affinity and resources are not satisfied?

Do we throw an exception? Will it just hang? If so, how will this be exposed to users?

This also requires how this API will play with autoscaler. (if it can't satisfy the resource request now, but it can after autoscaling up?)

ericl:
Thanks for the comments. Overall, I agree with the value of adding node labels for more straightforward and powerful affinity/anti-affinity scheduling.

I do think this is a pretty big change--- perhaps the biggest change to Ray scheduling since placement groups. The scope and design of the change are my main concerns. I don't think the current REP sufficiently outlines the changes necessary to fully support this feature in Ray.

So here's my proposal, we split this REP into two separate REPs:
1. Introducing the label mechanism for nodes. This should cover:
   - The proposed API for ray start, scheduling_strategy.
   - Ray autoscaler support for labels (including node config, changes to node selection strategy, infeasible error messages).
2. Extension of the label mechanism for actor affinity.
   - The proposed API.
   - How to support this in the Ray autoscaler.

Thoughts? Also cc @wuisawesome

stephanie-wang:
Thanks, @larrylian, I see the value of having labels, but I think we should try to get consensus on a more minimal short-term API change that will still be useful. Once we expand the scheduling API it becomes very difficult to go back, so we need to focus our efforts on APIs that will fit common use cases.

What do you think about the following, and we can split these into one REP each? These will likely not cover all of the use cases that you mentioned, but I believe they are uncontroversial and will already add value.
1. Scheduling strategies with static node labels (can only be set at node start)
2. ActorAffinity scheduling strategy using actor handles - the reasoning here is that using labels for actor affinity scheduling actually adds an additional unnecessary abstraction that users have to think about. For most users, I think the most intuitive API is to just specify actor handles to co-locate with or avoid.

Long-term, we can think about the following solutions, and for now just summarize the key use cases:
1. Dynamic labels / labels created by actor scheduling.
2. More advanced/flexible expressions in scheduling strategies.

wuisawesome:
```
So here's my proposal, we split this REP into two separate REPs:

Introducing the label mechanism for nodes. This should cover:
The proposed API for ray start, scheduling_strategy.
Ray autoscaler support for labels (including node config, changes to node selection strategy, infeasible error messages).
Extension of the label mechanism for actor affinity.
The proposed API.
How to support this in the Ray autoscaler.
Thoughts? Also cc @wuisawesome
```

I agree. The first part is very similar to the k8s API and seems clean and uncontroversial.

The second part (actor affinity) seems significantly more complicated and risky.

reps/2022-08-31-actor-affinity-apis.md
```diff
@@ -0,0 +1,538 @@
...
+
+
+## Design and Architecture
+
+### Brief idea
+
+1. Tag the Actor with key-value labels first
+2. Then affinity or anti-affinity scheduling to the actors of the specified labels.
+
+![ActorAffinityScheduling](https://user-images.githubusercontent.com/11072802/188054945-48d980ed-2973-46e7-bf46-d908ecf93b31.jpg)
```
wuisawesome, line 35: Are there any examples of existing systems that have this ActorAffinity concept? Or was there any existing inspiration for this?

larrylian:
> I believe we need a section when scheduling cannot happen. More concretely, what's the behavior when
> 
> 1. resources are enough, but affinity cannot be satisfied?
> 2. affinity can be satisfied, but resources are not enough?
> 3. both affinity and resources are not satisfied?
> 
> Do we throw an exception? Will it just hang? If so, how will this be exposed to users?
> 
> This also requires how this API will play with autoscaler. (if it can't satisfy the resource request now, but it can after autoscaling up?)

Thanks for the good question. For this abnormal scene, I will add a chapter to REP.   

Q1:  
> 1. resources are enough, but affinity cannot be satisfied?  -> Hang  & report schedule failed event and detail unstaisfed reason to exposed to users
> 2. affinity can be satisfied, but resources are not enough? -> Hang  & report schedule failed event and detail unstaisfed reason to exposed to users
> 3. both affinity and resources are not satisfied? -> Hang  & report schedule failed event and detail unstaisfed reason to exposed to users

Q2：
I don't think this is a big problem. This ActorAffinity/NodeAffinity is just a new scheduling strategy similar to the previous SPEAD/DEAFULT/PG. How autoscale is implemented now, just adapt it at that time.

reps/2022-08-31-actor-affinity-apis.md
larrylian: This ActorAffinity is similar to podAffinity/NodeAffinity of k8s.

larrylian:
> I agree. The first part is very similar to the k8s API and seems clean and uncontroversial.
> 
> The second part (actor affinity) seems significantly more complicated and risky.

ActorAffinity is the specific implementation feature discussed in the first part.   
It is similar to PodAffinity/NodeAffinity of K8s.  
It doesn't get any more complicated. You think it is complicated because k8s is described in yaml, but now it is described in python.

larrylian:
> I don't think the current REP sufficiently outlines the changes necessary to fully support this feature in Ray.

What is mentioned now is that ActorAffinity/NodeAffinity is actually part of PG 2.0.   
Now only PG can be used to achieve the ability of Affinity. However, PG has the ability of affinity, resource reserve, and atomic scheduling at the same time, so that its use method is too complicated and the implementation of PG is also very complicated. It will often cause problems under high-load use of large clusters, and its performance is not good.   
So now ActorAffinity is used to replace PG's Affinity ability alone, which is more pure and simple to use.
I will add this to the REP.

Q2:
> So here's my proposal, we split this REP into two separate REPs:

I am now actually divided into two REPs for discussion. Only now both REPs are in the same PR.  
2022-11-23-labels-mechanish-and-affinity-schedule-feature.md -> Introducing the label mechanism for nodes. and ActorAffiinty & NodeAffinity High level discussion and API form.  
2022-08-31-actor-affinity-apis.md -> Extension of the label mechanism for actor affinity.  API 、abnormal behavior and implencement

> Q3: autoscale   
That's a good question.
I think this problem is not difficult to solve.
This ActorAffinity/NodeAffinity is just a new scheduling strategy similar to the previous SPEAD/DEAFULT/PG.   
Now it only needs to be adapted according to the implementation of other scheduling strategies.

larrylian:
> 1. Scheduling strategies with static node labels (can only be set at node start)
> 2. ActorAffinity scheduling strategy using actor handles - the reasoning here is that using labels for actor affinity scheduling actually adds an additional unnecessary abstraction that users have to think about. For most users, I think the most intuitive API is to just specify actor handles to co-locate with or avoid.
> 
> Long-term, we can think about the following solutions, and for now just summarize the key use cases:
> 
> 1. Dynamic labels / labels created by actor scheduling.
> 2. More advanced/flexible expressions in scheduling strategies.

Thanks for your very valuable opinion.  

> Q1:ActorAffinity scheduling strategy using actor handles

This is a very good idea, so I added the syntactic sugar of using actor_handle to solution 1.
```
actor_0 = Actor.remote()
actor_1 = Actor.options(
        scheduling_strategy=actor_affinity(affinity(actor_0, false))
    ).remote()

actor_2 = Actor.options(
        scheduling_strategy=actor_affinity(anti_affinity(actor_0, false))
    ).remote()
```

>Q2: Scheduling strategies with static node labels (can only be set at node start)

For Node's Label, I also intend to do the same. The first step is to use a static label. Then expand the dynamic node label in the long term.  
But the dynamic label of Actor. I'll do it on the first step. Because this is the basis of the actorAffinity feature.

>Q3: 
> 1. Dynamic labels / labels created by actor scheduling.
> 2. More advanced/flexible expressions in scheduling strategies.

This is my opinion is the first step to achieve.   
Because my intention is to implement ActorAffinity first and then NodeAffinity.  
Then I mentioned the Option 1 API. In fact, it is relatively simple. If you simply use actor_handle to express afiinity/anti_affinity. It is unsatisfactory in complex nested actor scenes.

stephanie-wang:
> But the dynamic label of Actor. I'll do it on the first step. Because this is the basis of the actorAffinity feature.
> 
> > Q3:
> > 
> > 1. Dynamic labels / labels created by actor scheduling.
> > 2. More advanced/flexible expressions in scheduling strategies.
> 
> This is my opinion is the first step to achieve. Because my intention is to implement ActorAffinity first and then NodeAffinity. Then I mentioned the Option 1 API. In fact, it is relatively simple. If you simply use actor_handle to express afiinity/anti_affinity. It is unsatisfactory in complex nested actor scenes.

Thanks for updating the REPs! But I don't quite understand why we need labels to implement ActorAffinity. It seems much easier to just implement actor affinity directly with handles:

```
schedule_with_affinity(actor, handles):
  locs = get_node_locs(handles)
  schedule_at_nodes(actor, best_loc(locs))
```

Scheduling with dynamic labels on the other hand introduces complexity around what happens if the label fails to be created due to actor failure, if the label is created by something other than the intended actor due to user error, how to make this work with autoscaling, etc. With dynamic labels, it seems like there will always be an issue of knowing how long to wait for a label to appear. I can see how the labels could help with scalability of actor affinity in the future but the added complexity doesn't seem worthwhile for a first version.

In the interest of supporting the desired feature ASAP (actor affinity scheduling), I suggest:
1. We decouple the actor affinity REP from labels completely.
2. Defer all discussion of dynamic labels (including ones created by actors) to a later REP.

larrylian:
> Thanks for updating the REPs! But I don't quite understand why we need labels to implement ActorAffinity. It seems much easier to just implement actor affinity directly with handles:
> 
> ```
> schedule_with_affinity(actor, handles):
>   locs = get_node_locs(handles)
>   schedule_at_nodes(actor, best_loc(locs))
> ```
> 
> Scheduling with dynamic labels on the other hand introduces complexity around what happens if the label fails to be created due to actor failure, if the label is created by something other than the intended actor due to user error, how to make this work with autoscaling, etc. With dynamic labels, it seems like there will always be an issue of knowing how long to wait for a label to appear. I can see how the labels could help with scalability of actor affinity in the future but the added complexity doesn't seem worthwhile for a first version.
> 
> In the interest of supporting the desired feature ASAP (actor affinity scheduling), I suggest:
> 
> 1. We decouple the actor affinity REP from labels completely.
> 2. Defer all discussion of dynamic labels (including ones created by actors) to a later REP.

Let me give you a few examples to help you understand why you must use Label to achieve ActorAffinity. Help you understand the value of Label.

To sum up, there are three problems with only using actor_handle to implement ActorAffinity.
1. When Actor FO is scheduled, there will be a problem that the scheduling does not meet expectations. - this is the worst problem
2. In many scenes, it is impossible to obtain the actor handle, or it will be very complicated to obtain the actor handle.
3. In a scene with many actors, it is very complicated to only use the actor handle. Many jobs now consist of hundreds or thousands of actors.

1. Actor FO：
Spead Actor to each Node. The way to spead actor with use only actor_handle is:  
1. Actor A  normal schedule.
2. Actor B schedule_with_anti_affinity(A)
3. Actor C schedule_with_anti_affinity([A,B])
4. If actor A occur FO, A will maybe schedule to node which B or C.
Actor FO导致不是反亲和了


2. Can't obtain actor handle
> Driver  ->  Actor A  ->   Actor B
When A creates B, he wants B and A to have an affinity. Satisfy localization. At this time, A cannot obtain its own handle.

> A certain type of Actor in the multi-level Actor in the DAG wants to achieve affinity or anti-affinity.
Output Actors is created by Reduce Actors. But output actors want to co-locate。   
In this scenario, it is difficult for a single reduce to obtain all input actor handles. But using a label to label the Output Actor with the same label can easily achieve Affinity/anti-affinity.
image

3. Many actors.
Many DAGs or jobs now have hundreds of actors. You have to fill in hundreds of actors at the same time, which is not good to use. And the performance is not good.

stephanie-wang:
Thanks for providing the context, but I believe that these issues are not necessarily the common case for Ray users today so these seem like premature optimizations right now. It would be better to introduce the API first and from there gather feedback from the rest of the OSS community on usability/scalability before we go ahead with a potentially complex system. Do you see a problem with the approach of implementing a basic version using only actor handles first and later deciding whether/how to add a labels-based implementation?

In more detail:
> 1. When Actor FO is scheduled, there will be a problem that the scheduling does not meet expectations. - this is the worst problem

Failover semantics is indeed something that should be covered in the REP but actually I don't see why you can't accomplish the same thing using only actor handles. The GCS should know exactly which actors are placed where, so it can simply rerun the affinity/anti-affinity policy on FO.

> 2. In many scenes, it is impossible to obtain the actor handle, or it will be very complicated to obtain the actor handle.

It should be straightforward to support something like "get own handle" or "pass `self` to ActorAffinity strategy". Otherwise, the main problem is how difficult the API is to use, which we don't yet have evidence from the community on.

> 3. In a scene with many actors, it is very complicated to only use the actor handle. Many jobs now consist of hundreds or thousands of actors.

Again, this API doesn't exist yet so we don't have evidence on usability. Also, there are very few use OSS use cases right now that are using this many actors.

ericl:
Maybe let's think about one particular edge case: scheduling a lot of actors each with mutual anti-affinity. IIUC, the concern is this leads to O(n) sized constraints, such as:

```
a1 = A.options().remote()
a2 = A.options(anti_affinity=[a1]).remote()
a3 = A.options(anti_affinity=[a1, a2]).remote()
..
an = A.options(anti_affinity=[a1, a2, ..., an-1]).remote()
```

This is true, but I wonder why the user wouldn't use a single SPREAD or STRICT_SPREAD placement group in the first place to schedule these actors. There is a pending REP to substantially improve the placement groups API as well, to be more flexible.

I think the combination of enhanced placement groups, node labels, and actor handle affinity could be pretty powerful. If we can see common situations that can't be handled by these features together, maybe let's try to dive more into those scenarios.

larrylian:
@stephanie-wang 
> Actor FO
```
a1 = A.options().remote()
a2 = A.options(anti_affinity=[a1]).remote()
a3 = A.options(anti_affinity=[a1, a2]).remote()
..
an = A.options(anti_affinity=[a1, a2, ..., an-1]).remote()
```
If a1 failover, a1 will maybe schedule to node of a2~an。

You can see Ericl's reply. If you only use actor-handle to implement anti-affinity. There will be this problem.   
This problem is very fatal, and it will cause enterprise users who pay attention to stability to not consider using this feature at all.

> 2. Can't obtain actor handle

Although you said that you can add "get own handle" to solve it. But this is just one scenario I exemplified. There are many other scenarios where actor handles cannot be easily obtained. For example, the DAG multi-level scenarios and cross-language calls we mentioned above. And if the number of Actors is relatively large, the code development cost for users to obtain actor_handle is very high.

> 3. Also, there are very few use OSS use cases right now that are using this many actors.


The diagram of the "output" operator above is just an example. It could also be "map" or "reduce" etc. It may also be a topology map of other scenarios.

Now a job with hundreds of actors is actually very common. Our ray has developed for so many years now, I think we should pay more attention to these enterprise-level large-scale scenarios now.

larrylian:
@ericl 

> 1. why the user wouldn't use a single SPREAD or STRICT_SPREAD placement group in the first place to schedule these actors. 

PG is only suitable for scenarios where the number of Actors is known in advance. But now many jobs are long-time jobs, which will be dynamically scaled with time or traffic. The number of Actors it wants to aggregate is not fixed.

> 2.  There is a pending REP to substantially improve the placement groups API as well, to be more flexible. 

Although if the PG is enhanced, the new PG resize function can solve the above problems. But it is also very complicated for users to use.   
example:
1.  Set PG bundle 1 resource{CPU=10},   schedule 10 Actors(per actor use 1 CPU) to bundle 1.
2. If the business needs to increase to 20 Actors.
3. User must first resize pg bundles resource to  {CPU=20}.
4. Then schedule other 10 actors to bundle 1.
5. if reduce to 5 actors at the time, users also need to resize pg resource to avoid wasting resources.

1 > For scenarios that only require Affinity/Anti-affinity functions, using PG requires users to manage PG resources by themselves. This will increase the complexity of use, which is not the original intention of ray.

2 > The current PG is already very complicated, and it will be more complicated to implement PG resize.  At that time, performace, consistency, stability, etc. will have big problem.

3 >  Now the implementation of PG needs to consider the three functions of affinity, resource reservation, and gang-schedule submitted in two phases at the same time, and the internal implementation is very complicated. This leads to many problems under high load usage of large-scale clusters:
1. A PG has 100+ or 1000+ bundles. If a bundle fails to be created (and recreated too), it will cause the entire PG to fail to be created. Or the initialization of a certain node job is slow, which will cause the creation of the entire PG to become very slow. These problems may not appear in the scenario of several bundles. But it will often appear in large-scale enterprise-level clusters.
2. Now PGs are created serially, which will cause the scheduling of one PG to hang, and cause other PGs to also get hang. This is also often encountered within us.

Because of these problems above and others. Our internal users often complain about the PG function. Then more and more users have given up using PG.   
Then use the ActorAffinity implemented by Labels, whether it is the user's use or the internal implementation, it will be much simpler. Our internal ActorAffinity feature has been online for user to use for half a year, and there have been no problems so far.

stephanie-wang:
> If a1 failover, a1 will maybe schedule to node of a2~an。
> 
> You can see Ericl's reply. If you only use actor-handle to implement anti-affinity. There will be this problem. This problem is very fatal, and it will cause enterprise users who pay attention to stability to not consider using this feature at all.

Thank you for the explanation, this makes sense.
 
> > 4. Do you see a problem with the approach of implementing a basic version using only actor handles first and later deciding whether/how to add a labels-based implementation?
> 
> It is because of the problems in many of the above scenarios, especially the Actor Failover scenario. As a result, only using actor_handle cannot meet the needs of users.

From my perspective, it seems like the issues with the actor handle implementation mainly come from large-scale multitenant Ray clusters with long-lived services. While I agree this is an important scenario in large-scale enterprise cases  that we should pay more attention to, this is not yet the common case for most users, especially the multitenancy part. For example, the only major Ray library that partly fits this use case is Ray Serve, and AFAIK Serve does not have as hard requirements for actor affinity/anti-affinity.

To be clear, I am not against implementing the labels version; I just don't think it's the right thing to start with right now. If we start with a more minimal initial version, we can collect feedback for a more holistic solution (that could include the improved placement groups API that @ericl mentioned), and there is also less chance that we will paint ourselves into a corner with an insufficient API. It's possible we will end up converging on labels in the end anyway, but it's not clear to me yet what the alternatives are and how it will fit with other ongoing scheduling enhancements.

So, *considering the current OSS use cases*, do you see a problem with the approach of implementing a basic version using only actor handles first?

larrylian:
@stephanie-wang 
> So, considering the current OSS use cases, do you see a problem with the approach of implementing a basic version using only actor handles first?

The solution of actor handle can only solve the simple actor affinity function. The Anti-affinity scenario cannot be solved. You pay more attention to OSS related scenarios. Let me take the following example. When the input actor reads data from the OSS, it needs to slice it and then distribute it to the downstream operator. At this time, input actors need to be distributed on each node. If only the actor handle is used for implementation, there is the problem of actor failover.
image

ericl:
This is an interesting discussion with regards to fault tolerance and simplicity of APIs. I think I agree that there are significant pain points actor label affinity solves. However, considering other scheduling improvements this might not be the first one we'd want to land. For example, I think it would be better to support basic labeling support first and nail down those interactions with the scheduler.

I think it would be productive to think about flexible scheduling more in general. If scheduling constraints could implemented in a nicely "pluggable" way, then policies like these should be easy to prototype and introduce without big architectural ramifications.

So here's my proposal:
- Let's think about a pluggable "scheduler constraint" system that supports the use cases in this REP (i.e., soft and hard constraints based on static and runtime scheduling information).
- We should think about how this would be implemented in depth, including autoscaler integration.
- Then we should have a bunch of possible scheduler plugins, one of which could be the actor label one.

As an example, for Ray runtime envs we have a plugin system which has been pretty useful for handling both basic and advanced use cases without needing any architectural changes.

larrylian:
@ericl  @stephanie-wang  @scv119 

Now I also realize that using a complex set of APIs at once wiil indeed have a great impact, and everyone may not be able to accept ti immediately.  So I agree with your previous premise that sample and valueable parts should be implemented first, and discuss the complex solutions later after this piece is completed.

I will splite a new REP to discuss the first part. The first part is mainly the following feature:

1.  NodeAffinity - Add static labels to node (can only be set at node start)) and Enhance the function of NodeAffinity based on Labels.
2. ActorAffinity - Add actorAffinity scheduling strategy using actor handles .    The labels of actors will not be introduced first, but the internal implementation will be implemented in the way of labels, so as to facilitate the expansion of actorAffinity based on labels in the future.

edoakes:

PR #14, opened 2022-09-19
Ray on Spark REPS

WeichenXu123:
Ray on spark REPS

### DIFF
--- a/reps/2022-09-19-ray-on-spark.md
+++ b/reps/2022-09-19-ray-on-spark.md
@@ -92,6 +92,12 @@ services that are running on the Spark driver and Spark workers.
 Finally, by adjusting the resource allocation for Ray nodes and Spark tasks, this approach enables
 users to run multiple Ray clusters in isolation of each other on a single Spark cluster, if desired.
 
+Due to the fact that execution is performed via Spark BarrierExecution, a failed task (e.g., a 
+node loss event, a heartbeat timeout in the Spark cluster manager service for an executor, etc.) the 
+Spark task executing the ray job will be retried as a single concurrent unit. Because of this mode of 
+operation and the inability to perform map-reduce task resumption for this initial design, recovery 
+by a resumption of tasks 'in-place' is not possible.
+
 ### Key questions
 
 #### We use spark barrier mode job to launch a ray cluster, shall we make the spark barrier to be a background job and keep running, until user explicitly terminate it ?
@@ -101,10 +107,12 @@ on the ray cluster with little overhead.
 
 
 #### Launch Ray head node on spark driver node or spark task side ?
-In spark driver side.
-Because we can set Ray head node with `num-cpus` = 0, so that no ray task will be scheduled
-to ray head node, so that we can launch ray head node in driver side, it won't consume
-too much computing / memory resources of driver node.
+
+The head node will be initialized on the Spark Driver. To ensure that tasks will not submitted 
+to this node, we will configure `num-cpus`=0 for the head node. 
+Additionally, in order to ensure that the head node is running on hardware sufficient to ensure 
+that requisite ray system processes, a validation of minimum hardware configuration requirements will 
+be performed prior to initialization of the ray cluster (minimum CPU cores, memory requirements)
 
 
 #### Shall we make ray worker node the same shape (assigned with the same resources amount) ?
@@ -180,7 +188,22 @@ Past examples includes horovod, petastorm, xgboost, tensorflow-spark, all of the
 
 #### Does it support autoscaling ?
 Currently no. This implementation relies on spark barrier mode job, which does not support autoscaling currently.
+In a future iteration, we plan to investigate (if warranted) using the Spark job scheduler to handle the 
+creation of ray worker nodes by mapping a single ray worker to a Spark job running in shared resource context (non-FIFO).
 
+#### What is the fault tolerance of Barrier Execution mode and how are failures handled?
+BarrierRDD mapping forces all configured tasks to execute concurrently. With no concept of map-reduce 
+fault tolerance in this mode, faults that occur within processing (that are not related to loss of 
+an underlying VM that the Spark worker node and, by extension, the Ray worker node are running on) will 
+be retried as an atomic unit. That is, all pending task transactions that are submitted to the ray cluster 
+on all nodes will re-execute. 
+We intend to development a retry loop within the BarrierRDD mapping function to allow for a limited 
+number of retries of a task submission before raising an Exception up the stack to the Spark driver node 
+to terminate the ray cluster. 
+As mentioned above, a loss of a Spark worker is unrecoverable and will require a new initialization of a 
+ray cluster once the dropped node(s) are recovered by new VMs. 
+Note that this hardware-loss lack of fault tolerance will be mitigated if we end up pursuing a NodeProvider approach 
+in the future (wherein a new Spark Job would be submitted to a recovered VM worker node to resume execution in a retry loop).
 
 #### What’s the level of isolation spark provides ?
 Barrier mode spark task provides process-level isolation.
@@ -211,9 +234,14 @@ And connect to the cluster.
 
 Returns an instance of type `RayClusterOnSpark``
 
+The ray head node may be of a different configuration compared to that of the worker nodes, but 
+the ray workers will be homogenous with respect to CPU / GPU and memory available for heap and 
+object store utilization.
+A best-effort mapping of available Spark cluster resources to requested ray cluster resources for 
+worker nodes will be performed. A validation check during initialization will be done to:
+- In `safe_mode'=True raise an Exception if the configured resources are insufficient, providing detailed reconciliation steps for the user.
+- In `safe_mode`=False, log a warning of potentially insufficient resources with instructions on how to configure the Spark cluster to avoid this situation.
 
-For each ray node, it might have different / random CPUs assigned ,
-but for the whole ray cluster, the resources amount allocated is deterministic.
 
 e.g., your case: (8-CPU nodes, or 4-GPU nodes),
 suppose on a spark cluster with config:
--- a/reps/2022-09-19-ray-on-spark.md
+++ b/reps/2022-09-19-ray-on-spark.md
@@ -50,10 +50,14 @@ The architecture of a Spark cluster is as follows:
 ![spark-cluster-overview](https://spark.apache.org/docs/latest/img/cluster-overview.png)
 
 Using Spark's barrier mode, we intend to represent each Ray worker node as a long-running Spark
-task, where the Ray worker node has the same set of resources as the Spark task. Multiple Spark
+task where the Ray worker node has the same set of resources as the Spark task. Multiple Spark
 tasks can run on a single Spark worker node, which means that multiple Ray worker nodes can run
 on a single Spark worker node. Finally, we intend to run the Ray head node on the Spark driver
 node with a fixed resource allocation.
+
+Note that if multiple ray-node-per-spark-worker issues occur (whether due to shared object store 
+location for workers, dashboard visualization confusion, or other unforeseen issues), the implementation 
+can be modified to only permit a single Ray cluster per Spark cluster. 
 
 To clarify further, the following example demonstrates how a Ray cluster with 16 total worker CPU
 cores can be launched on a Spark cluster, assuming that each Ray node requires 4 CPU cores and
@@ -65,7 +69,8 @@ cores can be launched on a Spark cluster, assuming that each Ray node requires 4
 ray start --head --num-cpus=0
 ```
 
-The Ray head node will operate solely as a  manager node, without worker node functionality.
+The Ray head node will operate configured to optimistically not execute raylets due to combination of 
+locality-aware scheduling and the available cpu resources for raylet execution. 
 
 2. Create a Spark barrier mode job, which executes all tasks in the spark job concurrently. Each
 task is allocated a fixed number of CPUs and a fixed amount of memory. In this example, we will
@@ -80,7 +85,7 @@ running until the Ray cluster destroyed. The command used to start each Ray work
 follows:
 
 ```
-ray start —head --num-cpus=X --num-gpus=Y --memory=Z --address={head_ip}:{port}
+ray start --num-cpus=X --num-gpus=Y --memory=Z --object-store-memory=M --address={head_ip}:{port}
 ```
 
 4. After the Ray cluster is launched, the user's Ray application(s) can be submitted to the Ray
--- a/reps/2022-09-19-ray-on-spark.md
+++ b/reps/2022-09-19-ray-on-spark.md
@@ -27,6 +27,15 @@ TODO
 
 ## Design and Architecture
 
+### Preparation
+
+We need to have a spark cluster (apache/spark >= 3.3)
+We need install ray packages on every node of the cluster, by:
+```
+pip install ray
+pip install "ray[debug,dashboard,tune,rllib,serve]"
+```
+
 ### How to setup Ray cluster over spark cluster ?
 
 A spark cluster is like:
@@ -38,13 +47,13 @@ will be executed concurrently, and each task is allocated with several resources
 one task is allocate with 1 cpu core. If we want to setup a Ray cluster with  16 cpu cores resources
 in total, we can create a spark barrier mode with 16 tasks.
  
-- Inside each spark job task, launch the Ray Node (head / worker) and make the task run until
+- Inside each spark executor, launch one Ray Node (head / worker) and keep the spark job running until
 the Ray cluster destroyed. Note: one spark executor might contain multiple spark tasks,
 but we will only launch one ray node for each spark executor.
 e.g., suppose in executor1, there’re 3 spark tasks launched, then in task of local-rank-0,
 we launch a ray node with –num-cpus=3. Using the following commands to launch ray node in spark tasks:
   - **Head node:** ray start —head --num-cpus=X --num-gpus=Y --memory=Z
-  - **Worker node:** ray start —head --num-cpus=X --num-gpus=Y --memory=Z
+  - **Worker node:** ray start —head --num-cpus=X --num-gpus=Y --memory=Z --address={head_ip}:{port}
 
  
 - After Ray cluster launched, the Ray application can be submitted to the ray cluster via
@@ -72,7 +81,7 @@ So, I suggest to launch Ray head node on spark task side, we can pick the first 
 head node.
 Issue: How to send the ray head node IP / port to spark driver side ?
 So that on spark driver side user can run  the ray application on the ray cluster.
-Because the spark barrier job keeps running when the ray cluster is up, we cannot user
+Because the spark barrier job keeps running when the ray cluster is up, we cannot use
 `rdd.collect()` to send the head node IP / port  back to driver side, we need to broadcast
 the spark driver IP to task side and then in spark task, we can send the  head node IP / port
 back to driver node by TCP connection.
@@ -103,12 +112,11 @@ environment so that we can restrict Ray node only uses the GPUs allocated to cor
 #### Initialize ray cluster on spark API
 
 ```
-ray_cluster_address, ray_cluster_handler = ray.spark.init_cluster(num_spark_tasks)
+ray_cluster_address, ray_cluster_handler = ray.spark.init_cluster(num_cpus, num_gpus, memory)
 ```
 
-Init a ray cluster on the spark cluster, the total resources of the ray cluster equals to the total resources of the sum of `num_spark_tasks` spark tasks.
+Init a ray cluster on the spark cluster, the argumens specified the ray cluster can use how much cpus / gpus / memory.
 
-After this function called, the computing resources are booked from the spark cluster, and a ray cluster is set up, and a background spark job keep running until the ray cluster shutdown.
 
 Returns a `ray_cluster_address` string (Ray Head node IP / port) and a `ray_cluster_handler`
 
@@ -132,7 +140,8 @@ On databricks notebook, we can make databricks runtime automatically calls `ray_
 
 ## Compatibility, Deprecation, and Migration Plan
 
-N/A
+Support apache/spark >= 3.3 and latest ray version.
+
 
 ## Test Plan and Acceptance Criteria
 
--- a/reps/2022-09-19-ray-on-spark.md
+++ b/reps/2022-09-19-ray-on-spark.md
@@ -100,6 +100,16 @@ This corresponding to:
 spark.task.cpus config >= 4 and (SPARK_WORKER_MEMORY / TASK_SLOTS_PER_SPARK_WORKER) >= 10GB
 
 
+#### On a shared spark cluster, shall we let each user launch individual ray cluster, or one user can create a shared ray cluster and other user can also use it ?
+I think we can provide 2 API to support both cases.
+Shared mode:
+```ray.spark.get_or_create_shared_cluster```
+and
+```ray.spark.init_private_cluster```
+
+And if one shared mode ray cluster already exists, new ray cluster is not allowed to be launched.
+
+
 #### How to select the port number used by Ray node ?
 Ray node requires listening on several port, a spark cluster might be shared by many users,
 each users might setup their own ray cluster concurrently, to avoid port conflicts,
--- a/reps/2022-09-19-ray-on-spark.md
+++ b/reps/2022-09-19-ray-on-spark.md
@@ -89,8 +89,9 @@ Yes. Otherwise, the Ray cluster setup will be nondeterministic,
 and you could get very strange results with bad luck on the node sizing.
 
 
-#### how to deterministically config resources per node ?
+#### Does ray node 1:1 map with spark task or spark executor?
 In each spark task, launch one ray worker node. Each spark task has the same assigned resources shape.
+* All ray nodes will have homogeneous resources because spark task resource config must be homogeneous
 
 
 #### What's recommended ray node shape ?
@@ -121,6 +122,12 @@ if failed we can retry on another free port.
 Spark does not provide explicit API for getting task allowed memory,
 SPARK_WORKER_MEMORY / SPARK_WORKER_CORES * RAY_NODE_NUM_CPUS
 
+
+#### How to support Ray object store memory ?
+We can allocate part of the memory assigned to spark task to be ray object-store-memory.
+(30% percent by default)
+
+
 #### How to make ray respect spark GPU resource scheduling ?
 In spark task, we can get GPU IDs allocated to this task, so, when launching
 Ray node, besides specifying `--num-gpus` options, we need specify `CUDA_VISIBLE_DEVICES`
@@ -142,8 +149,13 @@ Currently no. This implementation relies on spark barrier mode job, which does n
 
 
 #### What’s the level of isolation spark provides ?
-Resources isolation.
-Each ray cluster are isolated and will use its own resources allocated to the corresponding spark job.
+Barrier mode spark task provides process-level isolation.
+No VM/container level isolation. Spark task does not support it.
+
+
+#### Custom resources support
+TPUs, specific accelerators, etc.
+Uncommon scenario, We can do it later?. Spark doesn't support this, spark can only schedule CPU, GPU, memory.
 
 
 ### API Proposal
--- a/reps/2022-09-19-ray-on-spark.md
+++ b/reps/2022-09-19-ray-on-spark.md
@@ -85,9 +85,12 @@ too much computing / memory resources of driver node.
 
 
 #### Shall we make ray worker node the same shape (assigned with the same resources amount) ?
-
 Yes. Otherwise, the Ray cluster setup will be nondeterministic,
 and you could get very strange results with bad luck on the node sizing.
+
+
+#### how to deterministically config resources per node ?
+In each spark task, launch one ray worker node. Each spark task has the same assigned resources shape.
 
 
 #### What's recommended ray node shape ?
--- a/reps/2022-09-19-ray-on-spark.md
+++ b/reps/2022-09-19-ray-on-spark.md
@@ -54,10 +54,8 @@ ray start —head --num-cpus=0
 This makes the Ray head node only works as a manager node, without worker node functionality.
 
 - Inside each spark executor, launch one Ray worker worker and keep the spark job running until
-the Ray cluster destroyed. Note: one spark executor might contain multiple spark tasks,
-but we will only launch one ray node for each spark executor.
-e.g., suppose in executor1, there’re 3 spark tasks launched, then in task of local-rank-0,
-we launch a ray node with –num-cpus=3. Using the following commands to launch ray worker node in spark tasks:
+the Ray cluster destroyed. Each ray worker node will be assigned with resources equal to the corresponding
+spark task (CPU / GPU / Memory). The command used to start ray node is:
 ```
 ray start —head --num-cpus=X --num-gpus=Y --memory=Z --address={head_ip}:{port}
 ```
@@ -86,6 +84,19 @@ to ray head node, so that we can launch ray head node in driver side, it won't c
 too much computing / memory resources of driver node.
 
 
+#### Shall we make ray worker node the same shape (assigned with the same resources amount) ?
+
+Yes. Otherwise, the Ray cluster setup will be nondeterministic,
+and you could get very strange results with bad luck on the node sizing.
+
+
+#### What's recommended ray node shape ?
+
+cpu cores >= 4 and memory >= 10 GB
+This corresponding to:
+spark.task.cpus config >= 4 and (SPARK_WORKER_MEMORY / TASK_SLOTS_PER_SPARK_WORKER) >= 10GB
+
+
 #### How to select the port number used by Ray node ?
 Ray node requires listening on several port, a spark cluster might be shared by many users,
 each users might setup their own ray cluster concurrently, to avoid port conflicts,
@@ -95,10 +106,7 @@ if failed we can retry on another free port.
 
 #### How much memory shall we allocate to Ray node service ( set via ray script --memory option) ?
 Spark does not provide explicit API for getting task allowed memory,
-I propose:
 SPARK_WORKER_MEMORY / SPARK_WORKER_CORES * RAY_NODE_NUM_CPUS
-But how to get SPARK_WORKER_MEMORY and SPARK_WORKER_CORES in spark task is an issue.
-
 
 #### How to make ray respect spark GPU resource scheduling ?
 In spark task, we can get GPU IDs allocated to this task, so, when launching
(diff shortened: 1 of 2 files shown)


WeichenXu123:
CC @jjyao @ericl Thanks!

jjyao (commented):
I think we need to discuss more about the APIs.

When users start a Ray cluster, for each node, they are able to specify several things: num_cpus, num_gpus, memory, object_store_memory, custom resources, storage, etc (see full options of `ray start`). From the proposed API, I don't see how those can be configured.

Also the proposal is adding the api to the Ray codebase, have we considered the opposite: adding the api to the spark codebase? Why do we choose one over the other?

scv119:
@WeichenXu123 in addition to @jjyao's comments, can you shed some light on the final picture of integration, i.e. how would a spark user (or a ray user) can make use of this integration? For example, it could be some simple e2e code snippet where spark and ray works together.

WeichenXu123:
> @WeichenXu123 in addition to @jjyao's comments, can you shed some light on the final picture of integration, where spark user (or a ray user) can make use of this integration? For example, it could be some simple e2e code snippet where spark and ray works together.

Do you mean some example code and prototype implementation code ?
I created a repo for this https://github.com/WeichenXu123/ray-on-spark but some APIs there haven't been updated.

reps/2022-09-19-ray-on-spark.md
```diff
@@ -0,0 +1,151 @@
...
+TODO
+
+### Shepherd of the Proposal (should be a senior committer)
+
+TODO
+
+## Design and Architecture
+
+### How to setup Ray cluster over spark cluster ?
+
```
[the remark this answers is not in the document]
WeichenXu123: I add a `### Preparation` section

reps/2022-09-19-ray-on-spark.md
```diff
@@ -0,0 +1,151 @@
...
+### API Proposal
+
+#### Initialize ray cluster on spark API
+
+```
+ray_cluster_address, ray_cluster_handler = ray.spark.init_cluster(num_spark_tasks)
+```
+
+Init a ray cluster on the spark cluster, the total resources of the ray cluster equals to the total resources of the sum of `num_spark_tasks` spark tasks.
```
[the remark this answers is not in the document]
WeichenXu123: Changed API.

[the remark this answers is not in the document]
WeichenXu123: We can know each spark task can use how many cpus / gpus /memory by spark cluster config
So we can control how much resources ray cluster can use by controlling num_spark_tasks of the spark job.

reps/2022-09-19-ray-on-spark.md
```diff
@@ -0,0 +1,160 @@
...
+one task is allocate with 1 cpu core. If we want to setup a Ray cluster with  16 cpu cores resources
+in total, we can create a spark barrier mode with 16 tasks.
+ 
+- Inside each spark executor, launch one Ray Node (head / worker) and keep the spark job running until
+the Ray cluster destroyed. Note: one spark executor might contain multiple spark tasks,
+but we will only launch one ray node for each spark executor.
+e.g., suppose in executor1, there’re 3 spark tasks launched, then in task of local-rank-0,
+we launch a ray node with –num-cpus=3. Using the following commands to launch ray node in spark tasks:
+  - **Head node:** ray start —head --num-cpus=X --num-gpus=Y --memory=Z
+  - **Worker node:** ray start —head --num-cpus=X --num-gpus=Y --memory=Z --address={head_ip}:{port}
```
ericl, line 56: This part is a little strange, it seems Ray nodes end up with potentially different / random CPUs assigned depending on how Spark happened to schedule the tasks. In many use cases, users may want to have nodes with specific numbers of resources (e.g., 8-CPU nodes, or 4-GPU nodes).

Would the API allow for configuration of that kind of cluster?

reps/2022-09-19-ray-on-spark.md
```diff
@@ -0,0 +1,160 @@
...
+
+Init a ray cluster on the spark cluster, the argumens specified the ray cluster can use how much cpus / gpus / memory.
+
+
+Returns a `ray_cluster_address` string (Ray Head node IP / port) and a `ray_cluster_handler`
+
+
+#### Initialize ray client API.
+
+Now User can run ray application on the returned ray cluster, by executing `ray.init(address=ray_cluster_address)`, and then run any ray application code.
```
ericl, line 126: It would be preferable to avoid Ray client (use Ray direct connection), at least by default. This is for improved performance with data transferred to the Ray driver.

WeichenXu123: > It would be preferable to avoid Ray client (use Ray direct connection)

Do you have example code or document link about this ? :)

reps/2022-09-19-ray-on-spark.md
WeichenXu123: >  it seems Ray nodes end up with potentially different / random CPUs assigned depending on how Spark happened to schedule the tasks. 

No. The resources amount allocated is deterministic.

e.g., your case: (8-CPU nodes, or 4-GPU nodes), 
suppose on a spark cluster with config:
```
spark.task.cpus 2 # means 2 cpu cores per spark task
spark.task.resource.gpu.amount 1 # means 1 GPU per spark task
```

Then the Ray on spark routine will create a spark job with **4 spark tasks** running concurrently, which means it books (8-CPU nodes, or 4-GPU nodes) resources for ray cluster.


One issue is the CPU / GPU cores ratio depends on spark cluster config, this is a limitation.

reps/2022-09-19-ray-on-spark.md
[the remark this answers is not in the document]
WeichenXu123: See https://github.com/ray-project/enhancements/pull/14#discussion_r974771869

I will refine reps the section too.

reps/2022-09-19-ray-on-spark.md
jjyao: Cluster level total resources is not enough. We need to control at the per node level. Otherwise, some tasks may not be able to run if there is no single node that has the enough resources.

WeichenXu123: Got it.
What's the minimal resources requirements for a Ray Node ?

WeichenXu123: This might be a limitation on spark cluster,

We cannot control the spark task distribution on spark cluster, it is possible that one executor only launch one spark task, in this case, the corresponding Ray Node will get resources amount equal to that one spark task gets.

WeichenXu123: But, we can increase per-spark-task resources config , by increase config of:
```
spark.task.cpus
spark.task.resource.gpu.amount
``

File: reps/2022-09-19-ray-on-spark.md (hunk 4 above)
Comment (ericl) [reply reps/2022-09-19-ray-on-spark.md:126]: If `ray start` has been run on a node, then `ray.init()` will automatically discover and connect to the raylet on that node. This works on any node (head or worker).

File: reps/2022-09-19-ray-on-spark.md (hunk 3 above)
Comment (ericl) [reply reps/2022-09-19-ray-on-spark.md:56]: Setting spark.task.cpus/gpus sounds like the way to go. It sounds like this can be set dynamically per Spark job? That's probably the ideal case, though it would limit you to Ray clusters of homogeneous node shape.

The API could look like `ray.spark.init_cluster(num_nodes=100, num_cpus_per_node=16, num_gpus_per_node=4)`

Comment (WeichenXu123) [reply reps/2022-09-19-ray-on-spark.md:56]: > It sounds like this can be set dynamically per Spark job?

Unfortunately this is spark cluster level setting, we cannot set it dynamically per spark job.

File: reps/2022-09-19-ray-on-spark.md (hunk 4 above)
Comment (WeichenXu123) [reply reps/2022-09-19-ray-on-spark.md:126]: > If ray start has been run on a node, then ray.init() will automatically discover and connect to the raylet on that node. 

Sounds good!
Would you tell me how it is implemented ?

**Comment by WeichenXu123** [comment]
Discussion list:

1. how to deterministically  config resources per node
Seems we cannot , related discussion https://github.com/ray-project/enhancements/pull/14#discussion_r974606624

2. how to support all options of ray start
Provide a `ray_node_options` argument (dict type) ? 

3. where should the code live: ray repo or spark repo
To be decided.

more advanced questions are:
1. does it support autoscaling
Currently no. This implementation relies on spark barrier mode job, which does not support autoscaling currently.

2. what’s the level of isolation spark provides:
Resources isolation. Each ray cluster are isolated and will use its own resources allocated to the corresponding spark job.
What else isolation do you need ?

File: reps/2022-09-19-ray-on-spark.md (hunk 3 above)
Comment (WeichenXu123) [reply reps/2022-09-19-ray-on-spark.md:56]: > Ray clusters of homogeneous node shape

I have a question @ericl 

For a ray cluster, suppose on a worker machine, we launch multiple ray node instances, shall we merge it into one ray node ? 
I prefer to merge it into one ray node so that it is seemingly beneficial for performance.

Comment (WeichenXu123) [reply reps/2022-09-19-ray-on-spark.md:56]: Another issue is, because we setup ray cluster on spark cluster, for each spark job, it's spark tasks distribution might be imbalanced, e.g.

A spark cluster with 2 worker machine (worker_1 and worker_2), each worker machine has 4 cpu cores. The spark cluster is shared with user A and user B.

Now, suppose user_A launch a spark job-A, job_A generates 4 spark tasks, 3 spark tasks launched in worker_1, and 1 spark task launched in worker_2.

At the same time, user_B launch a second spark job_B (job B is used to run a ray cluster), then the job_B generates 1 spark tasks  in worker_1, 3 spark tasks in worker_2.

For job_B, the ray cluster, we can get 1 cpu cores in worker_1, and 3 cpu cores in worker_2,
in my proposal ray on spark design, I will start a ray worker node in worker_1 with num-cpus=1 , and start a ray worker node in worker_2 with num-cpus=3  @ericl WDYT ?

Or you think we should start 3 ray worker nodes (each with num-cpus=1) in worker_2 ?

Comment (ericl) [reply reps/2022-09-19-ray-on-spark.md]: > What's the minimal resources requirements for a Ray Node ?

Typically we don't recommend fewer than 8 CPUs / 30GB per node. Maybe 4 CPUs is a reasonable minimum.

> For a ray cluster, suppose on a worker machine, we launch multiple ray node instances, shall we merge them into one ray node ?

I think this would be generally preferable, yes, since Ray works better with fewer large nodes.

> in my proposal of ray on spark design, it will start a ray worker node in worker_1 with num-cpus=1 , and start a ray worker node in worker_2 with num-cpus=3 @ericl WDYT ?

It could work, but I would see it causing unexpected issues. For example, it could accidentally create Ray clusters with many tiny nodes, or sometimes few big nodes, even when running on the same Spark cluster, just due to how other workloads happen to be occupying executors. This would make the performance / reliability of Ray-on-Spark workloads unpredictable in shared clusters.

It seems like it would be preferable to only support non-shared clusters, if it could guarantee more consistently predictable cluster setups.

Btw, there's not just cpus to be allocated; there's object_store_memory, which is typically set to ~30% of the node space. You'd need to make sure this is set proportionally as well to avoid over-loading the node.

File: reps/2022-09-19-ray-on-spark.md (hunk 4 above)
Comment (ericl) [reply reps/2022-09-19-ray-on-spark.md]: The Ray cluster is discovered by checking the file `/tmp/ray/ray_current_cluster`

File: reps/2022-09-19-ray-on-spark.md (hunk 3 above)
Comment (WeichenXu123) [reply reps/2022-09-19-ray-on-spark.md]: @ericl 

> It seems like it would be preferable to only support non-shared clusters

Customer's spark cluster might also running other workloads, this is common cases, so we should support shared clusters as well.

> it could accidentally create Ray clusters with many tiny nodes, or sometimes few big nodes

For preventing tiny nodes, we can increase configs `spark.task.cpus` and `spark.task.resource.gpu.amount`, so that it can set the lower bound of the node shape.

For preventing big nodes, we can split big ray node into multiple ray nodes.

> Btw, there's not just cpus to be allocated; there's object_store_memory, which is typically set to ~30% of the node space

When increasing `spark.task.cpus` and `spark.task.resource.gpu.amount`, each spark worker generates less spark tasks slots, it means each task can be allocated with more memories. We can provide helper method to tell user the optimal `spark.task.cpus` and `spark.task.resource.gpu.amount` setting satisfying Ray node requirements according to the spark worker memory capacity.

File: reps/2022-09-19-ray-on-spark.md (hunk 4 above)
Comment (WeichenXu123) [reply reps/2022-09-19-ray-on-spark.md]: @ericl 
Issue: If the Ray cluster is discovered by checking the file /tmp/ray/ray_current_cluster,
suppose in a shared spark cluster, multiple Ray cluster is launched (by different users), and Ray cluster head nodes are all launched in spark driver node, then they share the same `/tmp/ray/ray_current_cluster` file, how to make each user connect to their own ray cluster ?

File: reps/2022-09-19-ray-on-spark.md (hunk 3 above)
Comment (ericl) [reply reps/2022-09-19-ray-on-spark.md]: Requiring the user to increase the task config size can work. I think it would be best to require the user to set these values to configure the Ray node size, and then always launch one node per task. Maybe we can also require the node to have at least 4 CPUs or raise an error.

Otherwise, the Ray cluster setup will be nondeterministic, and you could get very strange results with bad luck on the node sizing.

File: reps/2022-09-19-ray-on-spark.md (hunk 4 above)
Comment (ericl) [reply reps/2022-09-19-ray-on-spark.md]: If there are multiple Ray clusters, then you'd have to implement discovery manually. This could be done by tracking the gcs port of the head node, and then passing that port explicitly to `ray.init(address=ip:port)`.

Comment (WeichenXu123) [reply reps/2022-09-19-ray-on-spark.md]: > tracking the gcs port of the head node, and then passing that port explicitly to ray.init(address=ip:port).

yep, this is the approach used in my prototype code

File: reps/2022-09-19-ray-on-spark.md (hunk 3 above)
Comment (WeichenXu123) [reply reps/2022-09-19-ray-on-spark.md]: > Otherwise, the Ray cluster setup will be nondeterministic, and you could get very strange results with bad luck on the node sizing

Do you mean different shape Ray worker nodes causing this ? If so, we can launch one ray node per spark task to prevent the issue.

Comment (WeichenXu123) [reply reps/2022-09-19-ray-on-spark.md]: @ericl 

Btw, I will set Ray head node with `--num-cpus=0` (prevent ray task scheduled to head node), and then set all ray worker nodes with the same shape. Does the head node different shape have some influence ?

Comment (ericl) [reply reps/2022-09-19-ray-on-spark.md]: That makes sense if you always put the head node on the Spark driver node. If the head node can live on other nodes, then I would consider setting num-cpus to be the same as the other nodes.

**Comment by WeichenXu123** [comment]
Support ray object memory.

**Comment by WeichenXu123** [comment]
Q: minimal ray node memory requirement.

**Comment by WeichenXu123** [comment]
Update REP: no VM/container level isolation

**Comment by WeichenXu123** [comment]
To be confirmed CC @jjyao @ericl 

1. Default memory ratio for Ray object store memory ? 
Proposal: 30% of total spark task available memory, and capped with 20GB 

2. Shall we allocate memory for --redis-max-memory ?

3. In ray on spark implementation, it starts ray node via CLI (ray start command), In my prototype code, I use `ray start --block ...` , and when shutdown node, I sent a termination signal (sigterm) to `ray start ` command process. Is it the correct way to shutdown the node ?

**Comment by ericl** [comment]
> Proposal: 30% of total spark task available memory, and capped with 20GB

Hmm is this per-task or per-node? Calculating this value manually can be tricky. If you exceed the /dev/shm size, this can result in segfaults. Using non-/dev/shm storage can cause swapping.

I would recommend something like:
- Calculating the available /dev/shm space.
- Dividing this by the number of Spark tasks that can run per node.

> Shall we allocate memory for --redis-max-memory ?

Ray no longer uses redis, so this flag does nothing.

> In ray on spark prototype implementation, it starts ray node via CLI (ray start command), In my prototype code, I use ray start --block ... , and when shutdown node, it sends a termination signal (sigterm) to ray start command process. Is it the correct way to shutdown the node ?

This sounds reasonable.

> What's the minimal recommended memory for ray worker node ? Is it 10GB ?

Also seems reasonable.

**Comment by WeichenXu123** [comment]
> Hmm is this per-task or per-node? Calculating this value manually can be tricky. If you exceed the /dev/shm size, this can result in segfaults. Using non-/dev/shm storage can cause swapping.
> 
> I would recommend something like:
> 
> Calculating the available /dev/shm space.
> Dividing this by the number of Spark tasks that can run per node, then scaling by say 0.8 for safety.
> 


1. This is per ray worker node.

2. `Dividing this by the number of Spark tasks that can run per node, then scaling by say 0.8 for safety.` sounds good.

3. Shall we config it to use non-/dev/shm memory as a fallback if it exceeds /dev/shm ? and how to config ?

**Comment by WeichenXu123** [comment]
@ericl 

I found a ray environment config : `RAY_OBJECT_STORE_ALLOW_SLOW_STORAGE`, if set it to 1, does it mean ray node use /dev/shm preferentially, and if exceeding /dev/shm capacity, fallback to use low storage (disk) ?

**Comment by jjyao** [comment]
@WeichenXu123 
> I found a ray environment config : RAY_OBJECT_STORE_ALLOW_SLOW_STORAGE, if set it to 1, does it mean ray node use /dev/shm preferentially, and if exceeding /dev/shm capacity, fallback to use low storage (disk) ?

Yes.
```
# Determine which directory to use. By default, use /tmp on MacOS and
# /dev/shm on Linux, unless the shared-memory file system is too small,
# in which case we default to /tmp on Linux.
```

**Comment by ericl** [comment]
Be warned that using /tmp for object store memory can cause extreme slowdowns. The reason is that the operating system will typically try to write back pages in /tmp to the filesystem at some point, and application memory accesses can trigger page loads from the filesystem. When Ray writes tens of GB of data, we've observed high levels of swapping and also system hangs.

It would be better to set the memory size smaller so it fits into /dev/shm, or otherwise configure /dev/shm to be large enough for the object store. Setting a smaller object store size just means Ray will spill data when object store memory runs out.

**Comment by ericl** [comment]
>  if set it to 1, does it mean ray node use /dev/shm preferentially, and if exceeding /dev/shm capacity, fallback to use low storage (disk)

To answer the question directly, that's not what this flag does. Ray will already spill to disk on exceeding /dev/shm memory. You should never set this flag in production, only for local testing purposes on small scales.

File: reps/2022-09-19-ray-on-spark.md [hunk 5]
```diff
@@ -0,0 +1,224 @@
...
+one task is allocate with 1 cpu core. If we want to setup a Ray cluster with  16 cpu cores resources
+in total, we can create a spark barrier mode with 16 tasks.
+ 
+- In spark driver side, launch Ray head node, by:
+```
+ray start —head --num-cpus=0
+```
+This makes the Ray head node only works as a manager node, without worker node functionality.
+
+- Inside each spark executor, launch one Ray worker worker and keep the spark job running until
```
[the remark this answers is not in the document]
dbczumar: Fixed in https://github.com/WeichenXu123/enhancements/pull/1 - thanks!

dbczumar (commented):
@jjyao @ericl I've filed https://github.com/WeichenXu123/enhancements/pull/1 against @WeichenXu123's PR branch to address your latest round of comments and suggestions. I've reached out to @WeichenXu123 requesting permission to edit his branch directly. Until then, can you take a look at https://github.com/WeichenXu123/enhancements/pull/1 and let me know if additional changes are necessary in order to get this REP approved? Thanks for your help!

cc @BenWilson2

reps/2022-09-19-ray-on-spark.md
```diff
@@ -0,0 +1,258 @@
...
+Ray node, besides specifying `--num-gpus` options, we need specify `CUDA_VISIBLE_DEVICES`
+environment so that we can restrict Ray node only uses the GPUs allocated to corresponding spark tasks.
+
+
+#### How to support all options of ray start
+Provide a ray_node_options argument (dict type).
+
+
+#### Where should the code live: ray repo or spark repo?
+Ray repo.
```
scv119, line 176: Does it make sense to set up a separate repo, instead of using main ray repro?

scv119:
Also have we considered creating a customized spark node provider? https://docs.ray.io/en/master/cluster/vms/user-guides/community/index.html#using-a-custom-cloud-or-cluster-manager

reps/2022-09-19-ray-on-spark.md
BenWilson2: HI @scv119,
I work with Weichen and Corey. 
We definitely initially entertained the idea of creating a plugin provider to enable Ray support for Apache Spark. However, there are a few reasons why we think that might dramatically curtail the usage of such an integration. With the large volume of requests we've had in the OSS community for enabling a fully-supported integration, providing the community with a 3rd party (or 2nd party) plugin creates the impression of flippancy and increases the amount of effort that users would have to perform to get this much-asked-for functionality.
To whit, we'd like to see a true 1st party integration that is fully maintained by our Eng team. While having the integration as a plugin would certainly be simpler in implementation, we feel that the perception of support (most plugins aren't well maintained and that blanket perception generally extends to all, regardless of reality), as well as the additional feature parity / versioning complexity that would arise from maintaining a dependency between core Ray and a plugin would result in the solution in becoming effective shelf-ware. 
That being said, we understand and fully appreciate the gravity of such an integration living within the main Ray repo. Be assured that we would handle and assume ownership of the integration in the same way that we treat our own OSS projects with.  That being a production-level inclusion to our on-call staff members to ensure that issues, feature requests, and CI failures would be handled no differently than how we respond to Spark or MLflow addressable items. 

Hopefully that expresses our general thoughts on this. 
Kindest regards,
Ben

dbczumar: Hi @scv119 , @ericl , @jjyao Any updates here?

WeichenXu123: Q: Does ray team also helps maintaining projects like kuberay / xgboost_ray ? If these separate ray projects are maintained well, we could consider to make spark_ray as a separate project.

WeichenXu123:
> Also have we considered the alternative of creating a customized spark node provider? https://docs.ray.io/en/master/cluster/vms/user-guides/community/index.html#using-a-custom-cloud-or-cluster-manager

Haven't.
In current REP, the main use-cases is that user want to run ray on databricks runtime.

reps/2022-09-19-ray-on-spark.md
dbczumar: > Q: Does ray team also helps maintaining projects like kuberay / xgboost_ray ? If these separate projects are maintained well, we could consider to make spark_ray as a separate project. A separate project has a benefit that we can build and locally run tests easier, otherwise we have to build the whole ray project, which is complex.

As @BenWilson2 , I would prefer to structure this as a contribution to the Ray repository to ensure full support and centralized communication / collaboration, especially since a lot of Spark users are asking for this.

@scv119 @ericl @jjyao What do we need to do to make a decision here? :)

ericl: I think this would also be better for the Ray repository, especially if we want to test and maintain it as a first class integration.

ericl:
> Also have we considered the alternative of creating a customized spark node provider? https://docs.ray.io/en/master/cluster/vms/user-guides/community/index.html#using-a-custom-cloud-or-cluster-manager

I believe this suggestion is that instead of using Spark's barrier mode to launch fixed-size Ray clusters, we could write a `SparkNodeProvider` plugin for Ray's autoscaler. You could implement this by mapping each Ray node request to a Spark job. So when the Ray cluster tries to autoscale up, the SparkNodeProvider could launch a Spark job to add Ray nodes.

Pros:
- Supports Ray autoscaling.

Cons:
- May generate a large number of Spark jobs (lots of clutter in the UI), less easy to understand what's going on.

Unless autoscaling is strictly needed, my feeling here is it may be better to start with barrier mode execution.

WeichenXu123:
> Unless autoscaling is strictly needed, my feeling here is it may be better to start with barrier mode execution.

I agree, we can start with barrier mode execution.

> Cons: May generate a large number of Spark jobs (lots of clutter in the UI), less easy to understand what's going on.

We can avoid the cons by: starting minimal requested ray nodes by spark barrier mode job, then if scaling up requested, then starting new ray node by separate spark jobs.

dbczumar:
> > Also have we considered the alternative of creating a customized spark node provider? https://docs.ray.io/en/master/cluster/vms/user-guides/community/index.html#using-a-custom-cloud-or-cluster-manager
> 
> I believe this suggestion is that instead of using Spark's barrier mode to launch fixed-size Ray clusters, we could write a `SparkNodeProvider` plugin for Ray's autoscaler. You could implement this by mapping each Ray node request to a Spark job. So when the Ray cluster tries to autoscale up, the SparkNodeProvider could launch a Spark job to add Ray nodes.
> 
> Pros:
> 
> * Supports Ray autoscaling.
> 
> Cons:
> 
> * May generate a large number of Spark jobs (lots of clutter in the UI), less easy to understand what's going on.
> 
> Unless autoscaling is strictly needed, my feeling here is it may be better to start with barrier mode execution.

@ericl Thank you for your analysis. That makes perfect sense, and we're happy to stick with barrier mode initially.

@ericl @jjyao @scv119 Do you have any other blocking feedback? We are very excited to get the REP approved and move forward with the implementation :D

dbczumar:
> > > Also have we considered the alternative of creating a customized spark node provider? https://docs.ray.io/en/master/cluster/vms/user-guides/community/index.html#using-a-custom-cloud-or-cluster-manager
> > 
> > 
> > I believe this suggestion is that instead of using Spark's barrier mode to launch fixed-size Ray clusters, we could write a `SparkNodeProvider` plugin for Ray's autoscaler. You could implement this by mapping each Ray node request to a Spark job. So when the Ray cluster tries to autoscale up, the SparkNodeProvider could launch a Spark job to add Ray nodes.
> > Pros:
> > 
> > * Supports Ray autoscaling.
> > 
> > Cons:
> > 
> > * May generate a large number of Spark jobs (lots of clutter in the UI), less easy to understand what's going on.
> > 
> > Unless autoscaling is strictly needed, my feeling here is it may be better to start with barrier mode execution.
> 
> @ericl Thank you for your analysis. That makes perfect sense, and we're happy to stick with barrier mode initially.
> 
> @ericl @jjyao @scv119 Do you have any other blocking feedback? We are very excited to get the REP approved and move forward with the implementation :D

@ericl @jjyao @scv119 @zhe-thoughts Following up here again. Is there any additional feedback, or can the REP be approved as-is?

reps/2022-09-19-ray-on-spark.md
```diff
@@ -0,0 +1,258 @@
...
+
+Yes. So every user only need to start ray cluster once and then user can run ray applications
+on the ray cluster with little overhead.
+
+
+#### Launch Ray head node on spark driver node or spark task side ?
+In spark driver side.
+Because we can set Ray head node with `num-cpus` = 0, so that no ray task will be scheduled
+to ray head node, so that we can launch ray head node in driver side, it won't consume
+too much computing / memory resources of driver node.
```
jjyao, line 107: Based on the discussion, while the ray head node won't run any application code, it still runs critical system processes so we should make sure it runs in a production machine instead of like user's laptop.

reps/2022-09-19-ray-on-spark.md
```diff
@@ -0,0 +1,258 @@
...
+ray_cluster_on_spark = ray.spark.init_cluster(num_spark_tasks)
+```
+
+Initialize a ray cluster on the spark cluster, the arguments specified the ray cluster can use how much cpus / gpus / memory.
+And connect to the cluster.
+
+Returns an instance of type `RayClusterOnSpark``
+
+
+For each ray node, it might have different / random CPUs assigned ,
```
jjyao, line 215: This is not true any more: the spark task resource assignment is fixed.

reps/2022-09-19-ray-on-spark.md
```diff
@@ -0,0 +1,258 @@
...
+
+4. After the Ray cluster is launched, the user's Ray application(s) can be submitted to the Ray
+cluster via the Ray head node address / port.
+
+5. To shut down the Ray cluster, cancel the Spark barrier mode job and terminate all Ray node
+services that are running on the Spark driver and Spark workers.
+
+Finally, by adjusting the resource allocation for Ray nodes and Spark tasks, this approach enables
+users to run multiple Ray clusters in isolation of each other on a single Spark cluster, if desired.
+
```
jjyao, line 94: Let's also describe fault tolerance as well: what will happen if a spark task (ray worker node) fails.

jjyao (commented):
LGTM for the high level design.

reps/2022-09-19-ray-on-spark.md
```diff
@@ -0,0 +1,258 @@
+## Summary
```
jjyao, line 1: For NodeProvider approach, I think we need to do some prototype in order to better understand it. Based on my current understanding, the pros and cons are:

Pros:
    1.  The same and consistent way to config nodes via yaml. For example we can specify different custom resources for each node type. We can also config docker for Ray nodes. (Those are not possible with the current `init_cluster()` api so it's more future looking).
    2.  Supports autocaling. (This is also something not supported by current proposal).
    
  Cons:
       1. The current NodeProvider abstraction doesn't match our model very well. For example, NodeProvider separates getting physical node and starting ray node while this design does both via a single spark barrier job. Also NodeProvider api creates one node at a time while ray-on-spark needs to create the whole cluster at once. I think it's still doable but it will be more complicated than the `init_cluster()` approach due to the need to conform to the NodeProvider api.
       
While NodeProvider approach might be more future-proof, the benefit probably doesn't justify the extra complexity for now and it has more unknowns without a prototype. We can probably just start with `init_cluster()` approach for now. In the future, we can always switch to the NodeProvider approach if needed.

Thoughts @ericl @scv119 @WeichenXu123 ?

reps/2022-09-19-ray-on-spark.md
BenWilson2: Excellent point. I'll update the REP and ensure that we perform validation on an instance with minimum available specs (similar to the change that we just pushed to the PR that performs a worker-level available resources check)

reps/2022-09-19-ray-on-spark.md
BenWilson2: Definitely. Will update once I get authorized access to the branch.

reps/2022-09-19-ray-on-spark.md
BenWilson2: I'll add a paragraph about the lack of fault tolerance in barrier execution mode and how entire tasks would need to retry based on the BarrierRDD implementation in Spark.

reps/2022-09-19-ray-on-spark.md
BenWilson2: We agree that a NodeProvider approach is appealing and definitely something to strive towards if users struggle with the barrier execution mode. We're a bit reluctant to move down that path initially though, mostly for the cons that you've listed so well.
In the interests of time and effort, implementing even a prototype would likely be leveraging internal Databricks functionality to get auto-scaling to control the initialization of worker nodes, perform health checks, and restart dropped nodes. We'd like to avoid such an implementation since many of those features are not part of OSS Spark and we'd like to avoid restricting users to the Databricks platform to use this integration. 
However, we'd still like to think about how to extend this functionality and properly scope this work after an initial best-effort implementation that utilizes barrier execution mode is done. If the need arises to build such advanced functionality, we can collaborate on a design and implementation that serves this use case in the future. 

Additional thoughts and comments @ericl @scv119 @dbczumar @WeichenXu123 ?

dbczumar: @BenWilson2 I completely agree. Echoing the assessment from @jjyao and @ericl, it seems very reasonable to start with the existing proposal which is clearly OSS-compatible and investigate a node provider solution in the coming months. I concur that there are some difficult questions that would need to be answered in order to integrate Ray's node provider abstraction with Spark. @ericl @scv119 @zhe-thoughts @jjyao do we have alignment here?

reps/2022-09-19-ray-on-spark.md
WeichenXu123: yes barrier mode spark job lack of fault tolerance,
we can add a loop with retry logic inside BarrierRDD mapper function as the workaround.
Note the workaround does not support fault tolerance when the physical machine crashed.

reps/2022-09-19-ray-on-spark.md
WeichenXu123: We can support NodeProvider in next step. But I think databricks spark cluster might be incompatible with NodeProvider.

jjyao: sgtm.

jjyao (approved):

reps/2022-09-19-ray-on-spark.md
```diff
@@ -0,0 +1,286 @@
...
+``SPARK_WORKER_MEMORY`` / ``TASK_SLOTS_PER_WORKER`` under the assumption that the
+``spark.task.cpus`` configuration is set to ``4``).
+
+3. In each constituent Spark task, launch one Ray worker node in blocking mode and allocate to it
+the full set of resources available to the Spark task (4 CPUs, 10GB of memory). Keep the Spark task
+running until the Ray cluster destroyed. The command used to start each Ray worker node is as
+follows:
+
+```
+ray start —head --num-cpus=X --num-gpus=Y --memory=Z --address={head_ip}:{port}
```
ericl, line 83:
```suggestion
ray start --num-cpus=X --num-gpus=Y --memory=Z --object-store-memory=M --address={head_ip}:{port}
```

ericl (approved):

reps/2022-09-19-ray-on-spark.md
```diff
@@ -0,0 +1,286 @@
...
+cores can be launched on a Spark cluster, assuming that each Ray node requires 4 CPU cores and
+10GB of memory:
+
+1. On the Spark driver node, launch the Ray head node as follows:
+
+```
+ray start --head --num-cpus=0
+```
+
+The Ray head node will operate solely as a  manager node, without worker node functionality.
```
rkooo567, line 68: Note that this is not strictly guaranteed. For example, actors use 0 CPU by default, and they can still be scheduled on this node.

reps/2022-09-19-ray-on-spark.md
```diff
@@ -0,0 +1,286 @@
...
+  ```
+
+### How to setup Ray cluster over spark cluster ?
+
+The architecture of a Spark cluster is as follows:
+![spark-cluster-overview](https://spark.apache.org/docs/latest/img/cluster-overview.png)
+
+Using Spark's barrier mode, we intend to represent each Ray worker node as a long-running Spark
+task, where the Ray worker node has the same set of resources as the Spark task. Multiple Spark
+tasks can run on a single Spark worker node, which means that multiple Ray worker nodes can run
```
rkooo567, line 54: > Multiple Spark
tasks can run on a single Spark worker node, which means that multiple Ray worker nodes can run
on a single Spark worker node.

Is this a requirement? Is it possible to promote 1 node <> 1 ray instance as the best practice? It technically works, but Ray may not play well with this setup sometimes (more than one runtime instance per node). For example, many dashboard features rely on node IP. There are some other hiccups we don't expect (like we rely on some files inside /tmp/ray).

rkooo567 (commented):
These comments are not a blocker

reps/2022-09-19-ray-on-spark.md
```diff
@@ -0,0 +1,291 @@
...
+cores can be launched on a Spark cluster, assuming that each Ray node requires 4 CPU cores and
+10GB of memory:
+
+1. On the Spark driver node, launch the Ray head node as follows:
+
+```
+ray start --head --num-cpus=0
+```
+
+The Ray head node will operate configured to optimistically not execute raylets due to combination of
```
jjyao, line 72: What does this mean? Raylet is a process that each ray node has (including head node). It's just when we set num-cpus to 0 for head node, all tasks with cpu requirements won't be able to run on it.

BenWilson2: I'll revise :)

reps/2022-09-19-ray-on-spark.md
```diff
@@ -0,0 +1,270 @@
...
+
+#### Where should the code live: ray repo or spark repo?
+Ray repo.
+Putting in spark repo is hard, It would be very hard to pass vote.
+Past examples includes horovod, petastorm, xgboost, tensorflow-spark, all of them tried to put in spark repo but failed.
+
+
+#### Does it support autoscaling ?
+We will be investigating and testing the feasibility of supporting autoscaling of ray worker nodes as 
+additional spark task slots become available upon the initiation of additional spark worker nodes.
```
wuisawesome, line 185: Non-blocking question, but could a spark expert perhaps point me to some code or docs about how "task slots" are managed?

WeichenXu123: e.g., suppose a spark worker has 8 CPU cores, 4 GPU cores, and a spark task is configured by 2 cpus per task, and 1 gpu per task, then spark scheduler will calculate that the maximum task slots number is 4 on the spark worker.

reps/2022-09-19-ray-on-spark.md
WeichenXu123: > Is it possible to promote 1 node <> 1 ray instance as the best practice?

yes. this is the best practice.

We can config spark task shape to make only one spark task slots on each spark worker node, so that it will satisfy the best practice rule.

zhe-thoughts:
With the approvals on this PR, and the passing vote on ray-committers, merging this REP. Thanks @WeichenXu123 for the great contribution, and reviewers especially @jjyao !

PR #19, opened 2022-12-15
[REP] Execution Optimizer for Ray Datasets

c21:
This REP introduces (1) lazy execution, (2) optimizer, and (3) vectorized execution with data batch, to improve user experience and performance for Ray Datasets.

### DIFF
--- a/reps/2022-12-15-optimizer-data.md
+++ b/reps/2022-12-15-optimizer-data.md
@@ -1,0 +1,340 @@
+# Execution Optimizer for Ray Datasets
+
+## Summary
+
+Build the breakthrough foundation to tackle a series of fundamental issues around Ray Data. The foundation is (1) **lazy execution**, (2) **optimizer**, and (3) **vectorized execution with data batch**.
+
+### General Motivation
+Two key ML workloads for Datasets are (1) **data ingest**, where trainer processes (e.g., PyTorch workers), read, preprocess, and ingest data, and (2) **batch inference**, where we want to apply a pretrained model across a large dataset to generate predictions.
+
+Currently Ray Data does not work out-of-box for both workloads with followed reasons:
+* Data batch size and parallelism is challenging to tune by users, often resulting in poor performance and frustration.
+* Expensive data conversion, copy, and materialization to object store, can happen everywhere.
+* Lack of optimizer for execution plan, caused too much hand-holding with users to write the best program by hand.
+
+To resolve the above issues, we need to change Ray Datasets in multiple perspectives:
+* User experience:
+  * Configurations auto-tuning: auto-tuning critical configurations such as data batch size and parallelism. Fewer configurations and less confusing for users.
+  * Minimize cognitive load: lazy-execution first and only.
+* Performance
+  * Memory efficiency: zero-copy batch iterator. Operators pushdown. Eliminate unnecessary batch copy and format conversion.
+* Extensibility
+  * Easier to develop other optimization in the future
+    * Special optimizer for AIR workload.
+    * Example: prune unused columns through the execution plan as early as possible (as ColumnPruningPushDownRule).
+* Simplify code base
+  * Unify `LazyBlockList` and `BlockList`, delete `StatsActor`, eager mode and other legacy behavior.
+  * Able to support more Datasets APIs.
+
+### Should this change be within `ray` or outside?
+
+The proposed change is inside the code of Ray Datasets, which is already part of the Ray project repository (“ray.data”).
+
+## Stewardship
+### Required Reviewers
+
+@clarkzinzow, @jianoaix, @ericl, @stephanie-wang
+
+### Shepherd of the Proposal (should be a senior committer)
+
+@ericl
+
+## Design and Architecture
+
+Architecture after REP:
+
+<img width="945" alt="new-architecture" src="https://user-images.githubusercontent.com/4629931/207807703-bb65db63-41a0-41d9-8e7b-154e1a0ed565.png">
+
+Architecture before REP:
+
+<img width="930" alt="old-architecture" src="https://user-images.githubusercontent.com/4629931/207808050-c07e51ed-ece4-4781-9f97-5a06ac107973.png">
+
+### 1. Enable lazy execution mode by default
+
+#### 1.1. How do things work today?
+
+Ray Datasets has two execution modes:
+* Eager mode
+  * Each operation is executed immediately (i.e. eagerly).
+* Lazy mode
+  * Each operation is not executed immediately, but being added to the execution plan (i.e. lazily). Upon user calling consumption/action APIs (e.g. `fully_executed()`, `iter_batches()`, `take()` etc), all operations are executed together.
+
+See https://docs.ray.io/en/latest/data/dataset-internals.html#lazy-execution-mode for more information.
+
+Ray Datasets support both `Dataset` and `DatasetPipeline`.
+* `Dataset` is using eager mode by default. Lazy mode can be enabled by calling `Dataset.lazy()`.
+* `DatasetPipeline` is using lazy mode by default.
+
+#### 1.2. Proposed Change
+
+Change execution mode to lazy execution mode by default.
+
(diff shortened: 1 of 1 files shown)


reps/2022-12-15-optimizer-data.md
```diff
@@ -0,0 +1,308 @@
...
+
+### Shepherd of the Proposal (should be a senior committer)
+
+@ericl
+
+## Design and Architecture
+
+Architecture after REP:
+
+<img width="945" alt="new-architecture" src="https://user-images.githubusercontent.com/4629931/207807703-bb65db63-41a0-41d9-8e7b-154e1a0ed565.png">
```
ericl, line 46: Since we are considering re-optimization outside the scope of this REP, can we also remove that from the diagram?

reps/2022-12-15-optimizer-data.md
```diff
@@ -0,0 +1,308 @@
...
+
+Data flow:
+
+<img width="795" alt="data-flow-after" src="https://user-images.githubusercontent.com/4629931/207812234-d94a6683-5e0d-4eb2-a7c3-6f7bb656f2b2.png">
+
+The new proposal would eliminate that batch to block copy and pass the batch directly to the following stage.
+
+#### 3.2.1. Interfaces
+
+NOTE: `OneToOneOperator` used here is the same as `OneToOneOperator` in "Native pipelining support in Ray Datasets" REP.
```
ericl, line 210: This section needs to be updated, since the other REP now only proposes `PhysicalOperator`.

reps/2022-12-15-optimizer-data.md
```diff
@@ -0,0 +1,308 @@
...
+  A rule to optimize logical or physical plan. The rule should be cost-based and generate the plan with minimal cost.  
+  """
+  
+  def apply(plan: Plan, statistics) -> Plan:
+    # specific optimization logic in each rule
+```
+
+##### 2.2.2. Milestones
+
+* Step1: Introduce `Optimizer` class.
```
stephanie-wang, line 147: At this stage, we won't have any optimization rules based on run-time stats, right? Might be good to:
1. Simplify the proposal to remove anything that is only needed for run-time optimization.
2. Add a simple `Rule` code example to demonstrate interface and functionality.

reps/2022-12-15-optimizer-data.md
```diff
@@ -0,0 +1,308 @@
...
+
+class LogicalOptimizer(Optimizer):
+  """Optimizer for logical plan.
+
+  An optimizer transforms the logical plan, with a list of optimization Rules and input statistics.
+  """
+
+  rules: List[Rule]
+
+  def optimize(plan: LogicalPlan, statistics) -> LogicalPlan:
```
stephanie-wang, line 117: How do we go from LogicalPlan to PhysicalPlan, and what is the difference?

Might be good to have a figure and examples for a logical vs physical rule.

reps/2022-12-15-optimizer-data.md
```diff
@@ -0,0 +1,308 @@
...
+```
+
+Corner cases:
+* When users set different `batch_size` for multiple `map_batches()`. We can choose (1). respect users setting (will incur copy between `map_batches`), or (2). decides a min/max `batch_size` and uses the same `batch_siz`e across `map_batches`.
+  * Filter can happen to reduce rows
+  * Image rotation can change number of images
+  * Solution: have a `Batcher` to buffer batches for different `batch_size`s specified in `map_batches`.
+* When the user's `map_batches()` function changes the number of rows in a batch, and has multiple `map_batches()`. We can choose (1). respect users settings, and aim to produce a batch with `batch_size` for the following `map_batches()`, or (2).do nothing for the output batch, feed directly into following `map_batches()`.
+
+### 4. Benchmark
```
stephanie-wang, line 270: Nice microbenchmarks!

stephanie-wang (changes requested):

reps/2022-12-15-optimizer-data.md
c21: @ericl - yeah updated. I need more thought to hook up `BatchedOperator.process_batches` with `PhysicalOperator.add_input/inputs_done/has_next/get_next`. But I think it should be implementation detail that we can figure it out later.

c21: @ericl - yeah updated. I need more thought to hook up `BatchedOperator.process_batches` with `PhysicalOperator.add_input/inputs_done/has_next/get_next`. But I think it should be implementation detail that we can figure it out later.

reps/2022-12-15-optimizer-data.md
c21: > How do we go from LogicalPlan to PhysicalPlan, and what is the difference?

There would be a class/utility to convert `LogicalPlan` into `PhysicalPlan`. Given the current logic would be fairly straightforward, so leave this as implementation detail for now.

Added explanation for `LogicalPlan` and `PhysicalPlan` in Step 3 below.

> Might be good to have a figure and examples for a logical vs physical rule.

Added explanation and examples for logical and physical rule in Step 1 below.

reps/2022-12-15-optimizer-data.md
c21: > Simplify the proposal to remove anything that is only needed for run-time optimization.

Sure, removed. We probably still need input statistics though, so keep input statistics only.

> Add a simple Rule code example to demonstrate interface and functionality.

Added a simple pseudo code example for limit push down rule.

reps/2022-12-15-optimizer-data.md
```diff
@@ -0,0 +1,308 @@
...
+Two key ML workloads for Datasets are (1) **data ingest**, where trainer processes (e.g., PyTorch workers), read, preprocess, and ingest data, and (2) **batch inference**, where we want to apply a pretrained model across a large dataset to generate predictions.
+
+Currently Ray Data does not work out-of-box for both workloads with followed reasons:
+* Data batch size and parallelism is challenging to tune by users, often resulting in poor performance and frustration.
+* Expensive data conversion, copy, and materialization to object store, can happen everywhere.
+* Lack of optimizer for execution plan, caused too much hand-holding with users to write the best program by hand.
+
+To resolve the above issues, we need to change Ray Datasets in multiple perspectives:
+* User experience:
+  * Configurations auto-tuning: auto-tuning critical configurations such as data batch size and parallelism. Fewer configurations and less confusing for users.
```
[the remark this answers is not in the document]
c21: Yes, this is in the scope of REP.

reps/2022-12-15-optimizer-data.md
```diff
@@ -0,0 +1,308 @@
+# Execution Optimizer for Ray Datasets
+
+## Summary
+
+Build the breakthrough foundation to tackle a series of fundamental issues around Ray Data. The foundation is (1) **lazy execution**, (2) **optimizer**, and (3) **vectorized execution with data batch**.
```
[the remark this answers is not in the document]
c21: Moved this under `General Motivation` to make it more coherent, as the top-level summary seem not strictly needed (not see in other REPs).

reps/2022-12-15-optimizer-data.md
c21: yeah, removed.

stephanie-wang (approved):

reps/2022-12-15-optimizer-data.md
```diff
@@ -0,0 +1,323 @@
...
+@clarkzinzow, @jianoaix, @ericl, @stephanie-wang
+
+### Shepherd of the Proposal (should be a senior committer)
+
+@ericl
+
+## Design and Architecture
+
+**Architecture after REP**:
+
```
zhe-thoughts, line 45: Nit: can we change the order of these two diagrams ("before" first, and then "after"), and make them equally large? Right now the "after" is much smaller

c21: @zhe-thoughts - sure, updated.

PR #20, opened 2023-01-10
[REP] AWS Fleet support for ray autoscaler

raghumdani:
This REP introduces EC2 fleet support on ray autoscaler for AWS provider.


reps/2023-01-09-aws-fleet-support.md
```diff
@@ -0,0 +1,138 @@
...
+
+### How will the demand be converted into resources?
+
+Autoscaling monitor receives several demands from the metrics like vCPUs, memory, gpus etc. The [resource_demand_scheduler.py](https://github.com/ray-project/ray/blob/master/python/ray/autoscaler/_private/resource_demand_scheduler.py) is responsible for converting the demand into number of workers per node type specified in the autoscaler config. However, it relies on the **node provider** to statically determine the CPU, memory, GPU and any custom resources. Hence, **node_provider** will determine CPU, memory and GPU from the `InstanceRequirements` and specified `InstanceType` parameters and aggregated them based on the latest family or high spec instance. Hence, autoscaler will end up spinning less nodes than necessary which avoid overscaling problem. However, the [current behavior](https://github.com/ray-project/ray/blob/master/python/ray/autoscaler/_private/aws/node_provider.py#L403) also does not guarantee target capacity will reached. Autoscaling monitor loop will figure out any underscaling issue
```
ericl, line 52: > aggregated them based on the latest family or high spec instance

To clarify, does this mean we consider the resource shape of a particular "node config" to be that of the largest instance in the group?

reps/2023-01-09-aws-fleet-support.md
```diff
@@ -0,0 +1,138 @@
...
+
+Today, AWS autoscaler requires developers to choose their EC2 instance types upfront and codify it in their [autoscaler config](https://github.com/ray-project/ray/blob/master/python/ray/autoscaler/aws/example-full.yaml#L73-L80). EC2 offers a lot of instances types across different families with varied pricing. Hence, developers can realize cost-savings if their workload does not have a strict dependency on a specific family of instances. [EC2 Fleet](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-fleet.html) is an AWS offering that allows developers to launch a group of instances depending on various parameters like, maximum amount per hour they are willing to pay, supplement their primary on-demand capacity with spot capacity, specify maximum spot prices for each instance, choose from multiple allocation strategies for on-demand and spot capacity etc. This proposal outlines the work needed to support EC2 Fleet on ray autoscaler.
+
+#### Key requirements
+
+-   Autoscaler must offer existing functionalities and hence changes must be backward compatible.
+-   Developers must be able to make use of most of the critical features AWS EC2 Fleet offers.
```
wuisawesome, line 10: Can we be a little more specific about what "critical" means here? Is it fallback functionality, pricing strategies, node "maintenance", etc

reps/2023-01-09-aws-fleet-support.md
```diff
@@ -0,0 +1,138 @@
...
+              ImageId: ami-04af5926cc5ad5248
+      TargetCapacitySpecification:
+        DefaultTargetCapacityType: 'on-demand'
+....
+```
+
+### What are the known limitations?
+
+-   The fleet request types `maintain` and `request` are not supported. Also, a hard limit on number of fleets of type `maintain` or `request` may block us from supporting that in future as well.
+-   EC2 fleet cannot span different subnets from same availability zone. This is more of a limitation of EC2 fleets which could impact availability.
```
wuisawesome, line 97: Just to clarify, we can still schedule across different subnets in _different_ AZs right? (that's also the standard ray setup)

reps/2023-01-09-aws-fleet-support.md
wuisawesome: For safety, it should probably be the smallest? +1 to clarifying either way.

raghumdani: Yes, that's correct. @wuisawesome, any thoughts on why it should be smallest? My understanding is that when resource shape of largest instance is reported, it can take autoscaler multiple iterations to scale up appropriately if EC2 fleet ends up creating instances of smaller instance (in terms of cpu, memory or gpu) types. However other case will lead to overscaling, hence autoscaler has to terminate excess nodes.

reps/2023-01-09-aws-fleet-support.md
raghumdani: Yes, that's correct.

reps/2023-01-09-aws-fleet-support.md
wuisawesome: I'm wondering about the case where a fleet includes p3.2xlarge and p3.16xlarge, if someone has a `ray.remote(num_gpus=4)`, using the smallest would hang/inifinite loop on launching resources forever.

raghumdani: Ah got it. In that case, we would want to show this error, correct?

```
No available node types can fulfill resource request {}. Add suitable node types to this cluster to resolve this issue.
```

Then I agree we should consider lowest spec instance. In that case, we may end up with nodes having more resources than necessary if EC2 fleet always spins up highest spec instance.

wuisawesome: yeah in the initial version of this, i think we'd want to scope this to only recommend people use instances of the same "resource shape" in the same node type (e.g. the spot/on-demand version of a node, or m6.Nxlarge, m5.Nxlarge, m4.Nxlarge) to avoid this problem and have that logic for safety. I think we'd need something that looks broader than the node provider interface to support different resource shapes well.

raghumdani: agree, updated the rep. Also, I'd think part of the problem is bringing in predictability of what EC2 fleets is going to choose so that autoscaler can make accurate resource planning.

ericl: Yup, +1 to recommend using instances of the same shape. If the user wants to use two different instance sizes, this could be handled as two different node config entries. In this case, the autoscaler would generally prefer to launch the larger node config if there is sufficient pending demand, but would otherwise launch the smaller node config.

reps/2023-01-09-aws-fleet-support.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Should this change be within `ray` or outside?
+
+Yes. Specifically within AWS autoscaler.
+
+## Stewardship
+
+### Required Reviewers
+
+-   TBD
```
ericl, line 27: I think myself, @scv119 should be required at least.

reps/2023-01-09-aws-fleet-support.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Shepherd of the Proposal (should be a senior committer)
+
+-   TBD
```
ericl, line 31: @wuisawesome do you want to shepherd this and suggest additional reviewers?

reps/2023-01-09-aws-fleet-support.md
wuisawesome: Let's also add @architkulkarni 

@DmitriGekhtman would love to get your opinion too, but won't block on your review if you're busy

raghumdani: added suggested reviewers.

reps/2023-01-09-aws-fleet-support.md
```diff
@@ -0,0 +1,147 @@
...
+
+### How will the demand be converted into resources?
+
+Autoscaling monitor receives several demands from the metrics like vCPUs, memory, gpus etc. The [resource_demand_scheduler.py](https://github.com/ray-project/ray/blob/master/python/ray/autoscaler/_private/resource_demand_scheduler.py) is responsible for converting the demand into number of workers per node type specified in the autoscaler config. However, it relies on the **node provider** to statically determine the CPU, memory, GPU and any custom resources. Hence, **node_provider** will determine CPU, memory and GPU from the `InstanceRequirements` and specified `InstanceType` parameters and aggregate them based on the low spec instance. Hence, autoscaler may end up spinning up more nodes than necessary which leads to overscaling. Autoscaler will also identify the idle nodes and remove them if required in the subsequent iterations. However, the [current behavior](https://github.com/ray-project/ray/blob/master/python/ray/autoscaler/_private/aws/node_provider.py#L403) also does not guarantee target capacity will reached. Autoscaling monitor loop will figure out any underscaling issue
```
DmitriGekhtman, line 59: The Ray scheduler will also use the resource capacities that are inferred here. I am not sure if that's desirable.
It may be possible to introduce some kludging to make sure the real resource capacities are perceived by the Ray internals.

raghumdani: Are you suggesting resource demand scheduler to bin-pack based on real resource capacities instead of inferring it this way? If I am not wrong, the inherent assumption is that every node type will represent finite amount of resources (vCPU and memory) and hence it is being statically retrieved currently. We will only predict future capacity offered by fleets as there is no guarantee that EC2 will provision requested amount of units due to availability within instance pools.

wuisawesome: > The Ray scheduler will also use the resource capacities that are inferred here

I think that was only a legacy code path for handling the old `worker_nodes` field that was removed in 2.0. In any case, we probably shouldn't rely on that...

wuisawesome: here's the [code path which isn't active ](https://github.com/ray-project/ray/blob/master/python/ray/autoscaler/_private/resource_demand_scheduler.py#L363)

DmitriGekhtman: > I think that was only a legacy code path.

You might be thinking of something else. I'm referring to the `RAY_OVERRIDE_RESOURCES` env, which is used to advertise the (auto-filled) resources from the autoscaling config to the Ray scheduler.

If the proposal is implemented as is, then both the autoscaler and the Ray scheduler will see the reduced resource capacities. Come to think of it, I don't see a straightforward way of avoiding that here.

Side: There's a related contortion [here](https://github.com/ray-project/ray/blob/1a8f330ae9502b5911bc34cd34a42929a3d789d9/python/ray/autoscaler/_private/resource_demand_scheduler.py#L215) that will update _absent_ resources from runtime.

wuisawesome: Oh yeah I think you're right here. I think we should double down on only supporting node types with the same resource shape for now to avoid these problems for the most immediate proposal, and generally make sure the autoscaler is more well informed about node resources (rather than trying to guess what a crazy ray core heuristic will do) to avoid these issues if we need to broaden the class of node types supported by this.

DmitriGekhtman: that sgtm

raghumdani: Makes sense, I misunderstood the problem earlier. Resolving this thread.

DmitriGekhtman: Probably the REP should updated to indicated that we will not initially recommend using instances with different resource capacities in a single fleet.

raghumdani: Makes sense. Below is one of the item `Proposed code changes`:

> Update documentation to recommend developers to use instances with similar resource shape to avoid overscaling though EC2 fleet allows them to acquire instances at lower costs.

Do you want to callout explicitly in any other section?

DmitriGekhtman: looks good

glindstr:
Hey I'm very interested in this REP. Could someone update me on the current status?

edoakes:

PR #22, opened 2023-02-06
 Add node affinity feature enhancements  REP

larrylian:
The value of introducing the Labels mechanism for scheduling was discussed in this REP（[Add labels mechanism] #13 ） before. NodeAffinity is now discussed separately from this REP.

In fact, everyone should know about NodeAffinity before. I think we can reach an agreement on the following points in this REP.

- [ ] Agreed on API format
- [ ] The main implementation plan is agreed
- [ ] Ask specific single-point questions, and then discuss and reach a consensus
- [ ] Some specific details can be discussed during the coding process.

### DIFF
--- a/reps/2023-02-03-node-affinity-feature-enhancements.md
+++ b/reps/2023-02-03-node-affinity-feature-enhancements.md
@@ -1,0 +1,437 @@
+## Summary
+### General Motivation
+
+The current node affinity feature is relatively simple, it can only support affinity with one node. 
+Many general functions cannot be supported.
+
+Current node affinity feature don't support functions:
+1. Don't support affinity to a batch of nodes. Current only support affinity to one node.
+2. Don't support anti-affinity to a node of a batch of nodes. 
+3. Don't support to classify nodes by labels, and then affinity or anti-affinity to a certain type of nodes.
+4. Don't support affinity/anti-affinity to nodes by node ip or node name. Current only support node id identification.
+
+This REP is to solve the above problems.
+
+
+### Should this change be within `ray` or outside?
+
+Yes, this will be a complement to ray core's ability to flexibly schedule actors/tasks/node.
+
+## Stewardship
+### Required Reviewers
+ 
+@ericl @stephanie-wang @wumuzi520 SenlinZhu @Chong Li  @scv119 (Chen Shen) @Sang Cho @jjyao (Jiajun Yao) @Yi Cheng
+### Shepherd of the Proposal (should be a senior committer)
+@scv119 (Chen Shen)
+
+## Design and Architecture
+
+### Brief idea
+1. Introduct the concept of the static labels of nodes. Add static labels to nodes(only be set on node start), used to classify nodes. Labels(key, value) = Map<String, String>
+2. Improve the expression of NodeAffinity to support affinity\anti-affinity and labels expressions.
+3. The actual principle of NodeAffinitySchedulingStratgy is. According to the label matching expression in the strategy, traverse and search for LabelsResource of all nodes. Pick out nodes that meet the expression requirements.
+
+
+![NodeAffinity Concept](https://user-images.githubusercontent.com/11072802/217133358-19ddc916-bf15-4f69-96e1-9cbd157b0e1d.png)
+
+Scheduling Policy | Label Owner | Label select operator
+-- | -- | --
+NodeAffinity | Node | in, not_in, exists, does_not_exist
+
+### API Design
+
+**The apis of actors/nodes add labels**  
+This interface is already very simple, so we will not set up multiple solutions for everyone to discuss.
+```python
+#  Node add static labels. 
+ray start ... --labels={"location": "dc_1"}
+
+# Dynamically updating node labels is not implemented now, and will be considered in the second stage.
+```
+
+**The apis of the actor-affinity/node-affinity scheduling.**  
+
+**Option 1: Simplify through syntactic sugar**
+
+
+```python
+# Scheduled to a node with a specific IP.
+actor_1 = Actor.options(
+        scheduling_strategy=node_affinity(label_in(key="node_ip", values=["xxx.xxx.xx.xx"], is_soft=false))
+    ).remote()
+
+# Try to schedule to the node with A100/P100 graphics card. If not, schedule to other nodes.
+actor_1 = Actor.options(
+        scheduling_strategy=node_affinity(label_in("gpu_type", ["A100", "P100"], is_soft=true))
+    ).remote()
+
+# Do not schedule to the two nodes whose node id is "xxxxxxx"\"aaaaaaaa".
+actor_1 = Actor.options(
+        scheduling_strategy=node_affinity(label_not_in("node_id", ["xxxxxxx", "aaaaaaaa"], is_soft=false))
+    ).remote()
+
+# Schedule to the node with the key label exist "gpu_type".
+actor_1 = Actor.options(
+        scheduling_strategy=node_affinity(label_exist("gpu_type"))
+    ).remote()
+
+# Don't schedule to the node with the key label exist "gpu_type".
+object_ref = Task.options(
+        scheduling_strategy=node_affinity(label_does_not_exist("gpu_type", is_soft=false))
+    ).remote()
+
+# Multiple label expressions can be filled in at the same time, and the relationship between the expressions is "and". The dispatch must satisfy each expression.
+# The actual meaning of this expression is that it must be scheduled to a node with a GPU, and as much as possible to a node with a GPU of the A100 type.
+actor_1 = Actor.options(
+        scheduling_strategy=node_affinity([
+            label_in("gpu_type", ["A100"], true),
+            label_exists("gpu_type", false)
+        ])
+    ).remote()
+
+# Note: This api is old. The old api will still be kept for compatibility with old users.
+actor_1 = Actor.options(
+    scheduling_strategy=NodeAffinitySchedulingStrategy(
+        node_id=ray.get_runtime_context().node_id,
+        soft=False,
+    )
+).remote()
+
+def node_affinity(...):
+    ...
+    return NodeAffinitySchedulingStrategy(...)
+
+def label_in(key, values, is_soft=false):
+    ...
+    return LabelMatchExpression(...)
+
+def label_not_in(key, values, is_soft=false):
+    ...
+    return LabelMatchExpression(...)
+
+def label_exists(key, is_soft=false):
+    ...
+    return LabelMatchExpression(...)
+
+def label_does_not_exist(key, is_soft=false):
+    ...
+    return LabelMatchExpression(...)
+
+```
+
+**Option 2: another syntactic sugar** 
+
+Personally, I think this Option is not as good as the above Option 1.  
+The label_in(key, values, is_soft) form of option 1 is more understandable and better than the form of ("location", LabelMatchOperator.IN, ["dc_1"], false).
+```python
+actor_1 = Actor.options(
+        scheduling_strategy=NodeAffinity([
+            ("node_ip", LabelMatchOperator.IN, ["xxx.xxx.xx.xx"], is_soft=false),
+    ).remote()
+
+actor_1 = Actor.options(
+        scheduling_strategy=NodeAffinity([
+            ("node_id", LabelMatchOperator.NOT_IN, ["xxxxxxx", "aaaaaaaa"], is_soft=true),
+    ).remote()
+
+object_ref = Task.options(
+        scheduling_strategy=NodeAffinity([
+            ("gpu_type", LabelMatchOperator.IN, ["A100"], is_soft=true),
+            ("gpu_type", LabelMatchOperator.Exist)).
+    ).remote()
+```
+
+**Option 3: Java-like form**  
+
+This form is similar to Java's syntax. The downside is that it's a bit complicated.
+```python
+SchedulingStrategyT = Union[None, str,
+                            PlacementGroupSchedulingStrategy,
+                            NodeAffinitySchedulingStrategy]
+
+class NodeAffinitySchedulingStrategy:
+    def __init__(self, match_expressions: List[LabelMatchExpression]):
+        self.match_expressions = match_expressions
+
+class LabelMatchExpression:
+    """An expression used to select instance by instance's labels
+    Attributes:
+        key: the key of label
+        operator: IN、NOT_IN、EXISTS、DOES_NOT_EXIST,
+                  if EXISTS、DOES_NOT_EXIST, values set []
+        values: a list of label value
+        soft: ...
+    """
+    def __init__(self, key: str, operator: LabelMatchOperator,
+                 values: List[str], soft: bool):
+        self.key = key
+        self.operator = operator
+        self.values = values
+        self.soft = soft
+
+actor_1 = Actor.options(scheduling_strategy=NodeAffinitySchedulingStrategy([
+                LabelMatchExpression(
+                    "node_ip", LabelMatchOperator.IN, ["xxx.xxx.xx.xx"], False)
+            ])).remote()
+
+actor_1 = Actor.options(scheduling_strategy=NodeAffinitySchedulingStrategy([
+                LabelMatchExpression(
+                    "gpu_type", LabelMatchOperator.IN, ["A100"], True),
+                LabelMatchExpression(
+                    "gpu_type", LabelMatchOperator.EXISTS, None, False)
+            ])).remote()
+```
+
+**Option 4: Like sql**  
+
+This solution is not recommended.  
+This method needs to parse SQL, and the workload will be much larger.  
+And users often write wrong sql when using it.
+```python
+# NodeAffinity use case
+actor_1 = Actor.options(
+        scheduling_strategy=NodeAffinity("gpu_type in [A100, T100]")
+    ).remote()
+```
+
+### Example
+
+**1. Affinity to the node of the specified Instance.**  
+```
+# m6in.2xlarge 8C32G instance node
+ray start ... --labels={"instance": "m6in.2xlarge"}
+
+# m6in.large 2C8G instance node
+ray start ... --labels={"instance": "m6in.large"}
+
+actor_1 = Actor.options(
+        scheduling_strategy=node_affinity(label_in(key="instance", values=["m6in.large"], is_soft=false))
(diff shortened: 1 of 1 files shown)


ericl:
This REP seems like it would be blocked on the autoscaler / GCS refactoring proposal, right @scv119 ? Do we need to have that out first before reviewing this one?

reps/2023-02-03-node-affinity-feature-enhancements.md
```diff
@@ -0,0 +1,367 @@
...
+  absl::flat_hash_map<string, absl::flat_hash_map<string, absl::flat_hash_map<string, string>>> labels;
+}
+```
+
+2. When a node registers with GCS and broadcasts to all nodes, add the static labels information to NodeResources.
+
+3. Scheduling optimization through Labels  
+Now any node raylet has node static labels information for all nodes.  
+when NodeAffinity schedules, if it traverses the Labels of each node, the algorithm complexity is very large, and the performance will be poor.   
+** Therefore, it is necessary to generate a full-cluster node labels index table to improve scheduling performance. </b>
```
ericl, line 280: What's the index strategy for the more complex expressions such as negation and containment? I wonder if we need to cache the list of valid nodes per task/expression to avoid O(n^2) scheduling slowdowns with many nodes and tasks.

larrylian:
> This REP seems like it would be blocked on the autoscaler / GCS refactoring proposal, right @scv119 ? Do we need to have that out first before reviewing this one?

Thanks for checking it so quickly, Now i'm still just a draft, i finished it today, and i'll @you to help review it.

reps/2023-02-03-node-affinity-feature-enhancements.md
larrylian: This tag indexing strategy is actually very simple, I added it in the REP. The tag index only needs to get the nodes with these key&values.
image

reps/2023-02-03-node-affinity-feature-enhancements.md
```diff
@@ -0,0 +1,399 @@
...
+
+**The apis of the actor-affinity/node-affinity scheduling.**  
+
+**Option 1: Simplify through syntactic sugar**
+
+
+```python
+# Scheduled to a node with a specific IP.
+actor_1 = Actor.options(
+        scheduling_strategy=node_affinity(label_in(key="node_ip", values=["xxx.xxx.xx.xx"], is_soft=false))
```
stephanie-wang, line 60: Minor, but seems like we don't really need the `node_affinity` wrapper here?

reps/2023-02-03-node-affinity-feature-enhancements.md
```diff
@@ -0,0 +1,399 @@
...
+
+# Note: This api is old. The old api will still be kept for compatibility with old users.
+actor_1 = Actor.options(
+    scheduling_strategy=NodeAffinitySchedulingStrategy(
+        node_id=ray.get_runtime_context().node_id,
+        soft=False,
+    )
+).remote()
+
+def node_affinity(...):
```
stephanie-wang, line 100: Instead of introducing a new `node_affinity` wrapper, we can just swap out the `NodeAffinitySchedulingStrategy` implementation to use labels instead.

reps/2023-02-03-node-affinity-feature-enhancements.md
```diff
@@ -0,0 +1,399 @@
...
+
+Because NodeInfo originally had a mechanism to synchronize to the entire cluster. So there is no need to modify too much code.
+
+4. Scheduling optimization through Labels 
+
+Now any node raylet has node static labels information for all nodes.  
+when NodeAffinity schedules, if it traverses the Labels of each node, the algorithm complexity is very large, and the performance will be poor.   
+<b> Therefore, it is necessary to generate a full-cluster node labels index table to improve scheduling performance. </b>
+
+The label index table is only used to obtain which nodes have this key and value. The complexity of this algorithm is O(1).
```
stephanie-wang, line 296: If there are multiple expressions to match, I think it will be O(kN) (k = num expressions, N = nodes), though, right?

reps/2023-02-03-node-affinity-feature-enhancements.md
```diff
@@ -0,0 +1,399 @@
...
+3. When using Labels to build a Labels index table, you need to parse the resources key.
+
+**DisAdvantages:**
+1. Labels unlike cpu resource these are numeric types. Compared with the above scheme. This will destroy the concept of coustrom resouce.
+2. Actor and Task are isolated by namespace. It is difficult to isolate through namespace if adding custom_resource.
+2. The Label index table of all nodes can be constructed from the ActorLabels information of each node. If you use Custom Resource, this requires parsing the resource_key and doing a lot of string splitting which will cost performance.
+3. If custom_resource happens to be the same as the spliced string of labels. Then it will affect the correctness of scheduling.
+
+### AutoScaler adaptation
+I understand that the adaptation of this part of AutoScaler should not be difficult. Now NodeAffinity is just one more scheduling strategy. Just follow the existing implementation of AutoScaler and adapt to the current strategy.
```
stephanie-wang, line 372: This proposal is more advanced than the autoscaling for NodeAffinity.

NodeAffinity is only used for nodes that are already in the cluster, so I believe the only change there was to make sure that the autoscaler didn't mistakenly start additional nodes for a queued task that wouldn't actually be able to be scheduled there.

Since this proposal is for more flexible labels, we will need the autoscaler to be aware of which nodes it should add or remove from the cluster based on the label constraints of queued tasks. So this actually is something the proposal should cover. Specifically:
- what API changes do we need at the cluster config level and for the interface between the autoscaler and the GCS?
- what will the autoscaler policy look like?
- Handling scalability issues? (If an application uses too many different label policies, we'll have to report these in the task load)

reps/2023-02-03-node-affinity-feature-enhancements.md
```diff
@@ -0,0 +1,399 @@
...
+  ...
+  // The user-provided identifier or name for this node.
+  string node_name = 12;
+  // The static labels of nodes.
+  map<String, String> labels = 13;
+}
+```
+2. Add the labels data structure to the resource synchronization data structure(NodeResources). 
+```
+NodeResources {
```
stephanie-wang, line 274: Wouldn't we need to add the label shapes requested by tasks to the ResourceRequests?

stephanie-wang (commented):
This looks great, thanks for putting it together!

The main comment we should resolve before approval is about the autoscaling implementation. Ideally we would not block this REP on the autoscaler refactor, but it depends on the complexity of this design. Could you address the comment I left about this, and then we can better determine order of operations?

The other question I had is about semantics for "soft". Can you confirm that soft only considers cluster feasibility of a request's label constraints? I.e., it would still schedule a task to a feasible node even if the node has high load?

reps/2023-02-03-node-affinity-feature-enhancements.md
ericl: Yeah I don't think this is quite so simple. The autoscaler runs emulated bin-packing to determine what types of nodes it needs to add to the cluster. For example, suppose a user requests {"dc": "zone-1"}. Then, the autoscaler will need to run its bin packing routine with this request and node labels in mind, in order to determine the nodes to launch: https://github.com/ray-project/ray/blob/master/python/ray/autoscaler/_private/resource_demand_scheduler.py

As you can see, the logic is algorithmically complex.

My assessment is that we probably need to refactor the autoscaler to unify this bin packing routine with the C++ implementation, to avoid having to write label affinity evaluation in both Python and C++.

reps/2023-02-03-node-affinity-feature-enhancements.md
larrylian: The benefits of wrapping with node_affinity here are as follows:
1. The label representation in node_affinity can be multiple. So it needs to be wrapped with node_affinity
```
# Multiple label expressions can be filled in at the same time, and the relationship between the expressions is "and". The dispatch must satisfy each expression.
# The actual meaning of this expression is that it must be scheduled to a node with a GPU, and as much as possible to a node with a GPU of the A100 type.
actor_1 = Actor.options(
        scheduling_strategy=node_affinity([
            label_in("gpu_type", ["A100"], true),
            label_exists("gpu_type", false)
        ])
    ).remote()
```
2. After actor_affinity also intends to introduce label expressions. At that time, actor_affinity & node_affinity can reuse the four expressions label_in\label_exist at the same time. The code implementation can be more concise and easier for users to understand.
```
Actor.options(
        scheduling_strategy=node_affinity(label_in(key="node_ip", values=["xxx.xxx.xx.xx"], is_soft=false)).remote()

Actor.options(
        scheduling_strategy=actor_affinity(label_in(key="actor_type", values=["dog"], is_soft=false)).remote()

```

reps/2023-02-03-node-affinity-feature-enhancements.md
larrylian: Same as above explanation. Label expressions are planned to be used in more places in the future. For example, ActorAffinity or get a batch of nodes through label. So it's better to wrap it with node_affinity/NodeAffinitySchedulingStrategy here. Then the label indicates that it can be reused in multiple places.
```
# get a batch of nodes through label.
Nodes = GetNodeInfoByLabel([label_in("zone", ["usa", "china"])
```

reps/2023-02-03-node-affinity-feature-enhancements.md
larrylian: Yes, In a scene with many expressions, each time an expression is processed, fewer and fewer nodes meet the requirements. Performance is good.

reps/2023-02-03-node-affinity-feature-enhancements.md
larrylian: This is for synchronizing the Node Labels information to each raylet node in the Raylet scheduling mode.
1. Putting it in NodeResources can reuse the existing resource synchronization mechanism without creating a new label synchronization mechanism
2. Adding labels here can also allow ActorAffinity to reuse this mechanism.
3. Under the Raylet scheduling mechanism, after the scheduling is completed, the local node needs to confirm whether the final resources meet the requirements. So it is convenient to put it here for final confirmation.

larrylian:
@scv119  @ericl @stephanie-wang 
1. We understand that you also have demands for Runtime Env acceleration, and I have also added a proposal to use Node Labels to achieve Runtime Env acceleration, you can take a look.

2. Do you have any comments on the two REPs Node affinity and Actor Affinity? Can you help advance the progress of these two REPs? This way I can start developing as soon as possible.

 Use Node Labels to achieve Runtime Env acceleration:  
image

reps/2023-02-03-node-affinity-feature-enhancements.md
```diff
@@ -0,0 +1,437 @@
...
+
+Scheduling Policy | Label Owner | Operator
+-- | -- | --
+nodeAffinity | NODE | In, NotIn, Exists, DoesNotExist, Gt, Lt
+podAffinity | POD | In, NotIn, Exists, DoesNotExist
+podAntiAffinity | POD | In, NotIn, Exists, DoesNotExist
+
+
+### what's the alternative to achieve the same goal?
+** Option 3: putting Labels into the custom resource solution  **
```
jjyao, line 343: Not necessarily saying that I'm proposing this but have we considered label being a k-v pair vs a string. Does a string label have the same expressiveness as a k-v label? A good thing of string label is that it can be easily implemented as custom resources internally.

larrylian: @jjyao 

>have we considered label being a k-v pair vs a string?
We should have discussed this before, and now the form of key-value should refer to the best form of other excellent experience.

1. The (key-value) labels mechanism of K8s is one of the highlights of K8s and has been recognized by countless people.
2. The key-value is simple and can be accepted by users naturally. But at the same time a certain flexibility can be submitted. For example: {gpu_type: "A100"} , {gpu_type: "T100"} . This setting is very natural and intuitive. But if it is replaced by a string: {"gpu_type@A100"},{"gpu_type@T100"}. It will be a little troublesome for users.

> implemented as custom resources internally

This issue has been discussed before. NodeAffinity uses custom_resource internally. Wouldn't be a good solution. It not only destroys the semantics of the original custom_resouce, but also the code implementation will be more complicated (1+1)> 2.

https://github.com/ray-project/enhancements/pull/13/files?short_path=5ae2129#diff-5ae21290dd6f2483719e2557543e9335c1867bea3d1f1cfe516e269d0bd9bc12

image

reps/2023-02-03-node-affinity-feature-enhancements.md
```diff
@@ -0,0 +1,437 @@
...
+### Should this change be within `ray` or outside?
+
+Yes, this will be a complement to ray core's ability to flexibly schedule actors/tasks/node.
+
+## Stewardship
+### Required Reviewers
+ 
+@ericl @stephanie-wang @wumuzi520 SenlinZhu @Chong Li  @scv119 (Chen Shen) @Sang Cho @jjyao (Jiajun Yao) @Yi Cheng
+### Shepherd of the Proposal (should be a senior committer)
+
```
zhe-thoughts, line 25: This is a required piece of information. I suggest @scv119

reps/2023-02-03-node-affinity-feature-enhancements.md
```diff
@@ -0,0 +1,437 @@
+## Summary
+### General Motivation
+
+The current node affinity feature is relatively simple, it can only affinity with one node.
```
zhe-thoughts, line 4: "it can only affinity" -> "it can only support affinity"

reps/2023-02-03-node-affinity-feature-enhancements.md
scv119: i think that's option2 right?

jjyao: Not exactly. Option 2 doesn't have `label_in` etc?

reps/2023-02-03-node-affinity-feature-enhancements.md
```diff
@@ -0,0 +1,437 @@
...
+```
+NodeResources {
+  ResourceRequest total;
+  ResourceRequest available;
+  /// Only used by light resource report.
+  ResourceRequest load;
+  /// Resources owned by normal tasks.
+  ResourceRequest normal_task_resources
+  /// Map<label_type, Map<namespace, Map<label_key, Map<label_value, ref_count>>>> Actors/Tasks/Nodes labels information
+  absl::flat_hash_map<string, absl::flat_hash_map<string, absl::flat_hash_map<string, absl::flat_hash_map<string, int>>>> labels;
```
scv119, line 282: For node affinity where the node labels are static, we don't need this complicated labels data structure, but only `map<String, String> labels` is sufficient?

reps/2023-02-03-node-affinity-feature-enhancements.md
scv119: Since ResourceRequest is also used for autoscaling, seemingly we need to add label requests as well.

reps/2023-02-03-node-affinity-feature-enhancements.md
larrylian: @jjyao  Do you mean "scheduling_strategy = NodeAffinitySchedulingStrategy([label_in("zone", ["usa", "china"]])?  

NodeAffinitySchedulingStrategy is replaced by node_affinity mainly for more concise and language style more in line with python.

reps/2023-02-03-node-affinity-feature-enhancements.md
larrylian: Yes. map<String, String> labels is sufficient.  Here we can use a concise data structure first. Change it later when we need to expand.

reps/2023-02-03-node-affinity-feature-enhancements.md
larrylian: Agree . I will add it.

reps/2023-02-03-node-affinity-feature-enhancements.md
larrylian: done

reps/2023-02-03-node-affinity-feature-enhancements.md
scv119: @larrylian we can stick with NodeAffinitySchedulingStrategy for the first version, and improve that in a separate PR.

larrylian: @scv119  Do you mean "scheduling_strategy = NodeAffinitySchedulingStrategy([label_in("zone", ["usa", "china"]])" ?    
It's ok to do that。

reps/2023-02-03-node-affinity-feature-enhancements.md
larrylian: Ok I'll try to update a rough design by next week

rkooo567:
oh I think it is the same comment as https://github.com/ray-project/enhancements/pull/22#discussion_r1101633839 actually. So I believe it is addressed

reps/2023-02-03-node-affinity-feature-enhancements.md
```diff
@@ -0,0 +1,564 @@
...
+### If the Match Expression Cannot be satisfied
+If the matching expression cannot be satisfied, The actor/task will be add to the pending queue. Util the matching expression all be statisfied.
+
+1. resources are enough, but node_id of affinity don't exist. -> throw exception, compatible with current behavior 
+
+2. resources are enough, but labels affinity  cannot be satisfied? -> Hang & report schedule failed event and detail unstaisfed reason to exposed to users
+
+3. labels affinity can be satisfied, but resources are not enough? -> Hang & report schedule failed event and detail unstaisfed reason to exposed to users
+
+4. both labelsaffinity and resources are not satisfied? -> Hang & report schedule failed event and detail unstaisfed reason to exposed to users
```
jjyao, line 329:
```suggestion
4. both labels affinity and resources are not satisfied? -> Hang & report schedule failed event and detail unstaisfed reason to exposed to users
```

jjyao (approved):

reps/2023-02-03-node-affinity-feature-enhancements.md
larrylian: done

rkooo567 (approved):

zhe-thoughts (approved):

PR #23, opened 2023-02-07
WASM On Ray Support REP

wilsonwang371:
This is a pull request for adding wasm support to Ray.

### DIFF
--- a/reps/2023-02-01-wasm-on-ray.md
+++ b/reps/2023-02-01-wasm-on-ray.md
@@ -8,6 +8,14 @@ To make it easier to support new languages, we can use WebAssembly (WASM) to sup
 
 Furthermore, it is also possible that we enable transparent Ray support in WASM. As a result, many legacy applications can be easily migrated to Ray and run in a distributed manner and Ray layer can be transparent to the application. This will be a huge benefit for Ray adoption in the generic serverless computing environment.
 
+In our internal FaaS environment, we have WASM function support. These functions are compiled from different programming languages such as Javascript, Rust, Go, C/C++, etc. Compared with normal container based FaaS solutions such as Knative, WASM based functions can be loaded with a shorter period of time while still providing the same or even better secure running environments. So far, WASM functions are still far from perfect. There are few pain points for WASM functions.
+
+1. If different functions want to communicate with each other, they need use either RPC, message queue or KV store. However, these solutions are not convenient for the developers. The FaaS function developers need a simpler communication mechanism between different functions and run tasks in parallel.
+2. WASM function is executed in a single thread. Although multi-threading support is on the way, it is still not mature enough. To run tasks in a single FaaS function in parallel, running WASM with Ray API support is a good solution. Besides, running a WASM function distributedly in Ray has the FT support out of the box.
+3. Our FaaS environment has a large number of edge servers. The resources on each of the servers are limited. It is not possible to provide more resources for a single function than the resources on the server. In Ray, all the resources in the Ray cluster is treated as a single resource pool. If we can support Ray in our FaaS environment, we can run WASM functions with larger resource requirements.
+
+As a result, Having Ray support in our FaaS WASM environment will be a huge benefit for us and it is a very important feature for us.
+
 One use case of WASM on Ray is the ML inference tasks. In our enterprise environment, we need to support running ML inference tasks in Ray and make these ML inference tasks share the Ray cluster resources. Compared to the current Ray Serve ML inference solution, which needs to keep the ML inference tasks running in the Ray cluster memory, the WASM ML inference solution can reduce the resource consumption of the Ray cluster. Currently, WASM runtime such as WasmEdge already support running ML inference tasks through WASI-NN support.
 
 In this REP, we will discuss the design of WASM support in Ray. We will first discuss the general architecture of WASM support in Ray. Then we will discuss how to support Ray APIs in WASM. Finally, we will discuss an example of how to use Ray APIs in WASM.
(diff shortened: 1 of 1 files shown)


wilsonwang371:
@iycheng @scv119

juntao:
This is great! I would like to add that, while there are challenges in supporting native AI/ML frameworks in WASM, there are successful use cases and even standards to accomplish this. The [WASI-NN](https://github.com/WebAssembly/wasi-nn) standard allows WASM bytecode programs to access ML frameworks. Tensorflow, Tensorflow Lite, PyTorch, and OpenVINO are all supported. There are Rust APIs for this:

https://wasmedge.org/book/en/write_wasm/rust/wasinn.html

and JavaScript APIs:

https://wasmedge.org/book/en/write_wasm/js/tensorflow.html

Would like to collaborate and see Ray libraries supported in WASM!

reps/2023-02-01-wasm-on-ray.md
```diff
@@ -0,0 +1,253 @@
...
+
+Furthermore, it is also possible that we enable transparent Ray support in WASM. As a result, many legacy applications can be easily migrated to Ray and run in a distributed manner and Ray layer can be transparent to the application. This will be a huge benefit for Ray adoption in the generic serverless computing environment.
+
+In our internal FaaS environment, we have WASM function support. These functions are compiled from different programming languages such as Javascript, Rust, Go, C/C++, etc. Compared with normal container based FaaS solutions such as Knative, WASM based functions can be loaded with a shorter period of time while still providing the same or even better secure running environments. So far, WASM functions are still far from perfect. There are few pain points for WASM functions.
```
iycheng, line 11: This seems like a big adds to WASM if Ray can have WASM workers, in summary, seems like:

Ray adds on WASM:
- ObjectRef for communication
- Ray worker for parallelism and scale up

WASM adds on Ray:
- Lightweight x-lang support
- Lightweight isolation

iycheng: cc @scv119

iycheng:
Synced it offline and I think the motivation is much clearer. I think the MVP features needed for this use case should be briefly discussed so for the people who don't understand WASM (or Ray), they know the work in high-level.

reps/2023-02-01-wasm-on-ray.md
```diff
@@ -0,0 +1,253 @@
...
+3. Our FaaS environment has a large number of edge servers. The resources on each of the servers are limited. It is not possible to provide more resources for a single function than the resources on the server. In Ray, all the resources in the Ray cluster is treated as a single resource pool. If we can support Ray in our FaaS environment, we can run WASM functions with larger resource requirements.
+
+As a result, Having Ray support in our FaaS WASM environment will be a huge benefit for us and it is a very important feature for us.
+
+One use case of WASM on Ray is the ML inference tasks. In our enterprise environment, we need to support running ML inference tasks in Ray and make these ML inference tasks share the Ray cluster resources. Compared to the current Ray Serve ML inference solution, which needs to keep the ML inference tasks running in the Ray cluster memory, the WASM ML inference solution can reduce the resource consumption of the Ray cluster. Currently, WASM runtime such as WasmEdge already support running ML inference tasks through WASI-NN support.
```
iycheng, line 19: I think for Ray Serve ML inference, the ML also runs in the memory. Could you give more detail about the difference between going with Ray Serve and WASI-NN ?

reps/2023-02-01-wasm-on-ray.md
```diff
@@ -0,0 +1,253 @@
...
+3. Lightweight Workers. WASM is only 10% the size of Python for comparable tasks. This means that the WASM worker is much smaller than the Python worker. WASM is lightweight and fast to load. It is possible to load the WASM binary in a few milliseconds. Compared to Python, it is much faster to load the WASM binary.
+4. Architecture Agnostic. WASM provides an abstract layer on top of the hardware. It is possible to support different hardware architectures in the same WASM binary.
+5. Easy to support different versions of the same language. For example, if we want to support different versions of JavaScript, we only need to compile different versions of JavaScript to WASM and then use the WASM support in Ray to run the WASM binary.
+
+### Cons:
+1. It is not easy to support all Ray APIs in WASM. For example, there is a concept of `Actor` in Ray. `Actor` is a special type of task that can maintain state. However, WASM is a collection of functions, a stack, and a heap. It is not a simple task to map the concept of `Actor` to WASM. Therefore, we will not support `Actor` in WASM in the first version of WASM support in Ray.
```
jjyao, line 158: It's a significant con.

wilsonwang371: I think this part can help us review some of the Ray core APIs so that we can improve the Ray core interface. For the MVP, I think we cannot yet support Actor. With more brainstorming and design sessions, I think there will be a solution. I didn't see any technical challenges so far.

reps/2023-02-01-wasm-on-ray.md
wilsonwang371: wasi-nn is a low-level ML computing API. https://github.com/WebAssembly/wasi-nn/blob/main/wasi-nn.wit.md such as loading a model, setting input, computing with the model, and getting the output. In contrast, Ray Serve is a high-level Python API for the model serving. In order to do something similar in WASM, we need to port some of the Ray Serve from the Python worker to the WASM worker.

reps/2023-02-01-wasm-on-ray.md
iycheng: @jjyao there are several things here:

- Build workers
    - support functions (first step)
    - support actors (third step)
- Define middle layer apis (second step)

The difficulty comes from ray side. To be more specific, ray doesn't have good developer api to add new languages. And this is one of the goal of this project, to figure out what might be good api we should define.

jjyao: Figuring out a good middle layer for multi-language support is indeed very important (IMO, more important than supporting WASM itself), are we going to have a design for that?

iycheng: @jjyao I know what you mean. The fact here is that WASM worker support is critical to @wilsonwang371 and middle layer support is important for Ray. Given that we don't know what's needed, it's a good chance we'll learn it from this work by doing first step (support functions).

TBH, depend on the resources we have it might or might not happen. But at lease we'll understand the requirement better after finishing step 1.

wilsonwang371: totally agreed. Actually, we are also willing to contribute to the middle layer API.

reps/2023-02-01-wasm-on-ray.md
```diff
@@ -0,0 +1,270 @@
...
```
jovany-wang, line 158: @iycheng @wilsonwang371 
From your design, if we don't re-design a better lower level API for developers, it's not easy to implement WASM actor, right?
I'm concerning that it's verbose to do all these stuffs in one.
We should separate this REP to 2 proposals: a better lower level API for multi-languages workers proposal + WASM worker proposal.

The benefits are significant:
1. For the lower level API proposal, I guess there're many committers and reviewers want to review, while that the WASM worker proposal is focusing on how we port WASM on Ray, which I believe less of reviewers is available to review due to it requires more knowledge about WASM(For example, I'm lack of that).
2. Different goals should use different REPs for better track in the future.

jovany-wang (changes requested):

reps/2023-02-01-wasm-on-ray.md
wilsonwang371: I think WASM workers will not require a better lower-level API to be there in the first place. But our Bytedance team is definitely willing to work together on the lower-level API design & implementation.

wilsonwang371: Besides, the lower-level API is not a part of this REP.

jovany-wang: I see your point that this REP doesn't include the lower level API here.

I'm worrying about whether there might be a risk, like once this REP getting merged, and then it's impossible to support WASM actor.

If you guys are confident for this, let us make it clearer.

wilsonwang371: there will be ways to support actor in wasm. But we need to choose a solution that is acceptable to most people. That's why this is not a part of the MVP version

jovany-wang: OK. Please let me know once you have design or some stuffs. Looking forward the contributions.

iycheng: I think the issues here are that no one knows how to do it :(

Before starting actor support, we probably should get to the api layer first and we'll feel comfortable about this at that stage.

iycheng (approved):

zhe-thoughts:
Given the vote on ray-committers list, merging this PR. Thanks for the contribution @wilsonwang371 and @iycheng for shepherding!

PR #24, opened 2023-02-08
Add actor affinity scheduling strategy by actor handle

larrylian:
The value of introducing the Labels mechanism for scheduling was discussed in this REP（[Add labels mechanism] https://github.com/ray-project/enhancements/pull/13 ) before.
The ActorAffinity API discussed before has changed too much, so now a separate REP is separated to discuss the first version of ActorAffinity. Now this version of the ActorAffinity API only uses actor_handle.

In fact, everyone should know about ActorAffinity before. I think we can reach an agreement on the following points in this REP.

- [ ]  Agreed on API format
- [ ]  The main implementation plan is agreed
- [ ]  Ask specific single-point questions, and then discuss and reach a consensus
- [ ]  Some specific details can be discussed during the coding process.

### DIFF
--- a/reps/2023-02-07-actor-affinity-schedule-by-actor-handle.md
+++ b/reps/2023-02-07-actor-affinity-schedule-by-actor-handle.md
@@ -1,0 +1,358 @@
+
+## Summary
+### General Motivation
+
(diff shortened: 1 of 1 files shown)


edoakes:

PR #25, opened 2023-03-11
[AIR/Data] Revamping APIs for Offline Batch Inference

amogkam:
Reduce the friction for batch inference of pre-trained by updating the APIs to do the following:
1. Remove AIR Checkpoints from the critical path for batch prediction.
2. Simpler interface for chaining multiple preprocessing and prediction stages.

### DIFF
--- a/reps/2023-03-10-batch-inference.md
+++ b/reps/2023-03-10-batch-inference.md
@@ -1,0 +1,149 @@
+# Revamping APIs for Offline Batch Inference
+
+## Summary
+Reduce the friction for batch inference of pre-trained by updating the APIs to do the following:
+1. Remove AIR Checkpoints from the critical path for batch prediction.
+2. Simpler interface for chaining multiple preprocessing and prediction stages.
+
+### General Motivation
+There are changes to Ray Datasets in 2.3/2.4:
+1. Ray Datasets became lazy by default in Ray 2.3. Execution is no longer triggered until consumption is explicitly called.
+2. In Ray 2.4, Ray Datasets will move to streaming execution by default, resulting in there no longer being a separation between `Dataset` and `DatasetPipeline`. 
+3. In Ray 2.4, Ray Datasets will support a `"zero_copy"` batch format, allowing the core Ray Data to decide the most optimal batch format to use.
+
+With these changes, certain API choices that were made for the AIR batch prediction API are no longer the most optimal for simplicity and usability.
+1. Previously, AIR would manually fuse preprocessing and prediction stages, requiring `BatchPredictor` to be aware of the preprocessors passed in via a `Checkpoint`. This is not needed anymore with Datasets lazy by default, as the fusion logic can be pushed down to the Datasets layer.
+2. Previously, the separation of `Dataset` and `DatasetPipeline` resulted in `BatchPredictor` needing 2 analgous APIs: `predict` and `predict_pipelined`. This is not needed anymore with streaming as default execution.
+3. Previously, the logic for determining batch format was decided in the AIR layer in `Preprocessor` and `BatchPredictor`. This is not needed anymore with zero-copy batch format, as this decision making can be pushed down to the Datasets layer.
+
+### Should this change be within `ray` or outside?
+main `ray` project. Changes are made to Ray Data and Ray AIR level.
+
+## Stewardship
+### Required Reviewers
+The proposal will be open to the public, but please suggest a few experience Ray contributors in this technical domain whose comments will help this proposal. Ideally, the list should include Ray committers.
+
+@ericl, @clarkzinzow, @c21, @gjoliver
+
+### Shepherd of the Proposal (should be a senior committer)
+To make the review process more productive, the owner of each proposal should identify a **shepherd** (should be a senior Ray committer). The shepherd is responsible for working with the owner and making sure the proposal is in good shape (with necessary information) before marking it as ready for broader review.
+
+@ericl
+
+## Design and Architecture
+
+## API Changes
+
+1. Introduce `map_preprocessor` API directly to Ray Datasets.
+
+```python
+def map_preprocessor(preprocessor: Preprocessor, batch_size: Optional[int], prefetch_batches: Optional[int]):
+  """Apply the transform of the provided preprocessor to batches of data.
+
+  Args:
+    preprocessor: A :class:`~ray.data.preprocessor.Preprocessor` used to transform batches of data. The provided preprocessor is
+      expected to already be fitted if it is fittable.
+    batch_size: Same as `map_batches`
+    prefetch_batches: Same as `map_batches`
+  """
+```
+
+2. Introduce a `map_predictor` API directly to Ray Datasets.
+
+```python
+def map_predictor(predictor: Union[Predictor, Callable[[], Predictor]], 
+                  batch_size: Optional[int], 
+                  feature_columns: Optional[List[str]] = None, 
+                  keep_columns: Optional[List[str]] = None, 
+                  prefetch_batches: Optional[int] = 0, 
+                  min_scoring_workers: int = 1,
+                  max_scoring_workers: Optional[int] = None,
+                  num_cpus_per_worker: Optional[int] = None,
+                  num_gpus_per_worker: Optional[int] = None, 
+                  **predict_kwargs):
+  """Predict on batches of data.
+
+  Args:
+    predictor: An instance of a Predictor to use for prediction. For Predictors that are not serializable, or are loading large models
+      or that are created from AIR Checkpoints, a function that creates a Predictor can be passed in instead. 
+    batch_size: Batch size to use for prediction.
+    feature_columns: List of columns in the preprocessed dataset to use for
+      prediction. Columns not specified will be dropped
+      from `data` before being passed to the predictor.
+      If None, use all columns in the preprocessed dataset.
+    keep_columns: List of columns in the preprocessed dataset to include
+        in the prediction result. This is useful for calculating final
+        accuracies/metrics on the result dataset. If None,
+        the columns in the output dataset will contain
+        just the prediction results.
+    min_scoring_workers: Minimum number of scoring actors.
+    max_scoring_workers: If set, specify the maximum number of scoring actors.
+    num_cpus_per_worker: Number of CPUs to allocate per scoring worker.
+      Set to 1 by default.
+    num_gpus_per_worker: Number of GPUs to allocate per scoring worker.
+      Set to 0 by default. If you want to use GPUs for inference, please
+      specify this parameter.
+    ray_remote_args: Additional resource requirements to request from ray.
+    predict_kwargs: Keyword arguments passed to the predictor's
+        ``predict()`` method.
+```
+
+3. Deprecate `Chain` Preprocessor, as chains can be represented via multiple `map_preprocessor` calls.
+4. Deprecate `BatchMapper`, as the same functionality can be achieved via `map_batches`.
+      
+## Example: Multi-stage Inference Pipeline from pre-trained model
+
+Note that Checkpoints are not on the critical path, but can still be used.
+
+### Before
+```python
+preprocessor1 = BatchMapper(lambda x: x+1)
+preprocessor2 = TorchVisionPreprocessor(torchvision.transforms.Crop(224, 224))
+chained = Chain([preprocessor1, preprocessor2])
+
+model = pretrained_resnet()
+checkpoint = TorchCheckpoint.from_model(model, preprocessor=chained)
+batch_predictor = BatchPredictor.from_checkpoint(checkpoint)
+
+ds = ds.read_parquet("s3://data")
+predictions = batch_predictor.predict(ds)
+```
+
+### After
+```python
+model = pretrained_resnet()
+predictor = TorchPredictor(model)
+
+# Alternatively, delayed model creation
+# predictor = lambda: TorchPredictor(pretrained_resnet())
+
+# Alternatively, create from AIR Checkpoint.
+# predictor = lambda: TorchPredictor.from_checkpoint(air_checkpoint)
+
+predictions = ds.read_parquet("s3://data")
+    .map_batches(lambda x: x+1)
+    .map_preprocessor(TorchVisionPreprocessor(torchvision.transforms.Crop(224, 224)))
+    .map_predict(predictor)
+```
+
+## Compatibility, Deprecation, and Migration Plan
+An important part of the proposal is to explicitly point out any compability implications of the proposed change. If there is any, we should thouroughly discuss a plan to deprecate existing APIs and migration to the new one(s).
+
+Ray 2.4: Onboard new users onto new APIs
+- Introduce `map_preprocessor` and `map_predictor` as Alpha APIs.
+- Rewrite all batch inference examples to use these APIs, and use `map_batches` directly instead of `BatchMapper` and `Chain`.
+- `BatchPredictor` will still be fully supported.
+
+Ray 2.5: Start migrating existing users to new APIs
+- Soft Deprecate `BatchPredictor`. Warn if this API is used, and remove from Documentation.
+- Add migration guide in documentation for guidance on how to migrate to the new APIs.
+
+Ray 2.6: Deprecate old APIs
+- Hard deprecate `BatchPredictor`.
+
+
+## Test Plan and Acceptance Criteria
+The proposal should discuss how the change will be tested **before** it can be merged or enabled. It should also include other acceptance criteria including documentation and examples. 
+
+- Unit and integration for new APIs
+- Documentation and examples on new API.
--- a/reps/2023-03-10-batch-inference.md
+++ b/reps/2023-03-10-batch-inference.md
@@ -384,9 +384,11 @@ def map_batches(
 
 ## FAQ
 1. If I pass in a `Predictor` directly, won't initiliazation happen for every batch?
+
 No. `Predictors` will require the use of a Callable Class, and so that we wrap the Predictor in a Callable class internally, like in the example above.
 
 2. If I have an AIR Checkpoint, won't this require initialization on the driver?
+
 No. If a Predictor is created from an AIR Checkpoint, we delay the initialization to a `setup` call that happens only in the 
 actor.
 
--- a/reps/2023-03-10-batch-inference.md
+++ b/reps/2023-03-10-batch-inference.md
@@ -294,6 +294,7 @@ class BatchMappable(Callable[[T], U]):
 
     def use_callable_class(self) -> bool:
         """Indicates whether this `BatchMappable` should use a callable class."""
+        return False
 ```
 
 `map_batches` will support instances of `BatchMappable` as a supported UDF
@@ -319,25 +320,36 @@ class Predictor:
         return True
 ```
 
-3. When using Predictors with AIR Checkpoints, move the construction logic to `setup` so that it can be called within the actor.
+3. When using Predictors with AIR Checkpoints, add an option for lazy instantiation. This allows us to load from the Checkpoint when inside the actor instead of on the driver.
 
 ```python
 class Predictor:
-    def _to_checkpoint(self) -> Checkpoint:
-        """Create a serializable Checkpoint from this Predictor.
-
-        Returns:
-            Checkpoint containing the state for this Predictor.
-        """
+    def __init__(self, ...):
+        self._initialized = False
         ...
-    
+
     def setup(self):
-        if self._checkpoint:
-            # Initialize the state from the Checkpoint.
+        if not self._initialized 
+            if self._checkpoint:
+                # Initialize the state from the Checkpoint.
+                self.__setstate__(self._checkpoint)
+            self._initialized = True
     
     @classmethod
-    def from_checkpoint(cls, checkpoint):
-        return Predictor(_checkpoint=checkpoint)
+    def from_checkpoint(cls, checkpoint: Checkpoint, lazy: bool = False):
+        if not lazy:
+            # Current logic for creating Predictor from checkpoint.
+            predictor = Predictor(...)
+            predictor._initialized = True
+            return predictor
+        else:
+            return Predictor(_checkpoint=checkpoint)
+
+    def predict(self, batch):
+        if not self._initialized:
+            # Initialize the Predictor if not already done so.
+            self.setup()
+        ...
 
     def use_callable_class(self) -> bool:
         return True
@@ -372,6 +384,8 @@ def map_batches(
             class _Wrapper:
                 def __init__(self, batch_mappable_ref: ObjectRef[BatchMappable]):
                     self.batch_mappable = ray.get(batch_mappable_ref)
+                    # Setup the BatchMappable.
+                    self.batch_mappable.setup()
                 
                 def __call__(self, *args, **kwargs):
                     return self.batch_mappable(*args, **kwargs)
--- a/reps/2023-03-10-batch-inference.md
+++ b/reps/2023-03-10-batch-inference.md
@@ -298,7 +298,7 @@ class BatchMappable(Callable[[T], U]):
 
 `map_batches` will support instances of `BatchMappable` as a supported UDF
 
-2. Add a private `to_checkpoint` serializer for `Predictor`, analagous to the current `from_checkpoint` deserializer. Also, move the 
+2. Add a private `to_checkpoint` serializer for `Predictor`, analagous to the current `from_checkpoint` deserializer. 
 
 ```python
 class Predictor:
--- a/reps/2023-03-10-batch-inference.md
+++ b/reps/2023-03-10-batch-inference.md
@@ -88,9 +88,7 @@ def map_predictor(predictor: Union[Predictor, Callable[[], Predictor]],
         ``predict()`` method.
   """
 ```
-
-3. Deprecate `Chain` Preprocessor, as chains can be represented via multiple `map_preprocessor` calls.
-4. Deprecate `BatchMapper`, as the same functionality can be achieved via `map_batches`.
+3. Deprecate `BatchMapper`, as the same functionality can be achieved via `map_batches`.
       
 ## Example: Multi-stage Inference Pipeline from pre-trained model
 
--- a/reps/2023-03-10-batch-inference.md
+++ b/reps/2023-03-10-batch-inference.md
@@ -124,7 +124,7 @@ predictor = TorchPredictor(model)
 predictions = ds.read_parquet("s3://data")
     .map_batches(lambda x: x+1)
     .map_preprocessor(TorchVisionPreprocessor(torchvision.transforms.Crop(224, 224)))
-    .map_predict(predictor)
+    .map_predictor(predictor)
 ```
 
 ## Compatibility, Deprecation, and Migration Plan
--- a/reps/2023-03-10-batch-inference.md
+++ b/reps/2023-03-10-batch-inference.md
@@ -86,6 +86,7 @@ def map_predictor(predictor: Union[Predictor, Callable[[], Predictor]],
     ray_remote_args: Additional resource requirements to request from ray.
     predict_kwargs: Keyword arguments passed to the predictor's
         ``predict()`` method.
+  """
 ```
 
 3. Deprecate `Chain` Preprocessor, as chains can be represented via multiple `map_preprocessor` calls.
--- a/reps/2023-03-10-batch-inference.md
+++ b/reps/2023-03-10-batch-inference.md
@@ -385,7 +385,7 @@ def map_batches(
 ## FAQ
 1. If I pass in a `Predictor` directly, won't initiliazation happen for every batch?
 
-No. `Predictors` will require the use of a Callable Class, and so that we wrap the Predictor in a Callable class internally, like in the example above.
+No. `Predictors` will require the use of a Callable Class, and so we wrap the Predictor in a Callable class internally, like in the example above.
 
 2. If I have an AIR Checkpoint, won't this require initialization on the driver?
 
--- a/reps/2023-03-10-batch-inference.md
+++ b/reps/2023-03-10-batch-inference.md
@@ -269,7 +269,7 @@ class BatchMappable(Callable[[T], U]):
         """
         pass
        
-    def validate_batch_format(self, batch_format: Optional[str]) -> Optional[str]:
+    def get_batch_format(self, batch_format: Optional[str]) -> Optional[str]:
         """Validate that the provided `batch_format` is compatible with this `BatchMappable`.
 
         Raises an error if the provided `batch_format` is not supported with this `BatchMappable` 
--- a/reps/2023-03-10-batch-inference.md
+++ b/reps/2023-03-10-batch-inference.md
@@ -30,74 +30,88 @@ To make the review process more productive, the owner of each proposal should id
 
 ## Design and Architecture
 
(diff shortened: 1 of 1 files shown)


reps/2023-03-10-batch-inference.md
```diff
@@ -0,0 +1,150 @@
...
+# Alternatively, delayed model creation
+# predictor = lambda: TorchPredictor(pretrained_resnet())
+
+# Alternatively, create from AIR Checkpoint.
+# predictor = lambda: TorchPredictor.from_checkpoint(air_checkpoint)
+
+predictions = ds.read_parquet("s3://data")
+    .map_batches(lambda x: x+1)
+    .map_preprocessor(TorchVisionPreprocessor(torchvision.transforms.Crop(224, 224)))
+    .map_predict(predictor)
```
ericl, line 127:
```suggestion
    .map_predictor(predictor)
```

reps/2023-03-10-batch-inference.md
```diff
@@ -0,0 +1,150 @@
...
+    num_gpus_per_worker: Number of GPUs to allocate per scoring worker.
+      Set to 0 by default. If you want to use GPUs for inference, please
+      specify this parameter.
+    ray_remote_args: Additional resource requirements to request from ray.
+    predict_kwargs: Keyword arguments passed to the predictor's
+        ``predict()`` method.
+  """
+```
+
+3. Deprecate `Chain` Preprocessor, as chains can be represented via multiple `map_preprocessor` calls.
```
ericl, line 92: For sake of keeping the scope narrow, I think we should not alter the preprocessor APIs in this REP.

reps/2023-03-10-batch-inference.md
```diff
@@ -0,0 +1,150 @@
...
+    .map_preprocessor(TorchVisionPreprocessor(torchvision.transforms.Crop(224, 224)))
+    .map_predict(predictor)
+```
+
+## Compatibility, Deprecation, and Migration Plan
+An important part of the proposal is to explicitly point out any compability implications of the proposed change. If there is any, we should thouroughly discuss a plan to deprecate existing APIs and migration to the new one(s).
+
+Ray 2.4: Onboard new users onto new APIs
+- Introduce `map_preprocessor` and `map_predictor` as Alpha APIs.
+- Rewrite all batch inference examples to use these APIs, and use `map_batches` directly instead of `BatchMapper` and `Chain`.
```
ericl, line 135: This breaks training->serving compatibility, so I think we should punt.

reps/2023-03-10-batch-inference.md
```diff
@@ -0,0 +1,150 @@
...
+        ``predict()`` method.
+  """
+```
+
+3. Deprecate `Chain` Preprocessor, as chains can be represented via multiple `map_preprocessor` calls.
+4. Deprecate `BatchMapper`, as the same functionality can be achieved via `map_batches`.
+      
+## Example: Multi-stage Inference Pipeline from pre-trained model
+
+Note that Checkpoints are not on the critical path, but can still be used.
```
ericl, line 97: The phrasing here is a bit confusing... are you saying that basically users don't need to know about Checkpoints in order to do batch inference with {AIR / custom predictors} in the new proposal? That seems like a win.

reps/2023-03-10-batch-inference.md
```diff
@@ -0,0 +1,150 @@
...
+      expected to already be fitted if it is fittable.
+    batch_size: Same as `map_batches`
+    prefetch_batches: Same as `map_batches`
+  """
+```
+
+2. Introduce a `map_predictor` API directly to Ray Datasets.
+
+```python
+def map_predictor(predictor: Union[Predictor, Callable[[], Predictor]],
```
ericl, line 54: Maybe it should always be a lambda. I'm not sure if most predictors are copyable.

reps/2023-03-10-batch-inference.md
amogkam: This is just changing the batch inference examples to the new API. The existing training->Ray Serve workflow remains the same?

reps/2023-03-10-batch-inference.md
amogkam: removed

reps/2023-03-10-batch-inference.md
amogkam: updated

reps/2023-03-10-batch-inference.md
ericl: Got it. Agree batch inference specific workloads can avoid the Preprocessor class entirely.

reps/2023-03-10-batch-inference.md
```diff
@@ -51,7 +51,7 @@ def map_preprocessor(preprocessor: Preprocessor, batch_size: Optional[int], pref
 2. Introduce a `map_predictor` API directly to Ray Datasets.
 
 ```python
-def map_predictor(predictor: Union[Predictor, Callable[[], Predictor]], 
+def map_predictor(predictor: Callable[[], Predictor]],
```
ericl, line 54:
```suggestion
def map_predictor(create_predictor: Callable[[], Predictor]], 
```

Not sure if we have an existing naming convention for factories.

reps/2023-03-10-batch-inference.md
```diff
@@ -0,0 +1,143 @@
...
+      Set to 1 by default.
+    num_gpus_per_worker: Number of GPUs to allocate per scoring worker.
+      Set to 0 by default. If you want to use GPUs for inference, please
+      specify this parameter.
+    ray_remote_args: Additional resource requirements to request from ray.
+    predict_kwargs: Keyword arguments passed to the predictor's
+        ``predict()`` method.
+  """
+```
+3. Deprecate `BatchMapper`, as the same functionality can be achieved via `map_batches`.
```
ericl, line 90: Also remove this as it's needed for Train->Serve?

reps/2023-03-10-batch-inference.md
```diff
@@ -0,0 +1,143 @@
...
+ds = ds.read_parquet("s3://data")
+predictions = batch_predictor.predict(ds)
+```
+
+### After
+```python
+predictor = lambda: TorchPredictor(pretrained_resnet())
+
+# Alternatively, create from AIR Checkpoint.
+# predictor = lambda: TorchPredictor.from_checkpoint(air_checkpoint)
```
ericl, line 115: Can you also add an example of how you'd implement a custom predictor? (hopefully it is no harder than using a custom callable cls)

ericl: I think in general we should be of the mindset that custom predictor/etc. should be presented as first-class in our APIs, and we only have framework-specific helpers like TorchPredictor as shortcuts. This avoids the problem where users do not know how to find the "escape hatch".

reps/2023-03-10-batch-inference.md
```diff
@@ -0,0 +1,143 @@
...
+
+predictions = ds.read_parquet("s3://data")
+    .map_batches(lambda x: x+1)
+    .map_preprocessor(TorchVisionPreprocessor(torchvision.transforms.Crop(224, 224)))
```
gjoliver, line 119: like the new dataset-centric way of doing things.
a stupid question, why do we have to differentiate preprocessor vs predictor?
can they all simply use the map_batches() api?
```
predictions = ds.read_parquet("s3://data")
    .map_batches(lambda x: x+1)
    .map_batches(TorchVisionPreprocessor(torchvision.transforms.Crop(224, 224)))
    .map_batches(TorchPredictor(pretrained_resnet()))
```
?

amogkam: 1. Avoid overloading the existing `map_batches` API and make the API more explicit
2. Not all of the `map_batches` arguments are compatible or needed for preprocessors/predictors.

reps/2023-03-10-batch-inference.md
amogkam: good point, removed

reps/2023-03-10-batch-inference.md
amogkam: added

reps/2023-03-10-batch-inference.md
amogkam: added

reps/2023-03-10-batch-inference.md
```diff
@@ -120,6 +119,110 @@ predictions = ds.read_parquet("s3://data")
...
+```python
+class MXNetPredictor(Predictor):
+    def __init__(
+      self,
+      net: gluon.Block,
+    ):
+        self.net = net
+        super().__init__()
+
+    def _predict_numpy(
```
ericl, line 196: Hmm this is a bit confusing, since you specify "numpy" both here and in the batch format arg. If one is mismatched, I guess you get an error?

Maybe it would be better to have just recommending overriding the predict() method.

amogkam: oh no, the `batch_format` arg is specified in `map_batches` for the preprocessing logic. `map_predictor` doesn't take in a batch format.

richardliaw:
Usability detail- it may be easy for a user to shoot themselves in a foot by serializing a large model into the scope of the function.

Can we make sure we have a section in the docs + an appropriate error message for this failure mode?

richardliaw:
Would it make sense for map_predictor to take in a predictor instead of a lambda currying a predictor?

amogkam:
Yeah one option can be to allow both:
1. Pass in a Predictor instance. This works for small models, etc. for better usability
2. Pass in a predictor_creator for large models or Predictors created from AIR Checkpoints. 

We can fail if attempting to serialize a very large model via the same mechanism as `tune.with_parameters`

reps/2023-03-10-batch-inference.md
ericl: I guess I'm saying--- in this example (and other examples we create), why not override the plain `predict` class instead? That way, the term "numpy" shows up only once, and the user doesn't need to understand the automatic batch format choosing magic in the Predictor class.

reps/2023-03-10-batch-inference.md
```diff
@@ -0,0 +1,246 @@
...
+## Summary
+Reduce the friction for batch inference of pre-trained by updating the APIs to do the following:
+1. Remove AIR Checkpoints from the critical path for batch prediction.
+2. Simpler interface for chaining multiple preprocessing and prediction stages.
+
+### General Motivation
+There are changes to Ray Datasets in 2.3/2.4:
+1. Ray Datasets became lazy by default in Ray 2.3. Execution is no longer triggered until consumption is explicitly called.
+2. In Ray 2.4, Ray Datasets will move to streaming execution by default, resulting in there no longer being a separation between `Dataset` and `DatasetPipeline`. 
+3. In Ray 2.4, Ray Datasets will support a `"zero_copy"` batch format, allowing the core Ray Data to decide the most optimal batch format to use.
```
ericl, line 12: We don't show how this is used in the REP--- should we either remove this line, or incorporate it into the REP?

reps/2023-03-10-batch-inference.md
amogkam: Not sure if I'm understanding...we still want to provide both a numpy path and a pandas path right?

When implementing a custom predictor, the user can pick one (or both) to implement. The term numpy is still showing up only in the Predictor. `map_predictor` does not take in a `batch_format` argument.

reps/2023-03-10-batch-inference.md
amogkam: removed

reps/2023-03-10-batch-inference.md
ericl: Stepping back, I'm kind of saying the entire way we have designed this is wrong...

Predictor shouldn't query dataset for it's "dataset format". The predictor should say, "this is the batch format I need", and the dataset should internally optimize to minimize format conversions.

In this example, the user ideally should be able to simply write:

```
class MyPrep(Preprocessor):
   batch_format = "numpy"
   
   def predict(self, data):
        assert data is in numpy format
```

ericl: In other words: why is this so much more complicated than the map_batches() signature? It shouldn't be.

amogkam: Yes completely agree on what you are saying. But I don't think the API needs to change to achieve this.

> The predictor should say, "this is the batch format I need"

This is achieved by the Predictor either implementing `_predict_pandas` or `_predict_numpy`.

I think the question is should this distinction be in the form of 2 separate methods, or overloading onto a single `predict` method, and having a `batch_format` class attribute.

Personally, I slightly prefer the former just to be more explicit.

amogkam: Either way, we will be removing the dataset format querying. But this is more of an implementation change than an API change.

ericl: OK, it seems fine for now if we remove batch format from map_predictor().

ericl (approved):

ericl:
LGTM. Feel free to submit to the list for a vote.

zhe-thoughts:
High level question: for pre-trained models, what are the main benefits of using `map_preprocessor` and `map_predictor` vs. directly using `map_batches`?

@amogkam @ericl  Thanks

amogkam:
Mainly so that we can provide certain guardrails for users when they use these APIs so they don't shoot themselves in the foot. Also, gives us APIs to add more prediction specific optimizations.

You can see some prior discussion here https://github.com/ray-project/ray/issues/32929#issuecomment-1450573546

ericl:
Re: `map_batches` vs `map_preprocessor` + `map_predictor`, there has been consistent feedback that this is somewhat confusing, because there are now multiple methods that do something very similar. We might be able to achieve the original goal of providing guardrails in other ways.

For _preprocessor_, I'd propose going back to the original idea of implementing `__call__` on Preprocessor classes, and requiring users to call it as follows:

```
ds.map_batches(preprocessor)
```

To choose a good batch format for the preprocessor, Datasets can call `preprocessor.preferred_batch_format()` (this can be a generic interface for specifying batch format on callable) to choose the default batch format. We can also warn at runtime if the user chooses a batch format that isn't preferred.

For _predictor_, we can also implement `__call__` and ensure predictors can be serialized in the common case. For delayed construction of predictor, we can tell users to write a callable class wrapper:

```
# Note: an error should be raised if you don't specify actor pool strategy
ds.map_batches(predictor, compute=ray.data.ActorPoolStrategy(8, 8))

-or-

class MyWrapper:
   def __init__(self):
        self.predictor = MyPredictor(args)

   def __call__(self, batch):
       return self.predictor.predict(batch)

# Note: an error should be raised if you don't specify actor pool strategy
ds.map_batches(MyWrapper, compute=ray.data.ActorPoolStrategy(8, 8))
```

We can also rely on `preferred_batch_format()` for the predictor to choose a good batch format / print a warning to the user if a sub-optimal batch format is used.

Thoughts?

amogkam:
Yeah I'm supportive of this, and goes back to the initial proposal.

I would like to figure out the details of `batch_format` some more though. In your proposal, how would you distinguish between sub-optimal vs. not suboptimal batch format?

Is it just 
```python
if batch_format != preprocessor.determine_transform_to_use():
    raise ValueError(...)
```

ericl:
> I would like to figure out the details of batch_format some more though. In your proposal, how would you distinguish between sub-optimal vs. not suboptimal batch format?

I was thinking that first, Datasets can query `.preferred_batch_format()` on the class passed to it. Within the Preprocessor, in the `__call__` you can check `isinstance(batch, my_preferred_type)`. If it's a mismatch, log once that the batch format is not optimal.

amogkam:
Yeah that sounds good to me. For the wrapped Predictor case, where we don't have direct access to the instance, we would have to directly rely on the runtime check for these warnings.

ericl:
copying from committer list: REP approved, conditional on changes above.

gjoliver (approved):

c21:
IMO for option 5: this `withColumns()` API semantic is quite complicated… (this would probably confuse people a lot coming from Data background as me). 

I vote for option 1: `TorchPredictor(model, feature_columns=["feature"], keep_columns=["label"])`, and we are doing all the stuff under the hood inside `map_batches`. If no preprocessor defined in checkpoint, we can also push down the feature and keep columns as optimizer rule.

ericl:
I'm in favor of Option 2/3, since I think a little bit of extra boilerplate code is easier to understand and adapt for new users than opaque API args. It's also easy to go from Option 2 to Option 1 at a later point in time if this turns out to be a common pain point, but otherwise harder to go the other direction.

I also don't think Option 3 requires deprecating anything, since we already require users to construct a custom callable if they are delaying the construction of their Predictor. We would just be extending this to tell users to construct a custom callable if they need either delayed construction or this label passthrough.

reps/2023-03-10-batch-inference.md
```diff
@@ -227,28 +237,176 @@ def preprocess(batch: Dict[str, np.ndarray]) -> Dict[str, np.ndarray]:
...
+
+    def validate_compute_strategy(self, strategy):
+        """Validate that the provided `strategy` is compatible with this `BatchMappable`.
+
+        For example, enforce that `ActorPoolStrategy` is provided for Predictors.
+        """
+        pass
+
+    def use_callable_class(self) -> bool:
+        """Indicates whether this `BatchMappable` should use a callable class."""
```
ericl, line 296: Should this always be true?

reps/2023-03-10-batch-inference.md
```diff
@@ -227,28 +237,176 @@ def preprocess(batch: Dict[str, np.ndarray]) -> Dict[str, np.ndarray]:
...
+class Predictor:
+    def _to_checkpoint(self) -> Checkpoint:
+        """Create a serializable Checkpoint from this Predictor.
+
+        Returns:
+            Checkpoint containing the state for this Predictor.
+        """
+        ...
+    
+    def setup(self):
```
ericl, line 334: Is the idea this is called on first use of the predictor? Should it be internal?

reps/2023-03-10-batch-inference.md
```diff
@@ -227,28 +237,176 @@ def preprocess(batch: Dict[str, np.ndarray]) -> Dict[str, np.ndarray]:
...
+```
+
+## FAQ
+1. If I pass in a `Predictor` directly, won't initiliazation happen for every batch?
+
+No. `Predictors` will require the use of a Callable Class, and so we wrap the Predictor in a Callable class internally, like in the example above.
+
+2. If I have an AIR Checkpoint, won't this require initialization on the driver?
+
+No. If a Predictor is created from an AIR Checkpoint, we delay the initialization to a `setup` call that happens only in the
```
ericl, line 392: Can we make this more explicit somehow? Also, we shouldn't expose `setup` publicly right?

reps/2023-03-10-batch-inference.md
```diff
@@ -227,28 +237,176 @@ def preprocess(batch: Dict[str, np.ndarray]) -> Dict[str, np.ndarray]:
...
     return batch
 
-predictor_creator = lambda: MXNetPredictor(gluon.model_zoo.vision.resnet50_v1(pretrained=True))
-dataset.map_batches(preprocess, batch_format="numpy").map_predictor(predictor_creator)
+predictor = MXNetPredictor(gluon.model_zoo.vision.resnet50_v1(pretrained=True))
+dataset.map_batches(preprocess, batch_format="numpy").map_batches(predictor_creator, compute=ray.data.ActorPoolStrategy(size=8))
 ```
 
+## Internal interfaces & implementation details
+1. Introduce a DeveloperAPI `BatchMappable` interface that `Preprocessor` and `Predictor` subclass. This interface will define the following:
```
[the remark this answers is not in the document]
amogkam: any suggestions?

reps/2023-03-10-batch-inference.md
amogkam: Idts, for Preprocessors we don't want to use callable classes.

But for Predictors, yes, should always be True

reps/2023-03-10-batch-inference.md
amogkam: It should never need to be called by users.

But, inside `map_batches` it is called here: https://github.com/ray-project/enhancements/pull/25/files#diff-0b39e135189387c4d60ea6f6d574d05c99eeb634ef29e2abae5a92442ce326fdR388

reps/2023-03-10-batch-inference.md
```diff
@@ -0,0 +1,428 @@
...
+model = resnet50()
+predictor = TorchPredictor(model=model)
+
+# Alternatively, create from AIR Checkpoint.
+# predictor = TorchPredictor.from_checkpoint(air_checkpoint)
+
+predictions = ds.map_batches(predictor, compute=ray.data.ActorPoolStrategy(size=8))
+```
+
+Exception is raised if ActorPoolStrategy is not specified.
```
pcmoritz, line 57: What is the rationale behind this? As a user I would prefer if I can do
```
predictions = ds.map_batches(predictor)
```
and it does the right thing. Why do I need to care about ActorPoolStrategy?

amogkam: The default `map_batches` API is TaskPoolStrategy.

If we want to make `ActorPoolStrategy` the default, then we'd have to introduce a new API (like `map_predictor`) which we decided not to do in order to have a single unified API for data processing.

reps/2023-03-10-batch-inference.md
ericl: I don't think we need this though right? For non-lazy Predictors, there is no setup. For lazy predictors, you can lazily do the setup on the first call.

Hence, we can eliminate this API.

reps/2023-03-10-batch-inference.md
```diff
@@ -0,0 +1,428 @@
...
+
+        Args:
+            batch_format: The `batch_format` that is provided to `map_batches`.
+
+        Returns:
+            The batch format to use.
+        """
+       return None
+
+    def validate_compute_strategy(self, strategy):
```
ericl, line 288: Shall we have a generic validate_args(kwargs) function instead of adding them one at a time?

reps/2023-03-10-batch-inference.md
```diff
@@ -0,0 +1,428 @@
...
+
+    def setup(self):
+        """Additional setup logic to run.
+
+        If a callable class is being used, this function will be run in the `__init__`
+        of the Callable class.
+        """
+        pass
+       
+    def validate_batch_format(self, batch_format: Optional[str]) -> Optional[str]:
```
ericl, line 272:
```suggestion
    def get_batch_format(self, batch_format: Optional[str]) -> Optional[str]:
```

(It's not just validating)

ericl (commented):
I have some blocking comments on BatchMappable. From the process perspective though, it's not critical that we resolve BatchMappable implementation details here--- we could instead say this is to be decided during PR review, and remove the specific detail from this REP.

reps/2023-03-10-batch-inference.md
```diff
@@ -0,0 +1,428 @@
...
+
+#### Before
+```python
+preprocessor1 = BatchMapper(lambda x: x+1)
+preprocessor2 = TorchVisionPreprocessor(torchvision.transforms.Crop(224, 224))
+chained = Chain([preprocessor1, preprocessor2])
+
+model = pretrained_resnet()
+checkpoint = TorchCheckpoint.from_model(model, preprocessor=chained)
+batch_predictor = BatchPredictor.from_checkpoint(checkpoint)
```
ericl, line 122: I think we want to support a `lazy=True` flag here, right? Otherwise, this wouldn't support deferred loading and/or may be confusing for users if it defaulted to lazy loading.

zhe-thoughts:
@amogkam could you do DCO https://www.secondstate.io/articles/dco/

reps/2023-03-10-batch-inference.md
amogkam: How would that work?

ericl: We could have a validator like

```
def validate_args(self, **kwargs):
   pass
```

And Datasets could pass in a kwargs dict like `{"compute": ..., "batch_size": ..., ...}`.

reps/2023-03-10-batch-inference.md
amogkam: It would be preferable to call this separately here: https://github.com/ray-project/enhancements/pull/25/files#diff-0b39e135189387c4d60ea6f6d574d05c99eeb634ef29e2abae5a92442ce326fdR388

Otherwise, we can't start the initialization until there are actual prediction tasks to dispatch.

This is purely a developer facing API, so has no user impact.

ericl: I see, that makes sense (create as soon as actor is created). Would be good to add this as a motivation for this API.

reps/2023-03-10-batch-inference.md
amogkam: This is the old BatchPredictor code path. This doesn't need to change right?

ericl: Ah, right, I mean to comment on the new path.

reps/2023-03-10-batch-inference.md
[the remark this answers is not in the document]
ericl: BatchUDF?

reps/2023-03-10-batch-inference.md
amogkam: Got it. I think we should just make lazy the default, so that users don't always need to do `lazy=True` for batch prediction.

ericl: I'm ok with that. You had a concern earlier that this would make single-process prediction benchmarks misleading though right? Since predict() would trigger loading and that could throw off timing of scripts.

reps/2023-03-10-batch-inference.md
pcmoritz: If a basic batch inference user needs to know about `ActorPoolStrategy` I think we are failing our users, it is not something the user should know about (this is an abstraction leak).

amogkam: @pcmoritz your recommendation is go with a separate `map_predictor` API?

ericl: I don't think it's very easy to avoid this abstraction leak *entirely*. Here are a few things we've learned from previous users of this API:
- No one uses the highest level BatchPredictor API, which does try to hide most of underlying details.
- Some users have expressed that they prefer lower-level APIs (such as Ray core), because they understand better what Ray is doing under the hood.

As for creating a special `map_predictor()` method, I am OK with that, but we have gotten strong feedback that it's confusing why there are two very similar map methods. However, upon reflection we may be able to address this concern by explicitly documenting this as the syntactic sugar / simpler form of `map_batches()` for batch inference, similar to how `map` is sugar for `map_batches`.

reps/2023-03-10-batch-inference.md
amogkam: Yeah on second thought, I don't think it's that big of a concern, and will just make the API more confusing.

amogkam: removed the `lazy` arg and made lazy by default

reps/2023-03-10-batch-inference.md
amogkam: Clarified in the docstring for `setup`

reps/2023-03-10-batch-inference.md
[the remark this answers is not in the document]
amogkam: that's already taken by https://github.com/ray-project/ray/blob/master/python/ray/data/block.py#L115 😅

reps/2023-03-10-batch-inference.md
amogkam: I think it's fine for now since it's just 2 (batch format, compute strategy). If we add more, then we can look into consolidating in the future.

reps/2023-03-10-batch-inference.md
zhe-thoughts: It's a non-trivial tradeoff. For now my preference is to go with `map_batches` with `ActorPoolStrategy`. If we see enough user feedback that a separate `map_predictor` is acceptable, we can do that down the road. I feel it is an incremental path. @pcmoritz what do you think?

Another reason why I prefer this sequence is, for really simple use inference use cases, "it works" without the `ActorPoolStrategy` argument, just slower. @amogkam what is the rough performance difference with vs. without it?

pcmoritz: Why not use `map_batches` and if the argument is a class (like an actor class or Predictor class), it will use the `ActorPoolStrategy`? I don't see the tradeoff here. It can have a reasonable default like "auto determine the number of actors" and if people want to override, they can specify the `ActorPoolStrategy`.

ericl: There is the tradeoff between explicitness/understandability and ease of getting started.

If we go with auto determining the number of actors (perhaps with a warning/tip on how to configure it), most users will probably start with this code path.

However if we raise an error and tell the user to specify ActorPoolStrategy(size=N), this starts users down the more explicit and understandable code path, though it adds an additional onboarding step.

The current API is preferring the more explicit route, because (1) users have tended to prefer explicit / easy to understand solutions, and (2) autoscaling introduces performance questionmarks (e.g., slow scale up) and/or complexities (e.g., predictive autoscaling) that can trip up new users trying to get something basic to work.

amogkam: This REP is just being consistent with the current behavior (https://github.com/ray-project/ray/blob/master/python/ray/data/dataset.py#L342-L352) and is not making a stance on whether there should be a default path for ActorPoolStrategy.

Whether there should be a default path for ActorPoolStrategy is an important discussion to have, but is orthogonal to this REP. Let's continue this discussion as a separate change on #api-changes. If we do decide to have a default ActorPoolStrategy path, then this REP will be modified accordingly.

pcmoritz: In the `before` code, the user didn't need to specify a number of actors, and now they have to, that looks like a clear regression with this proposal to me.

How was the number of actors determined beforehand?

amogkam: I'm specifically referring to this proposal that you suggested

> Why not use map_batches and if the argument is a class (like a Predictor class or another class that gets called), it will use the ActorPoolStrategy?

which is a more broader change that goes beyond this REP

zhe-thoughts:
Discussed with @amogkam  @ericl @pcmoritz offline. We will go with the API in the current proposal, and try to further simplify it in the future.

zhe-thoughts:
The further improvement for `ActorPool` is tracked here: https://github.com/ray-project/ray/issues/34034

GokuMohandas:
@amogkam sorry to dig this PR back up but are `map_preprocessor` and `map_predictor` under the 2.4 release (as per [this](https://github.com/ray-project/enhancements/blob/128dc3b907ed61532ea6e9c969e44d3fd9c80001/reps/2023-03-10-batch-inference.md#compatibility-deprecation-and-migration-plan)). If not yet, do you know if/how I can apply a predictor to a dataset for batch inference easily (without creating a separate callable class UDF)?

amogkam:
No they are not in 2.4...currently to use a Predictor with Ray Data directly (not using BatchPredictor), you would do it like so

```python
class Wrapper:
    def __init__(self):
        self.predictor = TorchPredictor(...)

    def __call__(self, batch):
        return self.predictor.predict(batch, ...)


ds.map_batches(Wrapper, ...)
```

PR #26, opened 2023-03-16
[AIR/Data] Simplifying Ray Train Ingest APIs for streaming

ericl:
This is the "train" side complement of https://github.com/ray-project/enhancements/pull/25.

The gist of the proposal is to deprecate the preprocessor and object store memory arguments for `ray.train.Trainer`, instead requiring the user to construct their preprocessing pipeline using Dataset APIs only.

### DIFF
--- a/reps/2023-03-15-train-api.md
+++ b/reps/2023-03-15-train-api.md
@@ -31,11 +31,11 @@ To make the review process more productive, the owner of each proposal should id
 
 ## API Changes
 
-1. Introduce a `resource_limits: ExecutionResources(object_store_memory=2 * GiB)` arg to `ray.air.DatasetConfig`. This effectively enables streaming by default, with a limit of 2GiB.
+1. Introduce a `resource_limits: ExecutionResources(object_store_memory=2 * GiB)` arg to `ray.air.DatasetConfig`. This enables streaming by default, with a limit of 2GiB, and deprecates the previous `max_object_store_memory_fraction` argument.
 
 2. Introduce `Dataset.get_logical_plan()`, which returns the logical plan that can be used to extract the lineage of preprocessors applied to this Dataset. If multiple preprocessors are applied, Train can return a `Chain` of the preprocessors. Non-preprocessor operations on the Dataset are ignored, and we can also allow ignoring preprocessors such as per epoch preprocessors. This function will be used by Train to persist fitted preprocessors with checkpoints.
 
-3. Deprecate the following `DatasetConfig` arguments when streaming is enabled: `max_object_store_memory_fraction`, `global_shuffle`, `fit`, `transform`, `randomize_block_order`, and `per_epoch_preprocessor`, as well as the following `Trainer` arguments: `preprocessor`.
+3. Deprecate the following additional Trainer configs when streaming is enabled: `global_shuffle` and `randomize_block_order` (user to use native Dataset shuffle ops), and the preprocessing args `fit`, `transform`, `preprocessor`, and `per_epoch_preprocessor` (user to setup preprocessing explicitly prior to creating Trainer).
 
 ## Example:
 
--- a/reps/2023-03-15-train-api.md
+++ b/reps/2023-03-15-train-api.md
@@ -33,7 +33,7 @@ To make the review process more productive, the owner of each proposal should id
 
 1. Introduce a `resource_limits: ExecutionResources(object_store_memory=2 * GiB)` arg to `ray.air.DatasetConfig`. This effectively enables streaming by default, with a limit of 2GiB.
 
-2. Introduce `Dataset.get_preprocessor_lineage()`, which returns the lineage of preprocessors applied to this Dataset. If multiple preprocessors are applied, it will return a `Chain` of the preprocessors. Non-preprocessor operations on the Dataset are ignored. This function will be used by Train to persist fitted preprocessors with checkpoints.
+2. Introduce `Dataset.get_logical_plan()`, which returns the logical plan that can be used to extract the lineage of preprocessors applied to this Dataset. If multiple preprocessors are applied, Train can return a `Chain` of the preprocessors. Non-preprocessor operations on the Dataset are ignored, and we can also allow ignoring preprocessors such as per epoch preprocessors. This function will be used by Train to persist fitted preprocessors with checkpoints.
 
 3. Deprecate the following `DatasetConfig` arguments when streaming is enabled: `max_object_store_memory_fraction`, `global_shuffle`, `fit`, `transform`, `randomize_block_order`, and `per_epoch_preprocessor`, as well as the following `Trainer` arguments: `preprocessor`.
 
@@ -132,7 +132,7 @@ While the "after" code is longer, note that all the data processing code is now 
 - A: We will deprecate the former.
 
 - Q: What happens if you apply multiple preprocessors to a Datasets?
-- A: The checkpoint will have the full chain, including per-epoch ones. The user can customize this manually afterwards.
+- A: The checkpoint will have the full chain, including per-epoch ones. Preprocessors can be tagged as used for training only / ignored during inference.
 
 - Q: Why not require the user to do all Data operations outside of Train, including the split?
 - A: This would break tuning, as Train needs to create a separate Data stream per trial. This is not possible post split as calling split is a consumption operation.
--- a/reps/2023-03-15-train-api.md
+++ b/reps/2023-03-15-train-api.md
@@ -135,6 +135,9 @@ While the "after" code is longer, note that all the data processing code is now 
 - Q: What happens if you apply multiple preprocessors to a Datasets?
 - A: The checkpoint will have the full chain, including per-epoch ones. Preprocessors can be tagged as used for training only / ignored during inference by setting an `ignore_for_inference` (constructor) attribute.
 
+- Q: What happens if you apply ordinary functions to the Dataset?
+- A: You'll get a warning that these functions are not captured in the preprocessing chain, and to use BatchMapper if you want that.
+
 - Q: Why not require the user to do all Data operations outside of Train, including the split?
 - A: This would break tuning, as Train needs to create a separate Data stream per trial. This is not possible post split as calling split is a consumption operation.
 
--- a/reps/2023-03-15-train-api.md
+++ b/reps/2023-03-15-train-api.md
@@ -85,13 +85,13 @@ prep = StandardScaler(["f1", "f2"])
 prep.fit(base)
 
 # Apply base preprocessing.
-train_ds = base.map_preprocessor(prep)
+train_ds = base.map_batches(prep)
 train_ds.cache()  # Optional: cache the base data in memory.
 
 # Per-epoch preprocessing.
 train_ds = train_ds \
     .randomize_block_order() \
-    .map_preprocessor(RandomNoisePreprocessor())
+    .map_batches(RandomNoisePreprocessor())
 
 # Trainer doesn't know about preprocessing at all.
 trainer = TorchTrainer(
@@ -128,7 +128,7 @@ While the "after" code is longer, note that all the data processing code is now 
 - Q: Will we still save the preprocessor in the Checkpoint?
 - A: Yes, this doesn't change.
 
-- Q: Should we have both `Preprocessor.transform` and `Dataset.map_preprocessor`?
+- Q: Should we have both `Preprocessor.transform` and `Dataset.map_batches`?
 - A: We will deprecate the former.
 
 - Q: What happens if you apply multiple preprocessors to a Datasets?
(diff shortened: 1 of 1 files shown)


reps/2023-03-15-train-api.md
```diff
@@ -0,0 +1,143 @@
...
+prep.fit(base)
+
+# Apply base preprocessing.
+train_ds = base.map_preprocessor(prep)
+train_ds.cache()  # Optional: cache the base data in memory.
+
+# Per-epoch preprocessing.
+train_ds = train_ds \
+    .randomize_block_order() \
+    .map_preprocessor(RandomNoisePreprocessor())
```
ericl, line 94: Maybe per-epoch operations should be a Trainer arg. On the other hand, it is confusing that the base preprocessor is per-epoch unless you `cache()` it.

reps/2023-03-15-train-api.md
```diff
@@ -0,0 +1,143 @@
...
+base = ray.data.read_parquet("s3://bucket/etl_output")
+fact_table = ray.data.read_csv("s3://bucket/my.csv")
+
+# Fit the preprocessor.
+prep = StandardScaler(["f1", "f2"])
+prep.fit(base)
+
+# Apply base preprocessing.
+train_ds = base.map_preprocessor(prep)
+train_ds.cache()  # Optional: cache the base data in memory.
```
ericl, line 89: This is kind of key--- operations before this are executed once.

reps/2023-03-15-train-api.md
```diff
@@ -0,0 +1,143 @@
...
+trainer = TorchTrainer(
+    model,
+    datasets={
+        "train_ds": train_ds,
+        "fact_table": fact_table,
+    },
+    scaling_config=ScalingConfig(num_workers=4),
+    dataset_config={
+        "train_ds": {
+            "resource_limits": ExecutionResources(
```
amogkam, line 106: is this resource limit only applicable to `train_ds`? do we need a per-dataset level granularity here, or is a global resource limit enough?

reps/2023-03-15-train-api.md
```diff
@@ -0,0 +1,143 @@
...
+    },
+)
+
+# Checkpoint includes fitted preprocessor.
+best_checkpoint = trainer.fit().checkpoint
+assert best_checkpoint.get_preprocessor() == prep
+```
+
+While the "after" code is longer, note that all the data processing code is now cleanly separated from the Trainer, which both a conceptual and practical simplification.
+
```
amogkam, line 119: Another win is that stats from the fitted preprocessor are now accessible from within the training function.

For example, if I use a OrdinalEncoder, I can then access the number of unique categories for my feature to know how large to make my model embedding.

This was not possible to do in the old API where fitting would happen under the hood.

amogkam:
```python
train_ds, val_ds = ds.read_parquet("s3://data").train_test_split(0.2)
ordinal_encoder = OrdinalEncoder()
per_epoch_preprocessor = BatchMapper(lambda x: add_random_noise(x))

def train_func():
   # I cannot access the preprocessor stats inside my training function.
   num_embeddings = ???

trainer = TorchTrainer(train_func, 
    preprocessor=ordinal_encoder, 
    datasets = {"train": train_ds, "val_ds": val_ds}
    dataset_config=
        {"train": DatasetConfig(per_epoch_preprocessor=per_epoch_preprocessor)
        "val": DatasetConfig(transform=False)}
)
trainer.fit()
```

reps/2023-03-15-train-api.md
ericl: Yup. I think per-dataset is probably reasonable, since it would be harder to do global (we don't have that implemented).

amogkam (approved):

reps/2023-03-15-train-api.md
stephanie-wang: I would prefer making it a Trainer arg since we don't want to save it in the checkpoint, and it's not clear that there's a simpler way to specify that.

Maybe we just need a better name here - "training_preprocessor" or something similar to imply that we will only apply it to the training dataset?

reps/2023-03-15-train-api.md
```diff
@@ -0,0 +1,162 @@
...
+## Design and Architecture
+
+## API Changes
+
+1. Introduce a `resource_limits: ExecutionResources(object_store_memory=2 * GiB)` arg to `ray.air.DatasetConfig`. This effectively enables streaming by default, with a limit of 2GiB.
+
+2. Introduce `Dataset.get_preprocessor_lineage()`, which returns the lineage of preprocessors applied to this Dataset. If multiple preprocessors are applied, it will return a `Chain` of the preprocessors. Non-preprocessor operations on the Dataset are ignored. This function will be used by Train to persist fitted preprocessors with checkpoints.
+
+3. Deprecate the following `DatasetConfig` arguments when streaming is enabled: `max_object_store_memory_fraction`, `global_shuffle`, `fit`, `transform`, `randomize_block_order`, and `per_epoch_preprocessor`, as well as the following `Trainer` arguments: `preprocessor`.
```
stephanie-wang, line 38: Hmm I think we need some more discussion about removing `randomize_block_order` and the general plan for specifying shuffle behavior. Will the streaming backend not support randomizing block order at all?

There was this [earlier idea](https://docs.google.com/document/d/1hs1407QclTosonQn25NdknkBYHovU8Z1hZlZ8vrnCTk/edit#heading=h.8rhtes4whx6s) about replacing `global_shuffle`, `randomize_block_order` and `local_shuffle_buffer_size` with a single Trainer `shuffle` arg; perhaps we should include this in the proposal?

ericl: I like the idea of consolidating it all on a single `shuffle` arg. So in this new view, the only things the Trainer would be configuring for the data processing would be:
- split
- shuffle
- per-epoch preprocessing?

I think this seems pretty reasonable. Splitting is inherently tied to Trainer ingest setup, so it makes sense to manage. Shuffling is also somewhat orthogonal to data processing (a non-relational operation) and more about ingest.

stephanie-wang: Would be good to add a sentence or table here explaining how the previous functionality for these args should be specified after these changes.

stephanie-wang (approved):

reps/2023-03-15-train-api.md
```diff
@@ -0,0 +1,162 @@
...
+            ),
+        },
+    },
+)
+
+# Checkpoint includes fitted preprocessor.
+best_checkpoint = trainer.fit().checkpoint
+assert best_checkpoint.get_preprocessor() == prep
+```
+
```
amogkam, line 117: can you add an example for training -> inference?

In particular, how would the user specify certain preprocessors to not use for inference?

reps/2023-03-15-train-api.md
```diff
@@ -0,0 +1,162 @@
...
+@stephanie-wang
+
+## Design and Architecture
+
+## API Changes
+
+1. Introduce a `resource_limits: ExecutionResources(object_store_memory=2 * GiB)` arg to `ray.air.DatasetConfig`. This enables streaming by default, with a limit of 2GiB, and deprecates the previous `max_object_store_memory_fraction` argument.
+
+2. Introduce `Dataset.get_logical_plan()` (DeveloperAPI), which returns the logical plan that can be used to extract the lineage of preprocessors applied to this Dataset. If multiple preprocessors are applied, Train can return a `Chain` of the preprocessors. Non-preprocessor operations on the Dataset are ignored, and we can also allow ignoring preprocessors such as per epoch preprocessors. This function will be used by Train to persist fitted preprocessors with checkpoints.
```
amogkam, line 36: This should be a private API entirely right? Only used by Ray Train?

We don't expect library developers to be using this?

reps/2023-03-15-train-api.md
```diff
@@ -0,0 +1,162 @@
...
+
```
amogkam, line 37: What exactly is the API for users to ignore preprocessors from the lineage? Do we need to add a `ignore_for_inference: bool` API to the `Preprocessor.__init__` interface?

reps/2023-03-15-train-api.md
```diff
@@ -0,0 +1,162 @@
...
+### Shepherd of the Proposal (should be a senior committer)
+To make the review process more productive, the owner of each proposal should identify a **shepherd** (should be a senior Ray committer). The shepherd is responsible for working with the owner and making sure the proposal is in good shape (with necessary information) before marking it as ready for broader review.
+
+@stephanie-wang
+
+## Design and Architecture
+
+## API Changes
+
+1. Introduce a `resource_limits: ExecutionResources(object_store_memory=2 * GiB)` arg to `ray.air.DatasetConfig`. This enables streaming by default, with a limit of 2GiB, and deprecates the previous `max_object_store_memory_fraction` argument.
```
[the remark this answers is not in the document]
ericl: In general, you want ingest memory to usually fit on that of a single node. This is because that gives you the best locality, and you don't need to ramp up throughput beyond that of a single node anyways.

Using total object store memory could lead to *worse* performance on large clusters, as well as more unpredictable / contentious performance depending on the number of concurrent workloads. Setting this to a fixed, small value is safer and more predictable by default.

reps/2023-03-15-train-api.md
ericl: That's right, though it should still be DeveloperAPI for cross-library usages.

reps/2023-03-15-train-api.md
ericl: Added as `ignore_for_inference` attribue.

reps/2023-03-15-train-api.md
```diff
@@ -0,0 +1,163 @@
...
+
+base = ray.data.read_parquet("s3://bucket/etl_output")
+fact_table = ray.data.read_csv("s3://bucket/my.csv")
+
+# Create the preprocessor.
+prep = StandardScaler(["f1", "f2"])
+
+# Create the per-epoch preprocessor, and flag it for train only.
+per_epoch_prep = RandomNoisePreprocessor()
+per_epoch_prep.ignore_for_inference = True
```
amogkam, line 53: can we make this a part of the `Preprocessor.__init__` interface

reps/2023-03-15-train-api.md
amogkam: I think we should add this to the base Preprocessor interface.

ericl: Sure. It can be like the daemon flag for threading.Thread--- you can set it in either way.

richardliaw:
Quick question - what happens to people who want to make sure they don't experience train/serve skew?

i.e., how do we make it so that when people deploy their models, their serving requests receive the same data preprocessing? This was the original intent of the preprocessors.

ericl:
> It seems like there is no impact? but It would be good to confirm.

Yes, there's no impact since you are recommended to still use Preprocessors, just passing them to Data instead of to Train.

reps/2023-03-15-train-api.md
```diff
@@ -0,0 +1,163 @@
...
+base = ray.data.read_parquet("s3://bucket/etl_output")
+fact_table = ray.data.read_csv("s3://bucket/my.csv")
+
+# Fit the preprocessor.
+prep = StandardScaler(["f1", "f2"])
+prep.fit(base)
+
+# Apply base preprocessing.
+train_ds = base.map_batches(prep)
+train_ds.cache()  # Optional: cache the base data in memory.
```
richardliaw, line 90: So if you have a BatchMapper as part of your preprocessing... you would do:

```
train_ds = base.map_batches(BatchMapper(func))
?
```

What happens when you pass in custom functions? Do they also get captured as part of the "fitted preprocessor" at the end?

ericl: Currently they wouldn't be captured, though we could probably automatically capture them by wrapping them as a BatchMapper.

Do you think that would be preferable?

richardliaw: Hmm, seems like a bit of magic, but the current state is ambiguous. 

IMO it would be better to raise a warning for people passing in datasets into trainers that certain funcs are not captured, and force users to use the BatchMapper if they want to mute the warning?

ericl: Yes, we can definitely log warnings.

ericl:

amogkam:
Don't we want to run this by some users (along with the batch inference REP) before merging?

reps/2023-03-15-train-api.md
amogkam: can we add this as an API change in the REP?

amogkam: and can we update the example https://github.com/ray-project/enhancements/pull/26/files#diff-a22d41b93b796f1fc783b529bd58d9317c44da788c4a287603724374f20a7103R95 to show how to do this?

amogkam: ah I see you had it here https://github.com/ray-project/enhancements/pull/26/files#diff-a22d41b93b796f1fc783b529bd58d9317c44da788c4a287603724374f20a7103R52.

This should be in the After example and not the Before example

zhe-thoughts:
Merging since vote has passed

PR #27, opened 2023-03-21
[air] Introduce `RunConfig.storage_path` to replace `RunConfig.local_dir` and `SyncConfig.upload_dir`

krfricke:
[Link to full proposal](https://github.com/ray-project/enhancements/blob/air-storage-path/reps/2023-03-20-air-storage-path.md)

We'd like to change how users of Ray AIR (specifically: Train + Tune) configure permanent storage locations.

Current way:
- Users can specify `RunConfig.local_dir` or `tune.run(local_dir)` for storage on the local node
- Users can specify `SyncConfig.upload_dir` for storage on remote nodes

Future way:
- We have a unified setting `RunConfig.storage_path` that can be set either to a local dir or a remote path.
- Setting the storage path to a cloud or NFS URI (e.g., `s3://`, or `file://` that points to a NFS mount). In these cases, data will be first written to a local cache dir on the worker, and then synced to a subdirectory in the storage path designated by `<experiment_name>/<trial_name>/`.
- Setting the storage path to a purely local URI (e.g., `/home/foo/ray_results`). In this mode, trial data is synced to the head node via the object store. We would generally recommend using a remote storage path or shared directory instead.

### DIFF
--- a/reps/2023-03-20-air-storage-path.md
+++ b/reps/2023-03-20-air-storage-path.md
@@ -97,6 +97,7 @@ and therefore has to be within Ray.
 ## Design and Architecture
 
 - We introduce a new `storage_path` argument to `air.RunConfig` and `tune.run` as the main configuration entrypoints
+- The `storage_path` argument defaults to the configured `RAY_STORAGE` environment variable
 - If the `storage_path` is set to a remote URI, the `local_path` is read from an environment variable `RAY_AIR_CACHE_DIR`
 - Backwards compatibility: If a `local_dir` is passed, we set this environment variable
 - In downstream components (`TrialRunner`, `Experiment`, `Trial`), we introduce respective arguments: `storage_path` and `experiment_path`
--- a/reps/2023-03-20-air-storage-path.md
+++ b/reps/2023-03-20-air-storage-path.md
@@ -9,7 +9,7 @@ Current way:
 Future way:
 - We have a unified setting `RunConfig.storage_path` that can be set either to a local dir or a remote path.
 - Setting the storage path to a cloud or NFS URI (e.g., `s3://`, or `file://` that points to a NFS mount). In these cases, data will be first written to a local cache dir on the worker, and then synced to a subdirectory in the storage path designated by `<experiment_name>/<trial_name>/`.
-- Setting the storage path to a purely local URI (e.g., `/home/foo/ray_results`). In this mode, there is no persistence once nodes die, nor is there any syncing. The user will have to figure out which node their result is stored on in order to retrieve their result manually. We would only recommend this mode for single node cluster operation generally.
+- Setting the storage path to a purely local URI (e.g., `/home/foo/ray_results`). In this mode, trial data is synced to the head node via the object store. We would generally recommend using a remote storage path or shared directory instead.
 
 
 ```python
@@ -49,7 +49,7 @@ and therefore has to be within Ray.
 
 ### Required Reviewers
 
-@richardliaw, @ericl
+@richardliaw, @ericl, @gjoliver
 
 ### Shepherd of the Proposal (should be a senior committer)
 
@@ -86,6 +86,19 @@ from ray import air, tune
 
 results = tune.Tuner(train_fn).fit()
 assert results.best_checkpoint().path.startswith("s3://foo/bar")
+```
+
+
+### Usage with local directories
+
+```python
+import ray
+from ray import air, tune
+
+results = tune.Tuner(train_fn, run_config=air.RunConfig(storage_path="/local/dir")).fit()
+
+# even if train_fn ran on a different node, its logs + artifacts will be in /local/dir
+assert results.best_checkpoint().path.startswith("/local/dir")
 ```
 
 ### Accessing storage paths
@@ -126,15 +139,6 @@ For public APIs (`tune.Tuner()`, `tune.run()`):
 
 For developer APIs (`Trial`, `TrialRunner`, `Experiment`) we can speed up hard deprecation
 by one release.
-
-#### Syncer class
-
-Currently, when a `local_dir` but no `upload_dir` is specified, we synchronize trial data (e.g. checkpoints)
-via the object store. This REP states that this path will be deprecated eventually and data will
-not be synced instead.
-
-This is subject to discussion. For now, we can keep this functionality as it is functionally orthogonal to
-the rest of the proposal.
 
 ### Migration plan
 
--- a/reps/2023-03-20-air-storage-path.md
+++ b/reps/2023-03-20-air-storage-path.md
@@ -152,8 +152,9 @@ Thus, when a local storage path is defined (cloud storage is disabled):
 - The trainable saves trial-based data such as checkpoints to their local `/local/path/experiment_name/trial_id`
 - **The trainable uses the Ray object store to transfer these checkpoints _synchronously_ to the head node.**
 
-The trainable can detect if the trial is running on the driver node, which will then 
-remove the need for any syncing to happen. This will also cover the case when
+The trainable can detect if the trial shares the storage path with the head node
+(e.g. if RAY_STORAGE is set, or if it's running on the same node). This then 
+removes the need for any syncing to happen. This will also cover the case when
 shared storage (e.g. NFS) is defined.
 
 ## Usage Example
--- a/reps/2023-03-20-air-storage-path.md
+++ b/reps/2023-03-20-air-storage-path.md
@@ -11,6 +11,15 @@ Future way:
 - Setting the storage path to a cloud or NFS URI (e.g., `s3://`, or `file://` that points to a NFS mount). In these cases, data will be first written to a local cache dir on the worker, and then synced to a subdirectory in the storage path designated by `<experiment_name>/<trial_name>/`.
 - Setting the storage path to a purely local URI (e.g., `/home/foo/ray_results`). In this mode, there is no persistence once nodes die, nor is there any syncing. The user will have to figure out which node their result is stored on in order to retrieve their result manually. We would only recommend this mode for single node cluster operation generally.
 
+
+```python
+import ray
+from ray import air, tune
+
+results = tune.Tuner(train_fn, run_config=air.RunConfig(storage_path="s3://foo/bar")).fit()
+assert results.best_checkpoint().path.startswith("s3://foo/bar")
+```
+
 ### General Motivation
 
 Users care about where their data is ultimately stored. They usually don't care about
--- a/reps/2023-03-20-air-storage-path.md
+++ b/reps/2023-03-20-air-storage-path.md
@@ -133,6 +133,9 @@ When cloud storage is disabled:
   This uses the Ray object store.
 - This also happens _synchronously_ every time a trial reports a checkpoint has been saved
 
+Currently, the object store-based transfer serializes all the data at the same time. Thus, the heap
+memory must be large enough to hold all the data to be transferred between nodes.
+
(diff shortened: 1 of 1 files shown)


reps/2023-03-20-air-storage-path.md
```diff
@@ -0,0 +1,151 @@
...
+
+results = tune.Tuner(train_fn).fit()
+
+assert results.path.startswith("s3://bucket/dir/experiment_dir")
+
+result = results.get_best_result()
+assert result.path.startswith("s3://bucket/dir/experiment_dir/trial0")
+
+checkpoint = result.checkpoint
+assert checkpoint.path.startswith("s3://bucket/dir/experiment_dir/trial0/checkpoint_0010")
```
ericl, line 105: Can you also add examples of what happens when a local storage path is configured? Would things still magically end up on the head node, or would that not be supported?

reps/2023-03-20-air-storage-path.md
```diff
@@ -0,0 +1,151 @@
...
+- We keep this behavior in Ray 2.5
+- In Ray 2.6 we will raise an error if those arguments are still used
+- In Ray 2.7 we remove the old code paths
+
+For developer APIs (`Trial`, `TrialRunner`, `Experiment`) we can speed up hard deprecation
+by one release.
+
+#### Syncer class
+
+Currently, when a `local_dir` but no `upload_dir` is specified, we synchronize trial data (e.g. checkpoints)
```
ericl, line 132: I think the end behavior here needs to be part of the REP.

reps/2023-03-20-air-storage-path.md
```diff
@@ -0,0 +1,151 @@
...
+We'd like to change how users of Ray AIR (specifically: Train + Tune) configure permanent storage locations.
+
+Current way:
+- Users can specify `RunConfig.local_dir` or `tune.run(local_dir)` for storage on the local node
+- Users can specify `SyncConfig.upload_dir` for storage on remote nodes
+
+Future way:
+- We have a unified setting `RunConfig.storage_path` that can be set either to a local dir or a remote path.
+- Setting the storage path to a cloud or NFS URI (e.g., `s3://`, or `file://` that points to a NFS mount). In these cases, data will be first written to a local cache dir on the worker, and then synced to a subdirectory in the storage path designated by `<experiment_name>/<trial_name>/`.
+- Setting the storage path to a purely local URI (e.g., `/home/foo/ray_results`). In this mode, there is no persistence once nodes die, nor is there any syncing. The user will have to figure out which node their result is stored on in order to retrieve their result manually. We would only recommend this mode for single node cluster operation generally.
```
richardliaw, line 12: Should we still support the "pure local URI" to work with multiple nodes? Same question as Eric below, not sure if this is going to be higher impact than we would like...

richardliaw: Actually I am reading through the PRD and I'm not seeing that we are actually intending to disable syncing to the head node. Are you seeing anything that says this differently?

reps/2023-03-20-air-storage-path.md
```diff
@@ -0,0 +1,151 @@
...
+
+This is an API change to Ray AIR (specifically, `air.RunConfig`, `tune.SyncConfig`, and `tune.run()`)
+and therefore has to be within Ray.
+
+
+## Stewardship
+
+### Required Reviewers
+
+@richardliaw, @ericl
```
ericl, line 52: @gjoliver can you also review?

reps/2023-03-20-air-storage-path.md
krfricke: This was based on an offline discussion. But I generally agree, the syncing/not syncing part is orthogonal. I'll remove this from the PRD and we keep the current node-to-node syncing behavior.

reps/2023-03-20-air-storage-path.md
krfricke: removed this section with respect to the discussion in the REP

ericl (changes requested):
The API looks reasonable at a high level. I think the main thing is fleshing out the implementation details a little more, which is necessary to evaluate the merits of the proposal.

Can you include a sketch of the basic algorithm for how remote storage upload works, as well as the algorithm / implementation for how local path sync will work?

It would also be good to introduce the basic Tune concepts of storage (e.g., what does Tune put in storage, what can user put in storage, what functionalities is storage used for).

reps/2023-03-20-air-storage-path.md
```diff
@@ -0,0 +1,155 @@
...
+
+We'd like to change how users of Ray AIR (specifically: Train + Tune) configure permanent storage locations.
+
+Current way:
+- Users can specify `RunConfig.local_dir` or `tune.run(local_dir)` for storage on the local node
+- Users can specify `SyncConfig.upload_dir` for storage on remote nodes
+
+Future way:
+- We have a unified setting `RunConfig.storage_path` that can be set either to a local dir or a remote path.
+- Setting the storage path to a cloud or NFS URI (e.g., `s3://`, or `file://` that points to a NFS mount). In these cases, data will be first written to a local cache dir on the worker, and then synced to a subdirectory in the storage path designated by `<experiment_name>/<trial_name>/`.
```
gjoliver, line 11: not sure if users know how to differentiate and tell local disk from nfs.
we can't detect this ourselves?

reps/2023-03-20-air-storage-path.md
```diff
@@ -0,0 +1,155 @@
...
+### General Motivation
+
+Users care about where their data is ultimately stored. They usually don't care about
+how it got there.
+
+However, currently they specify a local storage directory, from which data is periodically "synced"
+to a remote storage location. This flow (collect locally, sync to storage) is more of an implementation
+detail.
+
+Instead, users should just specify where their data ends up - with `storage_path`.
```
gjoliver, line 32: this sounds good. a quick question is what happens to the metadata and metrics that end up on the driver node today?
if users only specify an s3 path, where would they go under the new scheme.
also I assume the driver process can also upload things to specified s3 paths right? because there will be Trial level artifacts that folks need to save.

gjoliver (commented):
I think it's definitely the right direction.
a couple of quetions.

reps/2023-03-20-air-storage-path.md
krfricke: Currently the driver triggers a pull-based sync from the worker nodes to the head node.

Instead, we can move this to the trainable on the worker nodes in the same way that we do for cloud-based syncing today.

Then the driver can drop a marker file in the trial directory. The trainable can see if this marker exists - if so, no syncing is necessary as it's a shared directory (either the trial is running on the head node, or it's NFS). Syncing only has to be done if the file is not found.

I can add this to the REP.

ericl: Yeah, actually RAY_STORAGE guarantees the configured storage is available across the cluster already with this trick. So if you are able to obtain a filesystem from Ray storage, you can assume it's a NFS / S3 backed, or on a single node.

reps/2023-03-20-air-storage-path.md
```diff
@@ -0,0 +1,155 @@
...
+
+@richardliaw, @ericl, @gjoliver
+
+### Shepherd of the Proposal (should be a senior committer)
+
+@richardliaw
+
+## Design and Architecture
+
+- We introduce a new `storage_path` argument to `air.RunConfig` and `tune.run` as the main configuration entrypoints
```
ericl, line 60: This should default to RAY_STORAGE right?

reps/2023-03-20-air-storage-path.md
```diff
@@ -0,0 +1,155 @@
+## Summary
+
+We'd like to change how users of Ray AIR (specifically: Train + Tune) configure permanent storage locations.
+
+Current way:
+- Users can specify `RunConfig.local_dir` or `tune.run(local_dir)` for storage on the local node
+- Users can specify `SyncConfig.upload_dir` for storage on remote nodes
+
+Future way:
+- We have a unified setting `RunConfig.storage_path` that can be set either to a local dir or a remote path.
```
ericl, line 10:
```suggestion
- All remote storage is standardized on the built in RAY_STORAGE mechanism.
- We have a unified setting `RunConfig.storage_path` that can be used to override the default RAY_STORAGE path configured for the cluster.
```

krfricke:
I've added a background section about storage in Ray Tune as well as a description of the current implementation and possible future changes to it.

Let me know if this is too verbose

ericl (approved):
Thanks for adding the historical background, it is very useful. This looks good to me at a high level, though I have a few more comments on clarification of limitations.

reps/2023-03-20-air-storage-path.md
```diff
@@ -0,0 +1,248 @@
...
+- The trainable uses a `Syncer` to upload these checkpoints _synchronously_ to the remote storage after saving, e.g. `s3://bucket/location/experiment_name/trial_id/checkpoint_001`
+
+When cloud storage is disabled:
+
+- The driver saves experiment data to `/local/path/experiment_name`
+- The driver saves driver-based trial data to `/local/path/experiment_name/trial_id`
+- The trainable saves trial-based data such as checkpoints to their local `/local/path/experiment_name/trial_id`
+- The driver uses a `SyncerCallback` to synchronize contents form the trial directories (i.e. the missing trial-based data).
+  This uses the Ray object store.
+- This also happens _synchronously_ every time a trial reports a checkpoint has been saved
```
ericl, line 134: What are the limitations of this mechanism? Is there some max size limit?

reps/2023-03-20-air-storage-path.md
```diff
@@ -0,0 +1,248 @@
...
+the driver synchronously blocks until trial-based data is synchronized using the `SyncerCallback`.
+
+Instead, we should move this synchronization into the trial - analogous to the remote storage case.
+
+Thus, when a local storage path is defined (cloud storage is disabled):
+
+- The driver saves experiment data to `/local/path/experiment_name`
+- The driver saves driver-based trial data to `/local/path/experiment_name/trial_id`
+- The trainable saves trial-based data such as checkpoints to their local `/local/path/experiment_name/trial_id`
+- **The trainable uses the Ray object store to transfer these checkpoints _synchronously_ to the head node.**
```
ericl, line 153: Who receives the data here, is there a separate actor that receives the transfer?

reps/2023-03-20-air-storage-path.md
krfricke: Good point. The data must fit into the heap memory. I've added a comment.

reps/2023-03-20-air-storage-path.md
krfricke: Yes! Added a comment

krfricke:
Thank you very much for all your input.

cc @richardliaw @gjoliver would appreciate final thoughts from you. We can then send it out to the ray committers list.

reps/2023-03-20-air-storage-path.md
gjoliver: nice

richardliaw (approved):
great!

reps/2023-03-20-air-storage-path.md
gjoliver: ok, cool!

gjoliver (approved):

ericl:
@zhe-thoughts could we merge this?

krfricke:
Thanks! WIP Pr for the implementation is here: https://github.com/ray-project/ray/pull/33463

PR #29, opened 2023-04-28
[REP] Roll out "strict mode" for Ray Data

ericl:
This is a lighweight proposal for rolling out strict mode by default.

### DIFF
--- a/reps/2023-04-27-data-strict-mode.md
+++ b/reps/2023-04-27-data-strict-mode.md
@@ -1,0 +1,98 @@
+# Rollout out "strict mode" for Ray Data
+
+## Summary
+
+Make a breaking API change to always require data schemas in Ray Data.
+
+### General Motivation
+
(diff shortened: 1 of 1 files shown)


pcmoritz (approved):
Looks great to me, thanks a lot for driving this!

We might want to add a note that in addition to the unification and simplicity benefits, this aligns the Ray Data API closer to industry-standard distributed data APIs like Spark and also emerging standards for machine learning datasets like HuggingFace.

reps/2023-04-27-data-strict-mode.md
```diff
@@ -0,0 +1,98 @@
+# Roll out "strict mode" for Ray Data
+
+## Summary
+
+Make a breaking API change to always require data schemas in Ray Data.
+
+### General Motivation
+
+This REP proposes rolling out a breaking API change to Ray Data, termed "strict mode". In strict mode, support for standalone Python objects is dropped. This means that instead of directly storing, e.g., Python `Tuple[str, int]` instance in Ray Data, users will have to either give each field a name (i.e., `{foo: str, bar: int}`), or use a named object-type field (i.e., `{foo: object}`). In addition, strict mode removes the "default" batch format in place of "numpy" by default. This means that most users just need to be aware of `Dict[str, Any]` (non-batched data records) and `Dict[str, np.ndarray]` (batched data) types when working with Ray Data.
```
jjyao, line 9: Is the reason why we choose numpy as the default format that it has zero conversion cost from whatever the underling block format is (arrow or pandas)?

ericl: That's one of the reasons, but the main one actually that numpy format is as close to Pythonic as you can get (dict of arrays). With this, the API is more accessible to users not familiar with pandas.

ericl:
@zhe-thoughts ok to merge

PR #31, opened 2023-05-13
[Core] Add a new interface for submitting actor tasks in batches (Batch Remote)

larrylian:
**Core Motivation**:
1. Improve the performance of batch calling ActorTask.
2. Implement Ray's native collective communication library through this interface.  -- This will be discussed in a new REP afterwards.

Current situation of batch calling actor tasks:
```
actors = [WorkerActor.remote() for _ in range(400)]

# This loop's repeated invocation actually wastes a lot of performance.
for actor in actors:
  actor.compute.remote(args)
```

Using the new Batch Remote API:
```
actors = [WorkerActor.remote() for _ in range(400)]

# Calling it only once can greatly improve performance.  
Plan 1
batch_remote_handle = ray.experimental.batch_remote(actors)
batch_remote_handle.compute.remote(args)

Plan 2
batch_remote_handle = ray.experimental.BatchRemoteHandle(actors)
batch_remote_handle.compute.remote(args)
```

The Batch Remote API can save the following performance costs(The N is the number of Actors):  
1. Reduce (N-1) times of parameter serialization performance time.
2. Reduce (N-1) times of putting parameter into object store performance time for scenarios with large parameters.
3. Reduce (N-1) times of python and C++ execution layer switching and repeated parameter verification performance time.

[WIP][Core]Add batch remote api for batch submit actor task
https://github.com/ray-project/ray/pull/35597

### DIFF
--- a/reps/2023-05-10-actors-batch-remote-api.md
+++ b/reps/2023-05-10-actors-batch-remote-api.md
@@ -1,0 +1,205 @@
+## Summary
+### General Motivation
+**Core Motivation**:
+1. Improve the performance of batch calling ActorTask. The current performance bottleneck of batch invoking ActorTask mainly lies in the duplicate serialization of parameters, repeated puts parameters into object store, and frequent context switching between Python and C++. The Batch remote API will optimize performance for these three aspects.
+
+
+In distributed computing scenarios, such as big data computing、AI training and inference, there is a need to send a large number of RPC requests(ActorTask) to multiple Actors in batches.  
+For example, in a typical star-topology architecture with one master Actor and 400 worker Actors, the same computational request needs to be sent to 400 worker Actors. 
+However, the computing tasks of each Worker Actor are very short. In such a scenario, the performance requirements for executing batch actor task remote of a large number of Actor are very high.  
+Therefore, for the scenario of batch calling Actor tasks, I want to add a new optimization API, batch_remote(), to improve the performance of batch submission of Actor Task calls.   
+After my own performance testing and comparison, this API can achieve a performance improvement of 40% ~ 98%.
+
+Current situation of batch calling actor tasks:
+```
+actors = [WorkerActor.remote() for _ in range(400)]
+
+# This loop's repeated invocation actually wastes a lot of performance.
+for actor in actors:
+  actor.compute.remote(args)
+```
+
+Using the new Batch Remote API:
+```
+actors = [WorkerActor.remote() for _ in range(400)]
+
+# Calling it only once can greatly improve performance.
+ray.experimental.batch_remote(actors).compute.remote(args)
+```
+
+The current performance bottleneck of batch invoking ActorTask mainly lies in the duplicate serialization of parameters, repeated puts parameters into object store, and frequent context switching between Python and C++.  
+The Batch Remote API can reduce the following performance costs(The N is the number of Actors):  
+1. Reduce (N-1) times of parameter serialization performance costs.
+2. Reduce (N-1) times of putting parameter into object store performance costs for scenarios with large parameters.
+3. Reduce (N-1) times of context switching between Python and C++ and repeated parameter verification performance costs.
+
+
+### Should this change be within `ray` or outside?
+This requires adding a new interface in Ray Core.  
+The initial consideration is to add it to the ray.experimental module.
+
+## Stewardship
+### Required Reviewers
+...
+
+### Shepherd of the Proposal (should be a senior committer)
+...
+
+## Design and Architecture
+### API
+Use case
+```
+ray.experimental.batch_remote(actors).compute.remote(args)
+```
+
+### Implementation
+
+1. In the ActorTask calling logic, the parameter verification, parameter serialization, and parameter putting into the object store can be reused from the original logic.
+2. Then, add a BatchSubmitActorTask interface in the C++ CoreWorker layer, and loop through the original SubmitActorTask interface to submit multiple ActorTasks. This way, the batch submission of ActorTasks can be handled in the C++ CoreWorker layer, saving time on parameter serialization and putting into the object store. 
+
+Overall, the implementation is relatively simple, with most of it reusing the original ActorTask submission logic.
+
+```
+def batch_remote(actors: List[ray.actor.ActorHandle]) -> BatchRemoteHandle:
+  ...
+
+class BatchRemoteHandle:
+  def __init__(self, actors: List[ray.actor.ActorHandle]):
+    ...
+  
+  def _actor_method_call():
+    object_refs = worker.core_worker.batch_submit_actor_task(actor_ids, ...)
+    return object_refs
+
+C++: 
+class CoreWorker:
+  BatchSubmitActorTask(actor_ids, &return_ids) {
+    for( auto actor_id : actor_ids) {
+      SubmitActorTask(actor_id, ..., return_ids)
+    }
+  }
+```
+
+### Performance improvement comparison:
+We have already implemented and conducted extensive performance testing internally.  
+The following are the performance comparison results.
+
+**Table 1: Comparison of remote call time with varying parameter sizes and 400 Actors**
+
+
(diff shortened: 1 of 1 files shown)


valiantljk:
Curious how much performance gains do we expect? Any perf test?

reps/2023-05-10-actors-batch-remote-api.md
```diff
@@ -0,0 +1,204 @@
+## Summary
+### General Motivation
+**Core Motivation**:
+1. Improve the performance of batch calling ActorTask.
+2. Implement Ray's native collective communication library through this interface.
```
jovany-wang, line 5: We should split this REP into 2: 
1. batch remote in ray core REP
2. introducing RAY_NATIVE mode in ray collective lib REP

reps/2023-05-10-actors-batch-remote-api.md
```diff
@@ -0,0 +1,204 @@
+## Summary
+### General Motivation
+**Core Motivation**:
+1. Improve the performance of batch calling ActorTask.
```
jovany-wang, line 4: We should introduce the bottleneck aspects, such as frequent context switching between python and cpp, serializing the same object in multiple times.

reps/2023-05-10-actors-batch-remote-api.md
```diff
@@ -0,0 +1,204 @@
...
+```
+
+The Batch Remote API can save the following performance costs(The N is the number of Actors):  
+1. Reduce (N-1) times of parameter serialization performance time.
+2. Reduce (N-1) times of putting parameter into object store performance time for scenarios with large parameters.
+3. Reduce (N-1) times of python and C++ execution layer switching and repeated parameter verification performance time.
+
+
+### Should this change be within `ray` or outside?
+This requires adding a new interface in Ray Core.
```
jovany-wang, line 38: I believe this should be added as an experimental & internal API first:
```python
ray.experimental._batch_remote(actors).compute.remote(args)
```

jovany-wang (commented):
Please add the ad/dis sections.

jovany-wang:
I guess this REP also benefits RLlib's sampling and weights-syncing aspects. 

@gjoliver CC

reps/2023-05-10-actors-batch-remote-api.md
larrylian: Yes, This REP just adds Batch Rmote API.

reps/2023-05-10-actors-batch-remote-api.md
larrylian: Yes. At first, it should be placed in the experimental module.

reps/2023-05-10-actors-batch-remote-api.md
```diff
@@ -0,0 +1,205 @@
...
+
+Using the new Batch Remote API:
+```
+actors = [WorkerActor.remote() for _ in range(400)]
+
+# Calling it only once can greatly improve performance.
+ray.experimental.batch_remote(actors).compute.remote(args)
+```
+
+The current performance bottleneck of batch invoking ActorTask mainly lies in the duplicate serialization of parameters, repeated puts parameters into object store, and frequent context switching between Python and C++.
```
jjyao, line 30: Currently one way to avoid putting the same args repeatedly to object store is calling 'ray.put(args)' first and pass the obj ref to '.remote()'. Will this be good enough?

jjyao (commented):
A high level comment: can we do the optimization transparently without introducing a new api. For example, we cache the last args we see, if the new arg is the same, we reuse the previously serialized obj ref.

reps/2023-05-10-actors-batch-remote-api.md
larrylian: The method you mentioned can only reduce the performance cost of repeatedly serializing parameters. However, the performance cost of frequent switching context between python and c++ cannot be avoided.   
You can see my performance test even without parameters. Batch remote API performance optimization can reach 40%.

image

larrylian:
> A high level comment: can we do the optimization transparently without introducing a new api. For example, we cache the last args we see, if the new arg is the same, we reuse the previously serialized obj ref.

@jjyao 
The method you mentioned can only reduce the performance cost of repeatedly serializing parameters. However, the performance cost of frequent switching context between python and c++ cannot be avoided.   
You can see my performance test even without parameters. Batch remote API performance optimization can reach 40%.

image

reps/2023-05-10-actors-batch-remote-api.md
jovany-wang: @jjyao Yes, so I'd like to introduce it as an internal API.

jovany-wang:
Following up on the last offline meeting, please:
1) add a POC implementation to show the code structure.
2) update the failure section.

larrylian:
@scv119 @jjyao 
1. I have added a POC implementation PR to show the code strucature - https://github.com/ray-project/ray/pull/35597 
2. I have added a performance comparison chart with an objectRef as a parameter. In the case of using an objectRef, the batch remote approach shows a 3-4 times improvement in performance.
3. I have added section of "Failure & Exception Scenario"

cc @jovany-wang

reps/2023-05-10-actors-batch-remote-api.md
```diff
@@ -0,0 +1,264 @@
...
+### Shepherd of the Proposal (should be a senior committer)
+...
+
+## Design and Architecture
+### API
+Use case  
+Plan 1
+```
+batch_remote_handle = ray.experimental.batch_remote(actors)
+batch_remote_handle.compute.remote(args)
```
scv119, line 55: what happens if one of the actor failed? (i.e. killed or terminated?)

rkooo567:
> Reduce (N-1) times of python and C++ execution layer switching and repeated parameter verification performance time.

Is there any flamegraph that backs up this? IIUC, the verification is done in Cython, and there's no such thing as "switching" (cython is just C).

larrylian:
> Is there any flamegraph that backs up this? IIUC, the verification is done in Cython, and there's no such thing as "switching" (cython is just C).
@rkooo567  
This "switching" refers to the execution context switch between python and (cython & c++). 
I have confirmed through verification that after using BatchRemote optimization, the performance can be improved by 2~3 times in the scenario without parameters.
image

reps/2023-05-10-actors-batch-remote-api.md
larrylian: Failure & Exception Scenario.
**1. Exceptions occurred during parameter validation or preprocessing before batch submission of ActorTasks.** 
Since these exceptions occur before the process of submitting ActorTasks, they can be handled by directly throwing specific error exceptions as current situation.

**2. Some actors throw exceptions during the process of batch submitting ActorTasks.** 
When traversing and submitting ActorTasks in a loop, if one of the Actors throws an exception during submission, the subsequent ActorTasks will be terminated immediately, and the exception will be throwed to user.

Reason:
Submitting ActorTask is normally done without any exceptions being thrown. If an error does occur, it is likely due to issues with the code and will require modifications.
The exception behavior of this plan is the same as the current foreach remote.

rkooo567:
(it is not a blocker); That doesn't prove it is context switching cost though. I feel like it is sth else. Cython is just C code, so there should be no such things as Python <-> cpp context switching IIUC. It is different from Java <-> CPP

jovany-wang:
> (it is not a blocker); That doesn't prove it is context switching cost though. I feel like it is sth else. Cython is just C code, so there should be no such things as Python <-> cpp context switching IIUC. It is different from Java <-> CPP?

The most frequent context switching is `with gil()` and `without gil()`.

TRSWNCA:
> (it is not a blocker); That doesn't prove it is context switching cost though. I feel like it is sth else. Cython is just C code, so there should be no such things as Python <-> cpp context switching IIUC. It is different from Java <-> CPP?

I suppose the context switching happens when for loops continue?
Original way use for loop in python, while if we provide a interface, the for loop would be in Ray Core in C.

edoakes:

PR #35, opened 2023-07-07
[REP] Consolidated persistence API for Ray Train/Tune

ericl:
This REP proposes standardizing on the recently introduced `storage_path` option for Ray Train/Tune, deprecating other legacy persistence paths.


pcmoritz:
Thanks for putting this together and adding the implementation recommendations!

In the motivation you should consider adding the prevalence of larger and larger models in generative AI and LLMs.

ericl:
> In the motivation you should consider adding the prevalence of larger and larger models in generative AI and LLMs.

Good point, indeed that is one of the reasons why persistence has become increasingly a pain point over the past year.

reps/2023-06-06-simplify-sync.md
```diff
@@ -0,0 +1,109 @@
...
+For example, this is how you could record a Torch checkpoint:
+
+```python
+   torch.save(model, "path/to/dir")
+   train.report(checkpoint=Checkpoint.from_directory("path/to/dir"))
+```
+
+Train will make a copy of the specified checkpoint data in the trial dir for persistence. The checkpoint data is managed by Train (e.g., Train may restore only the latest checkpoint or delete previous checkpoints to save space according to checkpoint policy).
+
+The files of the recorded checkpoint can be accessed via `result.checkpoint`. The `Checkpoint` object itself is a logical tuple of `(path, pyarrow.fs.FileSystem)`. Users can also get and set arbitrary metadata to these checkpoints (e.g., preprocessor configs, model settings, etc), which will be recorded in a `metadata.json` file.
```
krfricke, line 62: How is the filesystem populated? The user e.g. creates a checkpoint using `Checkpoint.from_directory` - the filesystem is a local filesystem. Ray uploads the checkpoint to the storage path - it becomes a new checkpoint with a S3Filesystem. Can this be customized, e.g. when users need a customized pyarrow-compatible s3 syncer or uses a custom URI?

What happens when the user doesn't set a `storage_path` (so uses a local path) and saves checkpoints? Do we raise an error? As the local filesystem would be invalid on the head node in a distributed setup.

reps/2023-06-06-simplify-sync.md
```diff
@@ -0,0 +1,109 @@
...
+
+Checkpoints can be recorded from multiple ranks. By default, only checkpoint data from rank zero is preserved. Data from all ranks can be retained via a `keep_all_ranks` option.
```
krfricke, line 64: How do we store/organize these different checkpoints? E.g. if rank 0 and rank 1 both send a Checkpoint object to train.report, do we store and restore them separately by rank?

There are some applications that require the checkpoints from different local ranks to be stored in the same folder (e.g. pytorch lightning). I'm not fully sure about the details there cc @woshiyyya to shed some light on this

reps/2023-06-06-simplify-sync.md
```diff
@@ -0,0 +1,109 @@
...
+
+- What if I do not have NFS or cloud storage?
+
+	Train no longer supports persistence for distributed experiments unless you specify a `storage_path`. You can provide a custom Arrow filesystem to support custom storage solutions.
+
+## Open Questions
+
+- For multi-rank checkpointing, should we consolidate outputs from multiple ranks into a single directory, retain the outputs in separate directories (e.g., `rank_{i}`), or provide options to support both?
+
+- For backwards compatibility, should we provide a (slower, limited) head-node remote filesystem implementation?
```
krfricke, line 102: I think we should keep that at first with a warning and then phase out to collect feedback and see how users are managing without it

reps/2023-06-06-simplify-sync.md
```diff
@@ -0,0 +1,109 @@
...
+### Impact of Changes
+
+This is a breaking change for users that do not already use NFS or cloud storage. For other users, the changes are for the most part limited to beta APIs (i.e., Checkpoint).
+
+We are evaluating the impact of deprecating head node storage, by rolling the breaking change out early in Ray 2.6. This is to inform whether we should provide extended backwards compatibility.
+
+### Timeline
+
+- Ray 2.6: Head node sync is disabled and will raise an exception. It is possible to re-enable by setting a flag.
+- Ray 2.7: Head node sync will be fully removed.
```
krfricke, line 115: I think we should push this back to 2.8 for the breaking API change

krfricke (commented):
Generally I'm very much in favor of simplifying this.

Note that removing the support for automatic syncing will change the way users interact with the code. Until now, code "just worked" in distributed setups without specifying a storage path. 

Now, fault-tolerant distributed training will not work out of the box on distributed setups without additional configuration from the user side.

This means that the code that works locally may have to be changed to work on the cloud. We remove complexity from the library setup and push the responsibility to the cluster setup and configuration, i.e. to the user.

In many cases, this is preferred - synchronization of large model checkpoints was very inefficient, and was thus a footgun that could lead to problems. By making the storage configuration explicit we can remove inefficient behavior (syncing large checkpoints via the object store).

It's just something to be aware of - our examples will all have to include a storage path definition, or, if users rely on the RAY_STORAGE environment variable, they have to be aware of this in the docs to avoid a frustrating onboarding experience.

reps/2023-06-06-simplify-sync.md
```diff
@@ -0,0 +1,109 @@
...
+While trials can write arbitrary data to the persistent trial directory, they cannot be guaranteed they can read it back on checkpoint/restore. Data that is needed for resume must be recorded via the `train.report(checkpoint)` API.
+
+For example, this is how you could record a Torch checkpoint:
+
+```python
+   torch.save(model, "path/to/dir")
+   train.report(checkpoint=Checkpoint.from_directory("path/to/dir"))
+```
+
+Train will make a copy of the specified checkpoint data in the trial dir for persistence. The checkpoint data is managed by Train (e.g., Train may restore only the latest checkpoint or delete previous checkpoints to save space according to checkpoint policy).
```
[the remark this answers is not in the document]
ericl: Good point. I think we can retain this existing API contract (perhaps enforcing it with a `mv`).

How about one of these alternatives:
1. Train will take ownership of the specified checkpoint data, deleting the directory after the checkpoint finishes.
or
2. Train will upload the specified checkpoint data to its final storage location, and the user is responsible for deleting it after the report call.

?

reps/2023-06-06-simplify-sync.md
ericl: The filesystem will always reflect a valid location of the checkpoint data. The Checkpoint class becomes immutable, so if it is uploaded, then a new checkpoint will be created pointing to the remote data. We won't have any to_uri() method, so users shouldn't observe this happening in practice in any case.

Users cannot customize the sync, except by providing an alternative storage_filesystem implementation. We will always use pyarrow.fs.copy_files() to transfer the data.

> What happens when the user doesn't set a storage_path (so uses a local path) and saves checkpoints? Do we raise an error? As the local filesystem would be invalid on the head node in a distributed setup.

Yes, we will raise an error in this case (per FAQ: "If storage path is not set, sync is disabled, and an error is raised if a remote trial tries to persist a checkpoint.").

reps/2023-06-06-simplify-sync.md
ericl: We have recently discussed this in more detail and there is a consensus that we should generally merge the checkpoint dirs into a single folder.

The reason is this: users can always namespace their files/dirs by rank by the `train.context.world_rank` manually, so there isn't any reason why Train should do it for them. In fact, certain frameworks seems to assume merged directories and could break if we enforced per-rank subdirectories.

Similarly, for restoration the new Checkpoint API allows users to customize restoration strategy. For example, if they wanted to implement per-rank restore, they could call `pyarrow.fs.copy_files()` on selected files within the checkpoint path, rather than using the naive `Checkpoint.to_directory()` call.

woshiyyya: Actually not only for lightning, as long as you are using model-parallel strategies with `deepspeed` or `fsdp`, each worker saves its own model shard in the same folder, so that the full checkpoint can be consolidated with some utility function provided by those libraries. 

Therefore, I suggest to keep them in the same folder, and not separate them by rank. We can differentiate these shard file with a suffix/prefix.

reps/2023-06-06-simplify-sync.md
ericl: Sounds good. By the warning above, do you mean to keep the 2.6 soft-deprecation behavior for longer to collect feedback?

reps/2023-06-06-simplify-sync.md
```diff
@@ -0,0 +1,123 @@
...
+
+The files of the recorded checkpoint can be accessed via `result.checkpoint`. The `Checkpoint` object itself is a logical tuple of `(path, pyarrow.fs.FileSystem)`. Users can also get and set arbitrary metadata to these checkpoints (e.g., preprocessor configs, model settings, etc), which will be recorded in a `metadata.json` file.
+
+Checkpoints can be recorded from multiple ranks. By default, only checkpoint data from rank zero is preserved. Data from all ranks can be retained via a `keep_all_ranks` option.
+
+### FAQ
+
+- What's in this persistent trial directory?
+
+    This directory contains metric files written by Train loggers, such as JSON metrics, checkpoint data managed by Train, as well as arbitrary artifacts created by a trial.
```
woshiyyya, line 70: We may need a artifact whitelist for users to specify which artifacts they want to sync. 

If the users save checkpoints in the trial folder, then they will have two copies(one in trial dir, one in the AIR checkpoint dir), which introduced more syncing overhead.

ericl: Good point, we can add this to the REP. In the prototype, we are already using an internal (already exists actually) file filtering filesystem class, which filters out checkpoint data by defualt.

reps/2023-06-06-simplify-sync.md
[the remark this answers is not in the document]
woshiyyya: Another question here: why do we need to make a copy of the checkpoint? If checkpoints are synced to cloud storage, aren't they already persisted?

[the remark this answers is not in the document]
ericl: Yup I believe we do not need to make a copy.

reps/2023-06-06-simplify-sync.md
woshiyyya: Agree. From the AIR telemetry data, >85% of the users are using the default head node syncing now.

reps/2023-06-06-simplify-sync.md
krfricke: SGTM, thanks for clairfying

reps/2023-06-06-simplify-sync.md
krfricke: Yes - we keep the soft deprecation for longer and remove it (assuming no contrary feedback) fully in 2.8

reps/2023-06-06-simplify-sync.md
krfricke: I see, the missing link for me was the `storage_filesystem` setting. This should be sufficient to customize syncing behavior.

It would be good to get a better understanding of which APIs of pyarrow.fs.Filesystem have to be implemented by users to work with Ray AIRs operations, but that's a question for the documentation.

reps/2023-06-06-simplify-sync.md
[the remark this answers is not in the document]
krfricke: 1) is probably the easiest way to go forward with respect to the current implementation and mirrors most of the current functionality. In the future we may want to rethink the checkpoint creation API (there's quite a bit of discussion around this, but the general consensus is that users can usually easily adapt existing examples. Thus anything reasonable should work).

ericl:
Thanks for the comments @krfricke @woshiyyya , they should be incorporated now.

reps/2023-06-06-simplify-sync.md
```diff
@@ -0,0 +1,127 @@
...
+
+The files of the recorded checkpoint can be accessed via `result.checkpoint`. The `Checkpoint` object itself is a logical tuple of `(path, pyarrow.fs.FileSystem)`, and this tuple is immutable. Users can also get and set arbitrary metadata to these checkpoints (e.g., preprocessor configs, model settings, etc), which will be recorded in a `metadata.json` file.
+
+### Multi-rank checkpoints
+
+Checkpoints can be recorded from multiple ranks. By default, only checkpoint data from rank zero is preserved. Data from all ranks can be retained via a `keep_all_ranks` option. Train will merge checkpoint data from all ranks into a single directory for the checkpoint number. The reason is this: users can always namespace their files/dirs by rank by the `train.context.world_rank` manually, so there isn't any reason why Train should do it for them. In fact, certain frameworks such as Lightning assume merged directories and could break if we enforced per-rank subdirectories.
```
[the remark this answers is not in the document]
ericl: This makes a lot of sense. Was there a reason we didn't do it this way earlier?

reps/2023-06-06-simplify-sync.md
```diff
@@ -0,0 +1,127 @@
...
+
+Similarly, for restoration the new Checkpoint API allows users to customize restoration strategy. For example, if they wanted to implement per-rank restore, they could call `pyarrow.fs.copy_files()` on selected files within the checkpoint path, rather than using the naive `Checkpoint.to_directory()` call.
```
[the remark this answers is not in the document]
ericl: How about adding args to to_directory such as `include_paths` whitelist of files/dirs to include?

I was thinking it's much easier than the example you wrote though, for ex:

```
checkpoint = session.get_checkpoint()
os.makedirs("my_rank2_data")
pyarrow.fs.copy_files(checkpoint.path + "/rank_2", "my_rank2_data", source_filesystem=checkpoint.fs)
```

reps/2023-06-06-simplify-sync.md
```diff
@@ -0,0 +1,127 @@
...
+
+```python
+   torch.save(model, "path/to/dir")
+   train.report(checkpoint=Checkpoint.from_directory("path/to/dir"))
+```
+
+
+Train will take ownership of the specified checkpoint data, deleting the directory after the checkpoint upload finishes. The checkpoint data will then be managed by Train (e.g., Train may restore only the latest checkpoint or delete previous checkpoints to save space according to checkpoint policy).
+
+The files of the recorded checkpoint can be accessed via `result.checkpoint`. The `Checkpoint` object itself is a logical tuple of `(path, pyarrow.fs.FileSystem)`, and this tuple is immutable. Users can also get and set arbitrary metadata to these checkpoints (e.g., preprocessor configs, model settings, etc), which will be recorded in a `metadata.json` file.
```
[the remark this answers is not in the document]
ericl: I'm not sure I understand this. The `checkpoint.path` here will always point to the root of the checkpoint dir, and the user shouldn't be able to assume anything about the path above that. This is the only guarantee this API provides.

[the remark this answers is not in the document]
ericl: > checkpoint metadata gets loaded back

To be clear, the persisted metadata (metadata.json) shouldn't contain any path data at all.

krfricke, line 63: Is this metadata mutable after creating the Checkpoint object?

If not, no further questions, and everything makes sense to me.

If yes, I'm wondering about the persistence guarantees. Is the metadata file written to the storage immediately after updating it?

In my opinion, it may be easier to require the metadata to be set on checkpoint creation and make it immutable.

krfricke (commented):
Thanks for the update! One last question remaining from my side.

reps/2023-06-06-simplify-sync.md
ericl: The (path, fs) pair is immutable, but the contents are mutable (we cannot prevent this anyways), as is the metadata json which is part of the contents.

krfricke (approved):
Thanks for the clarifications

pcmoritz (approved):

reps/2023-06-06-simplify-sync.md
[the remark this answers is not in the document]
justinvyu: I don't think so. I think it's an implementation detail cause by the handling of "checkpoint" events and "report metrics" events from train workers. All workers need to send the result type to the driver on each round of event handling.

reps/2023-06-06-simplify-sync.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Checkpoints
+
+While trials can write arbitrary data to the persistent trial directory, they cannot be guaranteed they can read it back on checkpoint/restore. Data that is needed for resume must be recorded via the `train.report(checkpoint)` API.
+
+For example, this is how you could record a Torch checkpoint:
+
+```python
+   torch.save(model, "path/to/dir")
+   train.report(checkpoint=Checkpoint.from_directory("path/to/dir"))
```
matthewdeng, line 57: Just confirming, will there be a change to the structure in which this is persisted compared to the current?

In the current implementation, the _contents_ of the checkpoint directory will be moved to a `checkpoint_{id}` directory. If I remember correctly, with this proposal we would want to store the opaque checkpoint dir/file in the `checkpoint_{id}` directory. Is that right?

Original directory:
```
path/to/dir/model.pt
```
Current:
```
{storage_path}/{experiment}/{trial}/checkpoint_{id}/model.pt
```

Proposal?
```
{storage_path}/{experiment}/{trial}/checkpoint_{id}/dir
```

reps/2023-06-06-simplify-sync.md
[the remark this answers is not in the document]
justinvyu: I see, my comment is more about being careful how to do book-keeping on this (path, fs) tuple internally.

For an s3 checkpoint where `storage_path=s3://bucket`, the checkpoint tuple that can be used to download it would be:

`(s3://bucket/exp_dir/trial_dir/ckpt_000, s3 filesystem)`

But, we should make sure not rely on this absolute s3 path for restoration. This is what we currently do to figure out what checkpoint to resume training from. You run into a problem if the s3 bucket gets moved before restoring the run.

reps/2023-06-06-simplify-sync.md
```diff
@@ -0,0 +1,126 @@
...
+```
+
+
+Train will take ownership of the specified checkpoint data, deleting the directory after the checkpoint upload finishes. The checkpoint data will then be managed by Train (e.g., Train may restore only the latest checkpoint or delete previous checkpoints to save space according to checkpoint policy).
+
+The files of the recorded checkpoint can be accessed via `result.checkpoint`. The `Checkpoint` object itself is a logical tuple of `(path, pyarrow.fs.FileSystem)`, and this tuple is immutable. Users can also get and set arbitrary metadata to these checkpoints (e.g., preprocessor configs, model settings, etc), which will be recorded in a `metadata.json` file.
```
matthewdeng, line 63: nit: I saw in the prototype that the `metadata` file is using pickle (as opposed to JSON). Is this something that needs to be aligned on in this REP? cc @pcmoritz I think you were also fond of using `msgpack`?

reps/2023-06-06-simplify-sync.md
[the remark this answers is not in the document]
justinvyu: I'm thinking about the deepspeed checkpoint structure, where all shards are aggregated in a single directory rather than in separate rank folders. Uploading shouldn't require any special logic if rank-specific files are named properly. Downloading may be a pain for users to interact with this low level `pyarrow.fs` interface, if they only want certain files.

I think I'm fine just sticking with the simplest option for now, which is to not add any helpers on top of `pyarrow.fs`. However, I think it'd be helpful to include some snippets that show common custom usage patterns for uploading and downloading.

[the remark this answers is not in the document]
justinvyu: For example, how does a user do a custom upload? Do they use `pyarrow.fs.copy_files` directly, then call `session.report(..., checkpoint=Checkpoint.from_directory("s3://place/where/I/uploaded")`? In that case, would `Checkpoint.from_path` make more sense as the path of an arbitrary filesystem?

reps/2023-06-06-simplify-sync.md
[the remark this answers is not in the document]
ericl: Right, we shouldn't store absolute path data so that restores can happen even if the data dir was moved.

reps/2023-06-06-simplify-sync.md
ericl: There is no change, the current is the same as the new proposal behavior.

reps/2023-06-06-simplify-sync.md
ericl: I think the reason to use pickle here is to allow Preprocessor objects to be contained. Alternatively, we could implement a JSON-serializer for preprocessors.

reps/2023-06-06-simplify-sync.md
matthewdeng: Ah the current implementation will "rename" the directory to `checkpoint_{id}`. See this naive example:

```python
import os
from ray.train.torch import TorchTrainer
from ray.air import ScalingConfig, session, Checkpoint

def train_func():
    checkpoint_dir = "checkpoint_dir"
    os.makedirs(checkpoint_dir, exist_ok=True)
    open(f"{checkpoint_dir}/checkpoint_file1", "w").close()
    open(f"{checkpoint_dir}/checkpoint_file2", "w").close()
    session.report({}, checkpoint=Checkpoint.from_directory(checkpoint_dir))

trainer = TorchTrainer(train_loop_per_worker=train_func, scaling_config=ScalingConfig(num_workers=2))
trainer.fit()
```

Result:
```
├── TorchTrainer_70610_00000_0_2023-07-17_16-48-33
│   ├── checkpoint_000000
│   │   ├── _preprocessor
│   │   ├── checkpoint_file1
│   │   └── checkpoint_file2
```

Note that there is no `checkpoint_dir`  in `checkpoint_000000`.

In the proposal did we want to move to something like this?

```
├── TorchTrainer_70610_00000_0_2023-07-17_16-48-33
│   ├── checkpoint_000000
│   │   ├── _metadata
│   │   ├── checkpoint_dir
│   │   │   ├── checkpoint_file1
│   │   │   └── checkpoint_file2
```

reps/2023-06-06-simplify-sync.md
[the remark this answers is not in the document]
ericl: Ok, sounds like we should probably make this change, but it can be decoupled from this REP. I don't think there's any conflict beyond refactoring Train to lift the restriction.

reps/2023-06-06-simplify-sync.md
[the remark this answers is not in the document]
ericl: The contract for `report` is that they provide a valid Checkpoint, which can point locally or anywhere. You can construct checkpoints pointing to arbitrary files using the raw Checkpoint constructor (e.g., `Checkpoint("s3://foo/bar")` or `Checkpoint("s3://foo/bar", filesystem)`.

reps/2023-06-06-simplify-sync.md
ericl: I don't think so--- if you look at the prototype we don't have nested dirs created. The data is directly under `checkpoint_000000`.

matthewdeng: Oh I see. We previously discussed supporting creating a checkpoint directly from a single file (as opposed to directory), should we still support that?

ericl: I do not think we should support that.

reps/2023-06-06-simplify-sync.md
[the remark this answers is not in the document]
justinvyu: Just to walk through a concrete example of saving to different rank checkpoint folders:

```python
checkpoint_dir = "/tmp/ckpt/rank_k/model.pt"
session.report(Checkpoint("/tmp/ckpt")) # or Checkpoint("/tmp/ckpt/rank_k") with Matt's suggestion

# This will now show up as:
# storage_path/exp_dir/trial_dir/checkpoint_000x/rank_k/...
```

Is `Checkpoint.from_directory` even needed anymore? What about just having one way to create a checkpoint -- with the constructor.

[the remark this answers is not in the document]
justinvyu: One consideration for custom uploading:

```python
for step in range(100):
    # Upload checkpoint to s3 manually
    # Ex: I want all my model checkpoints in the same bucket
    checkpoint_uri = f"s3://bucket/checkpoints/ckpt_0000{step}/"
    pyarrow.fs.copy_files("/tmp/ckpt", checkpoint_uri)

    # Report this s3 checkpoint to AIR
    session.report(Checkpoint(checkpoint_uri))

    # Which behavior makes sense here?
    # 1. Move the checkpoint to the trial directory (in which case there was no point doing this custom uploading)
    # 2. Copy the checkpoint to the trial dir
    # 3. No move/copy, just keep the checkpoint where it already is, if the user passed in a non-local URI.
```

I think we can just stick with the "move" policy as the default, and add customization options to be able to achieve any of the 3 behaviors listed above. One possibility is allowing customization of a checkpoint upload function:

```python
def default_upload_fn(checkpoint, storage):
    pyarrow.fs.copy_files(...)
    pyarrow.fs.delete(...)
    return new_checkpoint_path

def copy_upload_fn(checkpoint, storage):
    pyarrow.fs.copy_files(...)
    return new_checkpoint_path

def noop_upload_fn(checkpoint, storage):
    return checkpoint.path
```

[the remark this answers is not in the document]
justinvyu: Actually, I think custom upload is not a very big issue. Reporting a generic local directory is already very flexible and covers most use cases.

reps/2023-06-06-simplify-sync.md
pcmoritz: Updated discussion from slack -- I think this is what we are aligning on but let me know if you disagree:

We will have a Preprocessor.from_json and Preprocessor.to_json. To start with, it can use pickle under the hood and give a warning that this is not a stable format cannot be used for persistence across library versions. Then going forward, we can implement a true JSON serializer for preprocessors as a P1 and swap it out.

pcmoritz: Also I updated my opinion and think `JSON` is going to be better than msgpack since it is more universal and also the pain for serializing tensors is about the same for both. Also in Ray Serve we already have some conventions to use tensors in JSON, so it is better to lean on that as a convention rather than introduce another thing :)

ericl:

PR #36, opened 2023-07-09
[REP] Refining the Ray AIR Surface API

pcmoritz:
This REP proposes to remove the `ray.air` namespace and put the functionality into the respective libraries Ray Data, Ray Serve and Ray Train.

### DIFF
--- a/reps/2023-07-08-air-surface-syntax.md
+++ b/reps/2023-07-08-air-surface-syntax.md
@@ -11,6 +11,7 @@ libraries
 - Ray Data for batch inference and last mile data processing and ingestion,
 - Ray Train for machine learning training and
 - Ray Serve for model and application serving
+
 together.
 
 One piece of feedback we have frequently received from users is that they are confused how Ray AIR
@@ -18,6 +19,7 @@ relates to the individual libraries. In particular:
 - When should I use AIR's abstractions (e.g. should I use `BatchPredictor` or use Ray Data's map functionality,
 should I use `PredictorDeployment` or deploy my model with Ray Serve directly?) and
 - How does the `ray.air` namespace relate to `ray.data`, `ray.train` and `ray.serve`?
+
 The `ray.air` namespace both containing low level common utilities as well as highler level
 abstraction adds to this confusion. We have also learned that the higher level abstractions we
 originally introduced for Ray AIR become unneccessary and the same functionality can nicely be achieved
@@ -91,6 +93,8 @@ air.session.get_world_rank       -> train.get_context().get_world_rank
 air.session.get_local_rank       -> train.get_context().get_local_rank
 air.session.get_local_world_size -> train.get_context().get_local_world_size
 air.session.get_node_rank        -> train.get_context().get_node_rank
+
+del air
 ```
 
 These changes are ready to try out with https://github.com/ray-project/ray/pull/36706 and we encourage user feedback on the changes.
--- a/reps/2023-07-08-air-surface-syntax.md
+++ b/reps/2023-07-08-air-surface-syntax.md
@@ -54,7 +54,7 @@ The proposal will be open to the public, but please suggest a few experienced Ra
 ### Shepherd of the Proposal (should be a senior committer)
 To make the review process more productive, the owner of each proposal should identify a **shepherd** (should be a senior Ray committer). The shepherd is responsible for working with the owner and making sure the proposal is in good shape (with necessary information) before marking it as ready for broader review.
 
-@pcmoritz
+@ericl
 
 ## Details of the API changes
 
--- a/reps/2023-07-08-air-surface-syntax.md
+++ b/reps/2023-07-08-air-surface-syntax.md
@@ -145,6 +145,23 @@ This is almost as simple but a lot more explicit, removes the magic, and can
 be easily adapted to different settings. Furthermore it is more unified with
 the Ray Serve documentation and the way Ray Serve is typically used.
 
+## Internal changes
+
+As part of this effort, we are also making the recommendation to completely
+remove the `air` namespace also for internal use (just to make things clearer
+for developers). This work does not need to be connected to a specific release
+and here is an idea on where things could go:
+
+- `air.examples` -- don't have the examples in the source tree, instead put
+them into the `ray/doc` folder
+- `air.execution` -- due to the layering of Tune depending on Train but not
+the other way around, most likely `train._internal` is the right place for these.
+- `air.util` -- the tensor extension functionality should go into `ray.data`,
+the torch related functions into `ray.train.torch`.
+
+If there are any other common internal utilities that are unaccounted for,
+most likely `train._internal` is a good place to put them.
+
 ## Migration Plan
 
 We acknowledge that these kinds of API changes are very taxing on our users and we paid special attention that the migration can be done
--- a/reps/2023-07-08-air-surface-syntax.md
+++ b/reps/2023-07-08-air-surface-syntax.md
@@ -124,19 +124,19 @@ to
 import pandas as pd
 from starlette.requests import Request
 from ray import serve
-from ray.train.xgboost import XGBoostPredictor
+from ray.train.xgboost import XGBoostTrainer
 
 # checkpoint = ...
 
 @serve.deployment
 class XGBoostService:
     def __init__(self, checkpoint):
-        self.predictor = XGBoostPredictor.from_checkpoint(checkpoint)
+        self.model = XGBoostTrainer.get_model(checkpoint)
 
     async def __call__(self, http_request: Request):
         input = await http_request.body()
         data = pd.read_json(input.decode(), **http_request.query_params)
-        return self.predictor.predict(data)
+        return self.model.predict(data)
 
 serve.run(XGBoostService.bind(checkpoint))
 ```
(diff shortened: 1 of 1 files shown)


reps/2023-07-08-air-surface-syntax.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Required Reviewers
+The proposal will be open to the public, but please suggest a few experienced Ray contributors in this technical domain whose comments will help this proposal. Ideally, the list should include Ray committers.
+
+@matthewdeng, @krfricke
+
+### Shepherd of the Proposal (should be a senior committer)
+To make the review process more productive, the owner of each proposal should identify a **shepherd** (should be a senior Ray committer). The shepherd is responsible for working with the owner and making sure the proposal is in good shape (with necessary information) before marking it as ready for broader review.
+
+@pcmoritz
```
ericl, line 57:
```suggestion
@ericl
```

Don't think you should self-shepherd :)

reps/2023-07-08-air-surface-syntax.md
```diff
@@ -0,0 +1,114 @@
...
+air.session.get_node_rank        -> train.get_context().get_node_rank
+
+del air
+```
+
+These changes are ready to try out with https://github.com/ray-project/ray/pull/36706 and we encourage user feedback on the changes.
+
+## Open Questions
+
+We are likely going to remove `PredictorWrapper` and `PredictorDeployment` and migrate the examples to use Ray Serve deployments
```
ericl, line 104: Could we flesh out an example of this migration in the REP? It may make it more clear the implications of this change.

reps/2023-07-08-air-surface-syntax.md
pcmoritz: Haha, good point :)

reps/2023-07-08-air-surface-syntax.md
pcmoritz: Yes I will do that, great point.

reps/2023-07-08-air-surface-syntax.md
```diff
@@ -0,0 +1,114 @@
...
+# Ray Train configurations:
+
+air.config.CheckpointConfig      -> train.CheckpointConfig
+air.config.FailureConfig         -> train.FailureConfig
+air.config.RunConfig             -> train.RunConfig
+air.config.ScalingConfig         -> train.ScalingConfig
+
+# Ray TrainContext methods:
+
+air.session.get_checkpoint       -> train.get_context().get_checkpoint
```
krfricke, line 85: IMO this should be top-level (`train.get_checkpoint`) c.f. `get_dataset_shard` and to avoid using the context for most workflows/examples

reps/2023-07-08-air-surface-syntax.md
```diff
@@ -0,0 +1,114 @@
...
+
+del air
+```
+
+These changes are ready to try out with https://github.com/ray-project/ray/pull/36706 and we encourage user feedback on the changes.
+
+## Open Questions
+
+We are likely going to remove `PredictorWrapper` and `PredictorDeployment` and migrate the examples to use Ray Serve deployments
+direcly, and we are also likely going to move `air.integrations` to `train.integrations` and tentatively the predictors to `ray.train`.
```
krfricke, line 105: There's a few more:

- `air.callbacks` --> also train?
- `air.examples` for cross-library examples --> docs? (related question: how do we restructure the docs? I guess out of scope for this REP...)
- `air.execution` --> tentatively train?
- `air.util` - looks like mostly data?

krfricke (commented):
I'm fine with this proposal and only have detail questions left.

Just for completeness (as it's not mentioned in the REP). An alternative to resolving this is by moving everything into AIR instead. Thus we would have `ray.air.training`, `ray.air.tuning`, etc. In such a world we would soften the boundaries between the libraries and keep shared modules in the AIR namespace. By moving the libraries "one level down" we would enforce a separation from Ray Core and double down on AIR as an umbrella for our downstream libraries.

It also avoids the question of where to put which modules. Integrations and callbacks can quite naturally remain in `ray.air`, and both training and tuning modules can access them.

pcmoritz:
On your higher level comment: Not only would introducing `ray.air.training`, `ray.air.tuning` (and we would also need `ray.air.serving`) be a far higher amount of API changes, but it would also lead us to a worse outcome -- everything actually fits neatly into one of the libraries and there is really no need for a `ray.air` namespace. The libraries `ray.train`, `ray.tune`, `ray.serve`, `ray.data` and `ray.rllib` are already well separated from ray core, so I don't think there is danger of confusion here :)

reps/2023-07-08-air-surface-syntax.md
pcmoritz: So I think the reasoning in https://github.com/ray-project/ray/pull/36706#discussion_r1253792073 is still valid here and this probably should be in the context, but happy to revisit that with the checkpointing changes @ericl @justinvyu @matthewdeng :)

reps/2023-07-08-air-surface-syntax.md
matthewdeng, line 105: nit: Predictors have always been in `ray.train`.

reps/2023-07-08-air-surface-syntax.md
matthewdeng: To me `get_checkpoint` feels similar to both `get_dataset_shard` and `report`.

If I pass in a dataset, I call `get_dataset_shard`. If I pass in a checkpoint, I call `get_checkpoint`.

To save a checkpoint, I call `report`. To load a checkpoint, I call `get_checkpoint`.

With that said, if this is controversial we can always add an alias in the future.

reps/2023-07-08-air-surface-syntax.md
matthewdeng: Would it make sense to capture all these in the REP to make it more comprehensive, i.e. to describe what it takes to fully "Disband the `ray.air` namespace"?

reps/2023-07-08-air-surface-syntax.md
amogkam: +1 to @ericl. 

I think the distinction can be made by separating configurations or Ray Train framework information vs. application information. I think this aligns with how context is used in other Ray libraries (like Ray DataContext).

So things like `experiement_name`, `trial_id`, `trial_resources` make sense under a context namespace.

However, informations that is useful for the application itself makes more sense as top level namespace. This includes `get_dataset_shard`, `get_checkpoint`, `local_rank`, `world_size`, etc. These concepts are not specific to Ray Train, but rather apply to ML training general. Unlike the previous category, this information is necessary for the actual ML application itself.

amogkam:
The only concern is if we are ok with this type of usage for Ray Tune only use case?

```python

from ray import tune
from ray import train

def objective(config):
    score = config["a"] ** 2 + config["b"]
    train.report({"score": score})

search_space = {
    "a": tune.grid_search([0.001, 0.01, 0.1, 1.0]),
    "b": tune.choice([1, 2, 3]),
}

tuner = tune.Tuner(objective, param_space=search_space)

results = tuner.fit()
print(results.get_best_result(metric="score", mode="min").config)
```

This exposes the `ray.train` namespace to Tune only users. This was one of the motivations for consolidating under `ray.air`, so just want to make sure we are considering this case.

reps/2023-07-08-air-surface-syntax.md
pcmoritz: Alright, I moved get_checkpoint into `train`.

I do think `local_rank` and `world_size` are infrequently enough accessed and close to a context that they should belong into the context. It is very analogous to `.get_node_id()` of `runtime_context` :)

reps/2023-07-08-air-surface-syntax.md
pcmoritz: fixed

reps/2023-07-08-air-surface-syntax.md
pcmoritz: added now

reps/2023-07-08-air-surface-syntax.md
```diff
@@ -0,0 +1,155 @@
...
+from starlette.requests import Request
+from ray import serve
+from ray.train.xgboost import XGBoostPredictor
+
+# checkpoint = ...
+
+@serve.deployment
+class XGBoostService:
+    def __init__(self, checkpoint):
+        self.predictor = XGBoostPredictor.from_checkpoint(checkpoint)
```
ericl, line 134:
```suggestion
        local_dir = checkpoint.to_directory()
        self.model: xgboost.Booster = pickle.loads(local_dir + "/saved_booster.pkl")
``

Since we are de-emphasizing predictors, shall we also change this to show something like this instead?

ericl (approved):

krfricke (approved):
Thanks for the updates

reps/2023-07-08-air-surface-syntax.md
pcmoritz: I fixed this now by adopting the new way from https://docs.google.com/document/d/1J-09US8cXc-tpl2A1BpOrlHLTEDMdIJp6Ah1ifBUw7Y/edit :)

pcmoritz:
@amogkam We decided that it is ok for Tune to depend on Train, since tuning training runs is going to be the most important use case of Ray Tune and also that's how most machine learning engineers / practitioners think about the relation between training and tuning :)

matthewdeng (approved):
Looks great and feels a lot cleaner!

ericl:
@zhe-thoughts this can be merged

zhe-thoughts:
Thanks for the work @pcmoritz @ericl @krfricke  @matthewdeng @amogkam

PR #39, opened 2023-07-27
[REP] AWS accelerators trn1_inf support

chappidim:
### Details

This enhancement proposal briefly talks about AWS accelerators (Trainium/Inferentia) support on Ray.

### Related
[33504](https://github.com/ray-project/ray/issues/33504)

### DIFF
--- a/reps/2023-07-27-aws-accelerator-support.md
+++ b/reps/2023-07-27-aws-accelerator-support.md
@@ -1,0 +1,131 @@
+# AWS accelerators support on Ray
+
+## Background
+*[Feel free to skip this section if you are already familiar with AI accelerator, AWS trn1 EC2 instance and NeuronCore]*
+
+An AI Accelerator is a dedicated processor designed to accelerate machine-learning (ML) computations. These are
+specialized hardware designed to improve performance, reduce latency and reduce cost of deploying ML based applications.
+
+In late 2022, AWS announced general availability of Trn1 EC2 instances which are powered by AWS Trainium accelerators.
+AWS Trainium accelerator is an AI accelerator, purpose built for high-performance deep learning (DL) training of 
+generative AI models, including large language models (LLMs) and latent diffusion models[1].
+
+Each Trainium accelerator (aka NeuronDevice) includes two second-generation NeuronCores(NeuronCore-v2). 
+It is designed for speed model training. NeuronCore-v2 is a fully-independent heterogeneous compute-unit,
+with multiple engines (Tensor/Vector/Scalar.. Engines), and on-chip memory, for maximizing data locality[2].
+Also, Inferentia2 accelerator(inf2 which supports neuron-core) is designed to speed up inference.
+
+
+## Summary
+
+[Phase-1] Currently, Ray supports limited accelerators (only NVIDIA hardware) for GPUs which does not include
+AWS Trainium/Inferentia2 accelerators.
+
+[Phase-2] Also, Ray Train only supports Pytorch but not Pytorch-XLA (Accelerated Linear Algebra) which is a connector
+between Pytorch deep learning framework and TPUs/NeuronCores.
+Without these, customers can neither use AWS Trainium/Inferentia2 accelerators on Ray cluster by default nor use it for
+distributed training on Ray Train.
+
+## Stewardship
+
+### Required Reviewers
+@scv119 @matthewdeng
+
+### Shepherd of the Proposal (should be a senior committer)
+@scv119 @matthewdeng
+
+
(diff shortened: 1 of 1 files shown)


scv119:
Will take a first round review this weekend!

scv119:
looks there are two main questions:

- should we use a different resource name other than GPU?

this is mainly because ray train code are expected no code change if GPU is available. it seems not the case for AWS’s accelator (requries code change). If we want to stick with GPU, we should have a way to differentiate nvidia GPU from other xla accelators.

- does all xla devices follow the same pytorch API?

this is somewhat related to the previous question. I assume it's the case, and we should make sure the API works out of box for other xla devices, like TPU

chappidim:
> 

Thanks for the initial review.

1. +1 to that, I'm open for adding new predefined resource-type if aws_accelerators doesn't fit as GPU. The only counterpart to that is per [this docs](https://github.com/ray-project/ray/blob/master/python/ray/util/accelerators/accelerators.md) we treat any other accelerator as GPU.
2. AWS expects users to install a custom built [torch_xla](https://pip.repos.neuron.amazonaws.com/torch-xla/). Implies, torch_xla_neuron may not support some of the APIs but the package follows all the API standards listed by open-source torch-xla.

reps/2023-07-27-aws-accelerator-support.md
```diff
@@ -0,0 +1,131 @@
...
+
+```text
+class _TorchBackend(Backend):
+    def on_start():
+        # support xla backend
+        # Configure master env of xla device related to torchrun/torch-elastic
+    def on_shutdown():
+        # cleanup NeuronCore cache if needed
+    def on_training_start():
+        # configure rank/world_size/node_rank based on xla device
```
matthewdeng, line 91: I'm a bit fuzzy on whether XLA is expected to sit at the same level in the stack as `nccl` and `gloo`, based on my knowledge of [Torch backends](https://pytorch.org/docs/stable/distributed.html). 

As an alternative approach, would the following interface make logical sense? Is this the right layer of abstraction? 
```python
class TorchXLAConfig(TorchConfig)

    @property
    def backend_cls(self):
        return _TorchXLABackend

class _TorchXLABackend(_TorchBackend):
    # XLA specific logic here

# User defined code
trainer = TorchTrainer(torch_config=TorchXLAConfig(...))
```

To better understand how to think about this, I'd love to learn more about how Torch XLA environment are typically set up and configured in practice - do you have any pointers to any best practices or other references I could take a look at?

chappidim: I'm a newbie to this space, as I learn more it makes sense to have separate XLAConfig with Backend. Wondering if we want to be more explicit on backend as it can vary per XLA device? `TorchXLAConfig/_TorchAwsNeuronXLABackend` where the basic setup (Rank/WorldSize,MasterPort/Addr) is already done by NeuronSDK [1] and include anything related to torchrun[2]. 


I'm happy to ask around some SMEs around this area but here's the information I gathered so far.
1. Configure TPU library configuration [3][4]
2. Configure pjrt(latest)/xrt [5] 
3. Configure world/local, rank/size, master add/port (for torchrun) - generic to torch


[1] https://sagemaker.readthedocs.io/en/v2.167.0/frameworks/pytorch/using_pytorch.html#distributed-training-with-pytorch-neuron-on-trn1-instances
[2] https://awsdocs-neuron.readthedocs-hosted.com/en/latest/frameworks/torch/torch-neuronx/tutorials/training/mlp.html?highlight=torchrun#multi-worker-data-parallel-mlp-training-using-torchrun
[3] https://github.com/pytorch/xla/blob/master/TROUBLESHOOTING.md#troubleshooting
[4] https://lightning.ai/docs/pytorch/stable/accelerators/tpu_faq.html
[5] https://github.com/pytorch/xla/blob/424d8c8eec4e7fd3af83749da2652b2fcd6d207a/configuration.yaml

chappidim:
> looks there are two main questions:
> 
> * should we use a different resource name other than GPU?
> 
> this is mainly because ray train code are expected no code change if GPU is available. it seems not the case for AWS’s accelator (requries code change). If we want to stick with GPU, we should have a way to differentiate nvidia GPU from other xla accelators.
> 
> * does all xla devices follow the same pytorch API?
> 
> this is somewhat related to the previous question. I assume it's the case, and we should make sure the API works out of box for other xla devices, like TPU

@scv119 Checking if we got enough quorum on adding `num_neuron_cores` as pre-defined resource? Thanks

edoakes:

PR #42, opened 2023-08-18
[Serve] Java Serve improvement

liuyang-my:


reps/2023-08-18-serve-java-dag-api.md
```diff
@@ -0,0 +1,266 @@
...
+        return inp + self.increment
+
+
+with InputNode() as inp:
+    model = Model.bind(increment=2)
+    output = model.predict.bind(preprocess.bind(inp))
+    serve_dag = DAGDriver.bind(output)
+
+handle = serve.run(serve_dag)
+assert ray.get(handle.predict.remote(1)) == 4
```
edoakes, line 31: @liuyang-my we are actually discouraging the use of the "call graph" api using `with InputNode` and instead encouraging users to use the deployment handle interface directly to perform composition. For this example it'd look like:

```python
@serve.deployment
def preprocess(inp: int) -> int:
    return inp + 1


@serve.deployment
class Model:
    def __init__(self, preprocess_handle: RayServeHandle, increment: int):
        self.preprocess_handle: DeploymentHandle = preprocess_handle.options(use_new_handle_api=True)
        self.increment = increment

    async def predict(self, inp: int): -> int:
        preprocessed = await self.preprocess_handle.remote(inp)
        return inp + self.increment

app = Model.bind(preprocess.bind(), increment=2)

handle = serve.run(app)
assert ray.get(handle.predict.remote(1)) == 4
```

I think we can match this API very closely in Java

reps/2023-08-18-serve-java-dag-api.md
```diff
@@ -0,0 +1,266 @@
...
+    Application deployment =
+        Serve.deployment().setDeploymentDef(DeploymentDemo.class.getName()).bind();
+    RayServeHandle handle = Serve.run(deployment).get();
+    System.out.println(handle.remote().get());
+  }
+}
+
+```
+In this demo, a DAG node is defined through the `bind` method of the Deployment, and it is deployed using the `Serve.run` API.
+Furthermore, a Deployment can bind other Deployments, and users can use the Deployment input parameters in a similar way to `RayServeHandle`. For example:
```
edoakes, line 76: Looks great!

reps/2023-08-18-serve-java-dag-api.md
```diff
@@ -0,0 +1,266 @@
...
+  private RayServeHandle modelBHandle;
+
+  public Driver(RayServeHandle modelAHandle, RayServeHandle modelBHandle) {
+    this.modelAHandle = modelAHandle;
+    this.modelBHandle = modelBHandle;
+  }
+
+  public String call(String request) {
+    ObjectRef<Object> refA = modelAHandle.remote(request);
+    ObjectRef<Object> refB = modelBHandle.remote(request);
```
edoakes, line 95: in the most recent Python API we are not returning an ObjectRef directly but have a wrapping object called a `DeploymentResponse`. we might consider doing the same for java.

the main reason for doing it in python is to avoid having to "double await" to get the final result. maybe that's less relevant here.

reps/2023-08-18-serve-java-dag-api.md
```diff
@@ -0,0 +1,266 @@
...
+
+### Deploying deployments of other languages through Python API
+In another REP ([Add Cpp Deployment in Ray Serve](https://github.com/ray-project/enhancements/pull/34)), it is mentioned how to deploy C++ deployments through Python. Deploying Java deployments through Python is similar. Since Java and C++ do not have the decorator mechanism like Python, a straightforward way is to directly instantiate the corresponding Deployment object:
+
+```python
+deployment_config = DeploymentConfig()
+deployment_config.deployment_language = JAVA # or CPP
```
edoakes, line 230: these are private APIs, so this won't work. the alternative below looks reasonable to me, though we should not use the private `_func_or_class` kwarg. it can just be the first positional.

edoakes (commented):
This looks like a great step forward. A few high-level comments:

- As I indicated below, we are moving away from the "InputNode" and "DAGDriver" concepts (they're unnecessarily complicated). For Java, let's prioritize matching the `.bind()` API for multiple deployments and using the handles that are substituted in the constructor.
- The recommended way for users to deploy in production these days is via config file (e.g., embedded in a KubeRay CR). Can we include in the proposal adding support to deploy Java this way as well?
- [P1] As part of this plan, it would be great to have a path forward to unify the handle implementations (e.g., shared C++ core).

reps/2023-08-18-serve-java-dag-api.md
liuyang-my: It has been adjusted.

reps/2023-08-18-serve-java-dag-api.md
liuyang-my: Alright, I have made the adjustments according to the suggestions.

reps/2023-08-18-serve-java-dag-api.md
liuyang-my: adjusted

reps/2023-08-18-serve-java-dag-api.md
```diff
@@ -0,0 +1,162 @@
...
+
+### Deploying Deployments of the Other Languages through Python API
+In another REP ([Add Cpp Deployment in Ray Serve](https://github.com/ray-project/enhancements/pull/34)), it is mentioned how to deploy C++ deployments through Python. Deploying Java deployments through Python is similar. Since Java and C++ do not have the decorator mechanism like Python, a straightforward way is to directly use the `serve.deployment` API (with the addition of a new `language` parameter):
+
+```python
+deployment = serve.deployment('io.ray.serve.ExampleDeployment', name='my_deployment', language=JAVA)
+
+```
+### Deploying through the Config File
+// TODO
```
edoakes, line 129: Please fill this in :) @zcin or @shrekris-anyscale can help point you in the right direction / provide suggestions

edoakes:
@liuyang-my looks great now, just need to add the config section!

reps/2023-08-18-serve-java-dag-api.md
```diff
@@ -0,0 +1,162 @@
+## Summary
+### General Motivation
+Compared to the latest Serve API in Python, Ray Serve Java lacks a major version update. Currently, the main form of the Serve API is the DAG API, which involves networking deployments through binding and deploying the graph. For example:
+```python
+import ray
+from ray import serve
+from ray.serve.dag import InputNode
```
sihanwang41, line 7: remove InputNode and DAGDriver. Deprecating in python.

reps/2023-08-18-serve-java-dag-api.md
```diff
@@ -0,0 +1,162 @@
...
+  public static void main(String[] args) {
+    Application deployment =
+        Serve.deployment().setDeploymentDef(DeploymentDemo.class.getName()).bind();
+    RayServeHandle handle = Serve.run(deployment).get();
+    System.out.println(handle.remote().get());
+  }
+}
+
+```
+In this demo, a DAG node is defined through the `bind` method of the Deployment, and it is deployed using the `Serve.run` API.
```
sihanwang41, line 72: no `DAG` node :)

reps/2023-08-18-serve-java-dag-api.md
```diff
@@ -0,0 +1,162 @@
...
+            url,
+            blocking);
+  }
+}
+```
+This is also done to maintain consistency with the Python API, and to allow for easy removal of the deprecated API in the future.
+## Test Plan and Acceptance Criteria
+Related test cases will be provided under ray/java/serve, and they will cover the three scenarios mentioned above.
+## (Optional) Follow-on Work
+- Modify the Ray Serve Java API to support the usage of DAGs.
```
sihanwang41, line 160: remove this

reps/2023-08-18-serve-java-dag-api.md
liuyang-my: Regarding the `Config File`, I have added a demo. Could you please review it again?

reps/2023-08-18-serve-java-dag-api.md
```diff
@@ -0,0 +1,225 @@
...
+        return (await hello_response) + (await world_response)
+
+
+hello = Hello.bind()
+world = world_java.bind()
+
+app = Ingress.bind(hello, world)
+
+```
+In this code, we define the Java Deployment `World` using the Python API and then bind it with a Python Deployment `Hello` into the `Ingress` to form an app. By using the `serve build` command, the Serve config file can be generated.
```
edoakes, line 163: If we only wanted to deploy `"io.ray.serve.World"`, would we still need to define this Python file?

Ideally we would be able to provide a java import path directly under `applications`. For example:

```
applications:
- name: python_app
   route_prefix: /python
   import_path: python_file:app
- name: java_app
   route_prefix: /java
   import_path: io.ray.serve.World # This needs to be some kind of similar "app builder" like the Python version!
```

edoakes: In python we have support for defining an "app builder" function which is imported and run with some provided args, then returns the built app object. Docs:
https://docs.ray.io/en/latest/serve/advanced-guides/app-builder-guide.html?highlight=builder#defining-an-application-builder

I think we could mirror this directly for Java.

liuyang-my: This suggestion is excellent, and I hadn't considered this aspect thoroughly before. I have now added the necessary supplements to address this.

reps/2023-08-18-serve-java-dag-api.md
```diff
@@ -0,0 +1,325 @@
...
+  }
+}
+
+```
+This code orchestrates an application within a static method named `app`. The CLI command for its deployment is as follows:
+```shell
+$ serve run java:io.ray.serve.repdemo.Text:app
+```
+
+> It should be noted that `java` is mentioned here to represent the programming language used for orchestrating the application. Currently, there is no clear plan on how to expand this parameter in the serve run command. One direct approach that comes to mind is to concatenate it before the `import_path`.
```
edoakes, line 176: How about just add a `--language=java` flag? and can also be set via env var?

reps/2023-08-18-serve-java-dag-api.md
```diff
@@ -0,0 +1,325 @@
...
+http_options:
+  host: 0.0.0.0
+  port: 8000
+
+grpc_options:
+  port: 9000
+  grpc_servicer_functions: []
+
+applications:
+- name: app
```
edoakes, line 277: Note that this can take `args:` parameter as well (will be passed in the same way that it is via command line)

edoakes (approved):
Looks great! Can't wait to see this implemented.

reps/2023-08-18-serve-java-dag-api.md
liuyang-my: Great, the example of adding args in the config file has been supplemented.

PR #46, opened 2023-10-13
[REP] Ray accelerator support

jjyao:
How to support new accelerators in Ray end-to-end.

### DIFF
--- a/reps/2023-10-13-accelerator-support.md
+++ b/reps/2023-10-13-accelerator-support.md
@@ -108,6 +108,24 @@ def task()
     ...
 ```
 
+##### Runtime Context API
+
+Currently we have `ray.get_gpu_ids()` and we need similar APIs for getting available ids of other accelerators.
+
+The new API will be inside RuntimeContext since it returns accelerator ids that are available to a worker.
+
+```python
+class RuntimeContext:
+    def get_accelerator_ids() -> Dict[str, List[str]]:
+        """
+        Get the current worker's available accelerator ids.
+
+        Returns:
+            A dictionary keyed by the resource name. The values are list
+            of ids. E.g., `{'GPU': ['0', '1']}`, `{'neuron_cores': ['0', '3']}`.
+        """
+```
+
 #### Implementation
 
 Core defines an `AcceleratorManager` base class and each accelerator needs to implement a subclass (e.g. `TPUAcceleratorManager`):
--- a/reps/2023-10-13-accelerator-support.md
+++ b/reps/2023-10-13-accelerator-support.md
@@ -218,18 +218,17 @@ Different accelerator family uses different backend to launch deep learning trai
 
 | Accelerator | Communication Backend | Developer APIs | User APIs | 
 | - | - | - | - |
-| GPU      | `nccl`, `gloo` | TorchBackend | TorchConfig |
-| TPU      | `xla[tpu]`      | TorchTPUXLABackend | TorchXLAConfig |
-| Trainium | `xla[neuronx]`  | TorchAwsNeuronXLABackend | TorchXLAConfig  |
-| HPU      | `hccl`         | TorchHPUBackend | TorchHCCLConfig |
+| GPU      | `nccl`, `gloo` | TorchBackend | `TorchConfig()` |
+| TPU      | `xla[tpu]`      | TorchTPUXLABackend | `TorchConfig("xla[tpu]")` |
+| Trainium | `xla[neuronx]`  | TorchAwsNeuronXLABackend | `TorchConfig("xla[neuronx]")`  |
+| HPU      | `hccl`         | TorchHCCLBackend | `TorchConfig("hccl")` |
 
 
 Putting things together, users should be able to set up the resources and backend with these two APIs in `TorchTrainer`.
 
 ```python
 from ray.train import ScalingConfig
-from ray.train.torch import TorchTrainer
-from ray.train.torch.tpu import TorchXLAConfig
+from ray.train.torch import TorchTrainer, TorchConfig
 
 # TPU Training
 trainer = TorchTrainer(
@@ -239,7 +238,7 @@ trainer = TorchTrainer(
         accelerator_type="TPU-V4", 
         resources_per_worker={"TPU": 1}
     ),
-    torch_config=TorchXLAConfig(backend="tpu")
+    torch_config=TorchConfig(backend="xla[tpu]")
 )
 
 # AWS Trainium
@@ -249,7 +248,7 @@ trainer = TorchTrainer(
         num_workers=4, 
         resources_per_worker={"neuron_cores": 1}
     ),
-    torch_config=TorchXLAConfig(backend="neuronx")
+    torch_config=TorchConfig(backend="xla[neuronx]")
 )
 
 # HPU
@@ -259,7 +258,7 @@ trainer = TorchTrainer(
         num_workers=4, 
         resources_per_worker={"HPU": 1}
     ),
-    torch_config=TorchHCCLConfig()
+    torch_config=TorchConfig(backend="hccl")
 )
 ```
 
--- a/reps/2023-10-13-accelerator-support.md
+++ b/reps/2023-10-13-accelerator-support.md
@@ -1,4 +1,4 @@
-# Ray Accelerator Support 
+# Ray Accelerator Support
 
 ## Summary
 
@@ -214,7 +214,10 @@ To auto populate `rayStartParams` resources using pod resources, some code is ne
 See https://github.com/ray-project/kuberay/blob/b7bc7ae8c983160bb700d6b8bb4014dda183ecea/ray-operator/controllers/ray/common/pod.go#L687 as an example of how GPU resource is populated.
 
 ### Observability
-TODO
+
+Currently we emit metrics about the physical usages of GPUs and also show them on Ray dashboard.
+Similarly, for other accelerators, we need to emit those metrics as well and update the Ray dashboard
+and Grafana dashboard to show them.
 
 ### Docker
 
@@ -227,5 +230,5 @@ TODO
 ### Testing
 
 For some of the CI tests, we can mock without actually running on the machine with those accelerators.
-But for some other CI tests and release tests, we do need machines with those accelerators to make sure real workoads can run using those accelerators successfully.
+But for some other CI tests and release tests, we do need machines with those accelerators to make sure real workoads can run successfully using those accelerators.
 
--- a/reps/2023-10-13-accelerator-support.md
+++ b/reps/2023-10-13-accelerator-support.md
@@ -1,0 +1,231 @@
+# Ray Accelerator Support 
+
+## Summary
+
+Holistic design of supporting accelerators in Ray.
+This also serves as the reference doc for contributors
+to add new accelerator support in the future.
+
+### General Motivation
+
+Nowadays, more and more specialized accelerators (e.g. TPU, HPU) are coming out to speed up AI workloads.
+Supporting those accelerators natively in Ray will increase Ray's adoption as the computing framework for scaling AI applications.
+
+### Should this change be within `ray` or outside?
+
+Inside `ray` project since we want to support those accelerators out of the box.
+
+## Stewardship
+
+### Required Reviewers
+The proposal will be open to the public, but please suggest a few experienced Ray contributors in this technical domain whose comments will help this proposal. Ideally, the list should include Ray committers.
+
+@pcmoritz, @ericl, @scv119
+
+### Shepherd of the Proposal (should be a senior committer)
+To make the review process more productive, the owner of each proposal should identify a **shepherd** (should be a senior Ray committer). The shepherd is responsible for working with the owner and making sure the proposal is in good shape (with necessary information) before marking it as ready for broader review.
+
+@scv119
+
+## Design
+
+### Ray Core
+
+#### API
+Users will use accelerators through Ray resources. For GPU accelerators (e.g. Nvidia, AMD, Intel GPUs), they will use the `GPU` resource name. For other accelerators, they will each have their own resource name (e.g. TPU accelerator has resource name `TPU`). `accelerator_type` option can also be used to specify a particular type of accelerator for a task or actor.
+
+```python
+@ray.remote(num_gpus=1)
+def task():
+    ...
+
+@ray.remote(num_gpus=1, accelerator_type="A100")
+def task():
+    ...
+
+@ray.remote(num_gpus=1, accelerator_type="Intel-Max-1550")
+def task():
+    ...
+
+@ray.remote(resources={"TPU": 1})
+def task():
+    ...
+
+@ray.remote(resources={"TPU": 1}, accelerator_type="TPU-V4")
+def task():
+    ...
+```
+
+##### Alternatives
+Other alternative APIs considered:
+
+Similar to `num_gpus`, having a top level parameter for each accelerator family:
+
+```python
+@ray.remote(num_tpus=1)
+def task():
+    ...
+
+@ray.remote(num_tpus=1, accelerator_type="TPU-V4")
+def task():
+    ...
+```
+
+A single `ACCELERATOR` resource for the ability to schedule a task on any accelerator
+and an `accelerator_family` is also introduced to allow specifying the family of the accelerators to use:
+
+```python
+@ray.remote(resources={"ACCELERATOR": 1})
+def task()
+    ...
+
+@ray.remote(resources={"ACCELERATOR": 1}, accelerator_family="TPU")
+def task()
+    ...
+
+@ray.remote(resources={"ACCELERATOR": 1}, accelerator_type="TPU-V4")
+def task()
+    ...
+```
+
+Similar to the above but having a top level `num_accelerators` parameter:
+
+```python
+@ray.remote(num_accelerators=1)
+def task()
+    ...
+
+@ray.remote(num_accelerators=1, accelerator_family="TPU")
+def task()
+    ...
+
+@ray.remote(num_accelerators=1, accelerator_type="TPU-V4")
+def task()
+    ...
+```
+
+#### Implementation
+
+Core defines an `Accelerator` base class and each accelerator needs to implement a subclass (e.g. `TPUAccelerator`):
+
+```python
+@DeveloperAPI
+class Accelerator(ABC):
+    @staticmethod
+    @abstractmethod
+    def is_available() -> bool:
+        """Detect if the accelerator is available on this node."""
+
+    @staticmethod
+    @abstractmethod
+    def get_resource_name() -> str:
+        """Get the name of the resource representing this accelerator."""
+
+    @staticmethod
+    @abstractmethod
+    def get_visible_accelerator_ids_env_var() -> str:
+        """Get the env var that sets the ids of visible accelerators."""
+
+    @staticmethod
+    @abstractmethod
+    def get_num_accelerators() -> int:
+        """Get the total number of accelerators on this node."""
+
+    @staticmethod
+    @abstractmethod
+    def get_visible_accelerator_ids() -> Optional[List[str]]:
+        """Get the ids of accelerators visible to Ray."""
+
+    @staticmethod
+    @abstractmethod
+    def get_accelerator_type() -> Optional[str]:
+        """Get the type of the accelerator on this node.
+
+        The result should only be used when get_num_accelerators() > 0.
+
+        Return None if it's unknown.
+        """
+
+    @staticmethod
+    @abstractmethod
+    def validate_resource_request_quantity(quantity: float) -> Optional[str]:
+        """Validate the resource request quantity of this accelerator resource.
+
+        Return error message if the quantity is invalid.
+        """
+
+    @staticmethod
+    @abstractmethod
+    def set_visible_accelerator_ids(ids: List[str]) -> None:
+        """Set the ids of accelerators visible to the Ray task or actor."""
+
+    @staticmethod
+    def get_ec2_num_accelerators(instance_type: str, instances: dict) -> Optional[int]:
+        """Get the number of accelerators of this aws ec2 instance type.
+
+        Return None if it's unknown.
+        """
+        return None
+
+    @staticmethod
+    def get_ec2_accelerator_type(instance_type: str, instances: dict) -> Optional[str]:
+        """Get the accelerator type of this aws ec2 instance type.
+
+        Return None if it's unknown.
+        """
+        return None
+```
+
+Besides defining the accelerator subclass, the accelerator resource needs to be marked as unit instance resource in config `RAY_custom_unit_instance_resources`.
+
+### Ray Train
+TODO
+
+### Ray RLLib
+TODO
+
+### Ray Serve
+Nothing needs to be changed on the serve side.
+
+Users can define a deployment using accelerators via `ray_actor_options`:
+
+```python
+@serve.deployment(ray_actor_options={"num_gpus": 1})
+class Deployment:
+    ...
+
+@serve.deployment(ray_actor_options={"resources": {"TPU": 1}})
+class Deployment:
+    ...
+```
+
+### Ray Data
+Nothing needs to be changed on the data side.
+
+Users can use the corresponding accelerator resources for data operations:
+
+```python
+ds.map_batches(Predicator, resources={"TPU": 1})
+```
+
+### Kuberay
+
+To auto populate `rayStartParams` resources using pod resources, some code is needed.
+See https://github.com/ray-project/kuberay/blob/b7bc7ae8c983160bb700d6b8bb4014dda183ecea/ray-operator/controllers/ray/common/pod.go#L687 as an example of how GPU resource is populated.
+
+### Observability
+TODO
+
+### Docker
+
+Currently `rayproject/ray-ml` has `-gpu` versions that are based on `NVIDIA CUDA` image.
+Similarly, for other accelerators, we need to publish docker images with their software installed.
+
+### Documentation
+TODO
+
+### Testing
+
+For some of the CI tests, we can mock without actually running on the machine with those accelerators.
+But for some other CI tests and release tests, we do need machines with those accelerators to make sure real workoads can run using those accelerators successfully.
+
--- a/reps/2023-10-13-accelerator-support.md
+++ b/reps/2023-10-13-accelerator-support.md
@@ -241,9 +241,9 @@ Besides defining the accelerator subclass, the accelerator resource needs to be 
 Ray Train will provide 2 new APIs to support different accelerator types. 
 
 
-#### 1. Specify Accelerator Type in `ScalingConfig`
+#### 1. Specify Resources in `ScalingConfig`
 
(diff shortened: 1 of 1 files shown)


reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,231 @@
...
+    ...
+```
+
+#### Implementation
+
+Core defines an `Accelerator` base class and each accelerator needs to implement a subclass (e.g. `TPUAccelerator`):
+
+```python
+@DeveloperAPI
+class Accelerator(ABC):
```
jjyao, line 113: Implementation is here: https://github.com/ray-project/ray/pull/40286

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,234 @@
...
+
+Users can use the corresponding accelerator resources for data operations:
+
+```python
+ds.map_batches(Predicator, resources={"TPU": 1})
+```
+
+### Kuberay
+
+To auto populate `rayStartParams` resources using pod resources, some code is needed.
```
jjyao, line 213: @kevin85421 for review kuberay section.

kevin85421 (approved):
The KubeRay part looks good to me. Some details might be helpful:

* How to schedule multiple bundles in the same placement group to the hosts in a SuperPOD rather than multiple SuperPODs?
* How does autoscaling work for SuperPods?
* How to set up accelerators related environment variables (e.g. TPU_WORKER_ID, TPU_WORKER_HOSTNAMES)?

See https://github.com/ray-project/ray/issues/39781#issuecomment-1743658075 for more details.

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,234 @@
...
+
+Besides defining the accelerator subclass, the accelerator resource needs to be marked as unit instance resource in config `RAY_custom_unit_instance_resources`.
+
+### Ray Train
+TODO
+
+### Ray RLLib
+TODO
+
+### Ray Serve
```
jjyao, line 187: @edoakes for reviewing the serve section.

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,234 @@
...
+@serve.deployment(ray_actor_options={"num_gpus": 1})
+class Deployment:
+    ...
+
+@serve.deployment(ray_actor_options={"resources": {"TPU": 1}})
+class Deployment:
+    ...
+```
+
+### Ray Data
```
jjyao, line 202: @raulchen for reviewing the data section.

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,234 @@
...
+## Summary
+
+Holistic design of supporting accelerators in Ray.
+This also serves as the reference doc for contributors
+to add new accelerator support in the future.
+
+### General Motivation
+
+Nowadays, more and more specialized accelerators (e.g. TPU, HPU) are coming out to speed up AI workloads.
+Supporting those accelerators natively in Ray will increase Ray's adoption as the computing framework for scaling AI applications.
```
edoakes, line 12: We should define what "supporting them natively" means at a high level. i.e., what does native support enable?

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,234 @@
...
+@ray.remote(resources={"TPU": 1})
+def task():
+    ...
+
+@ray.remote(resources={"TPU": 1}, accelerator_type="TPU-V4")
+def task():
+    ...
+```
+
+##### Alternatives
```
edoakes, line 59: For the alternatives, can you briefly describe why we went with the above instead of them?

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,234 @@
...
+    ...
+```
+
+#### Implementation
+
+Core defines an `Accelerator` base class and each accelerator needs to implement a subclass (e.g. `TPUAccelerator`):
+
+```python
+@DeveloperAPI
+class Accelerator(ABC):
```
edoakes, line 113: This class name is sort of odd given -- from the name I assumed it represented/managed a single physical accelerator, but it seems to instead define various utilities for interacting with all accelerators on a node. I'd suggest something like `AcceleratorManager` or `AcceleratorManagementInterface`

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,234 @@
...
+        """Get the total number of accelerators on this node."""
+
+    @staticmethod
+    @abstractmethod
+    def get_visible_accelerator_ids() -> Optional[List[str]]:
+        """Get the ids of accelerators visible to Ray."""
+
+    @staticmethod
+    @abstractmethod
+    def get_accelerator_type() -> Optional[str]:
```
edoakes, line 141: how's this different from `get_resource_name`?

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,234 @@
...
+#### Implementation
+
+Core defines an `Accelerator` base class and each accelerator needs to implement a subclass (e.g. `TPUAccelerator`):
+
+```python
+@DeveloperAPI
+class Accelerator(ABC):
+    @staticmethod
+    @abstractmethod
+    def is_available() -> bool:
```
edoakes, line 116:
```suggestion
    def is_available_on_current_node() -> bool:
```

Otherwise, by default i'd assume this is checking if the accelerator is available on the cluster.

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,234 @@
...
+        """Get the type of the accelerator on this node.
+
+        The result should only be used when get_num_accelerators() > 0.
+
+        Return None if it's unknown.
+        """
+
+    @staticmethod
+    @abstractmethod
+    def validate_resource_request_quantity(quantity: float) -> Optional[str]:
```
edoakes, line 151: why's the return value optional?

reps/2023-10-13-accelerator-support.md
edoakes: LGTM

edoakes (commented):
Looks like a good proposal overall. Some general comments on the API. Serve part LGTM.

reps/2023-10-13-accelerator-support.md
jjyao: `get_resource_name()` returns "TPU" but `get_accelerator_type()` returns "TPU-V4", "TPU-V5" etc.

reps/2023-10-13-accelerator-support.md
jjyao: This returns the error message if validation fails. Return None if it's valid.

reps/2023-10-13-accelerator-support.md
jjyao: I'll add more details. But basically the same support level as existing Nvidia GPU.

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,234 @@
...
+    ...
+
+@ray.remote(num_accelerators=1, accelerator_type="TPU-V4")
+def task()
+    ...
+```
+
+#### Implementation
+
+Core defines an `Accelerator` base class and each accelerator needs to implement a subclass (e.g. `TPUAccelerator`):
```
matthewdeng, line 109: A few questions that I wasn't sure if we needed to clarify in the implementation section:

1. Anything to add for accelerator types here?
2. Do we need to clarify between virtual and physical resources?
3. Maybe out of scope for this REP, but do we need to consider partial resources?

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Design
+
+### Ray Core
+
+#### API
+Users will use accelerators through Ray resources. For GPU accelerators (e.g. Nvidia, AMD, Intel GPUs), they will use the `GPU` resource name. For other accelerators, they will each have their own resource name (e.g. TPU accelerator has resource name `TPU`). `accelerator_type` option can also be used to specify a particular type of accelerator for a task or actor.
+
+```python
+@ray.remote(num_gpus=1)
```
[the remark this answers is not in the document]
jjyao: Yes. Although you have to use the `num_gpus` syntax sugar.

reps/2023-10-13-accelerator-support.md
jjyao: I got the naming from https://github.com/Lightning-AI/lightning/blob/master/src/lightning/fabric/accelerators/accelerator.py.  But if people think its confusing, we can change to `AcceleratorManager`.

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,234 @@
...
+        """
+        return None
+
+    @staticmethod
+    def get_ec2_accelerator_type(instance_type: str, instances: dict) -> Optional[str]:
+        """Get the accelerator type of this aws ec2 instance type.
+
+        Return None if it's unknown.
+        """
+        return None
```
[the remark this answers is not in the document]
jjyao: Ideally yes but currently only aws node provider supports this.

reps/2023-10-13-accelerator-support.md
jjyao: > Anything to add for accelerator types here?

You need to implement the `get_accelerator_type()` abstract method.

> Do we need to clarify between virtual and physical resources?

Here it's for auto detecting physical resources. 

> Maybe out of scope for this REP, but do we need to consider partial resources?

Partial resources are supported by Ray out of the box. We don't need to do anything here. But if the accelerator doesn't support fractional, you should return error in `validate_resource_request_quantity`

reps/2023-10-13-accelerator-support.md
edoakes: oh... maybe return a `Tuple[bool, str]` with `(valid, msg)`?

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,234 @@
...
+
+@ray.remote(num_gpus=1, accelerator_type="Intel-Max-1550")
+def task():
+    ...
+
+@ray.remote(resources={"TPU": 1})
+def task():
+    ...
+
+@ray.remote(resources={"TPU": 1}, accelerator_type="TPU-V4")
```
woshiyyya, line 54: Who defines these accelerator types? Should it be defined by the user, or do we detect it automatically?

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,234 @@
...
+To make the review process more productive, the owner of each proposal should identify a **shepherd** (should be a senior Ray committer). The shepherd is responsible for working with the owner and making sure the proposal is in good shape (with necessary information) before marking it as ready for broader review.
+
+@scv119
+
+## Design
+
+### Ray Core
+
+#### API
+Users will use accelerators through Ray resources. For GPU accelerators (e.g. Nvidia, AMD, Intel GPUs), they will use the `GPU` resource name. For other accelerators, they will each have their own resource name (e.g. TPU accelerator has resource name `TPU`). `accelerator_type` option can also be used to specify a particular type of accelerator for a task or actor.
```
woshiyyya, line 35: Are we going to maintain a mapping from accelerator type to resource name?

Do we have something like
```python
ACCELERATOR_MAPPING = {
   "A100": "GPU",
   "Intel-Max-1550": "GPU",
   "TPU-V4": "TPU",
   "Trainium": "neuron_cores",
}
```

reps/2023-10-13-accelerator-support.md
raulchen: > `ds.map_batches(Predicator, resources={"TPU": 1})`

The correct syntax should be `ds.map_batches(..., ray_remote_args={"resources": {"TPU": 1}})`
Also, we've deprecated the `Predicator` API. So better not to add it in the example. 
Otherwise looks good.

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,317 @@
...
+        """Get the accelerator type of this aws ec2 instance type.
+
+        Return None if it's unknown.
+        """
+        return None
+```
+
+Besides defining the accelerator subclass, the accelerator resource needs to be marked as unit instance resource in config `RAY_custom_unit_instance_resources`.
+
+### Ray Train
```
woshiyyya, line 181: @matthewdeng Can you review this and we can discuss offline.

reps/2023-10-13-accelerator-support.md
jjyao: Changed to `AcceleratorManager`.

reps/2023-10-13-accelerator-support.md
jjyao: Added comments to explain this.

reps/2023-10-13-accelerator-support.md
jjyao: Changed.

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,234 @@
...
+
+    @staticmethod
+    @abstractmethod
+    def get_resource_name() -> str:
+        """Get the name of the resource representing this accelerator."""
+
+    @staticmethod
+    @abstractmethod
+    def get_visible_accelerator_ids_env_var() -> str:
+        """Get the env var that sets the ids of visible accelerators."""
```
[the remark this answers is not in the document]
jjyao: Yes.

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,234 @@
...
+
+    @staticmethod
+    @abstractmethod
+    def get_num_accelerators() -> int:
+        """Get the total number of accelerators on this node."""
+
+    @staticmethod
+    @abstractmethod
+    def get_visible_accelerator_ids() -> Optional[List[str]]:
+        """Get the ids of accelerators visible to Ray."""
```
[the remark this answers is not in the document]
jjyao: Renamed to `get_current_process_visible_accelerator_ids`

reps/2023-10-13-accelerator-support.md
jjyao: It's auto detected.

reps/2023-10-13-accelerator-support.md
jjyao: No. Is there a use case for that?

reps/2023-10-13-accelerator-support.md
jjyao: Added.

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,374 @@
...
+#### 1. Specify Accelerator Type in `ScalingConfig`
+
+Ray Core automatically detects the accelerator type, and treats it as an additional resource tag. Ray Train will therefore be able to launch training jobs on specific accelerator types. We add a new initialization argument `accelerator_type` for `ray.train.ScalingConfig`.
+
+
+```python
+# CPU
+ScalingConfig(num_workers=4)
+
+# GPU
```
jjyao, line 253: @woshiyyya  Can you add an example of Intel GPU?

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,374 @@
...
+
+Internally, Ray Train adds the `accelerate_type` parameter in `ray.remote()` to start the training workers.
+
+#### 2. Specify PyTorch Backend for each Accelerator Family
+
+Different accelerator family uses different backend to launch deep learning training. Ray Train will implement a different backend class for each accelerator family to launch distributed groups and set environment variables.
+
+| Accelerator | Communication Backend | Developer APIs | User APIs | 
+| - | - | - | - |
+| GPU      | `nccl`, `gloo` | TorchBackend | `TorchConfig()` |
```
jjyao, line 279: @woshiyyya Can you add an example of Intel GPU

woshiyyya: Done.

reps/2023-10-13-accelerator-support.md
scv119: let's document the mapping in a table in this doc?
"GPU" -> Nvidia, Intel, AMD
"TPU" -> TPU
...

reps/2023-10-13-accelerator-support.md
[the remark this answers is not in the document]
scv119: if we use `num_gpus` and not specify accelerators, the train or other library might get a intel/amd gpu. How would  the worker detect that and work with different type of gpus?

reps/2023-10-13-accelerator-support.md
jjyao: Added.

reps/2023-10-13-accelerator-support.md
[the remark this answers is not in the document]
jjyao: So on each node you only have intel or amd gpus and we set the correct visible accelerator ids env var based on what the node has physically.

richardliaw:
@sriram-anyscale , would be great to get your thoughts on this as well.

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,399 @@
...
+| ...         | ...                       |
+
+`accelerator_type` option can also be used to specify a particular type of accelerator for a task or actor.
+
+```python
+@ray.remote(num_gpus=1)
+def task():
+    ...
+
+@ray.remote(num_gpus=1, accelerator_type="A100")
```
sriram-anyscale, line 52: How would say that you are OK with either A100 or A10G?

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,399 @@
...
+@scv119
+
+## Design
+
+### Ray Core
+
+#### API
+Users will use accelerators through Ray resources. For GPU accelerators (e.g. Nvidia, AMD, Intel GPUs), they will use the `GPU` resource name. For other accelerators, they will each have their own resource name (e.g. TPU accelerator has resource name `TPU`).
+
+| Resource    | Accelerator               |
```
sriram-anyscale, line 37: Is the distinction between GPU, TPU, HPU, etc. important? Can't we use a more generic term like "num_accelerators"? It also seems a bit inconsistent to have "num_gpus", but not "num_tpus", etc. - I'm guessing it is because you don't want too many of those "num_*" attributes.

You do mention that "num_accelerators" was an alternative you considered - but discarded it. The alternative you have chosen to special case GPUs with num_gpus but not for the others seems like it could do with a bit more cleanup.

reps/2023-10-13-accelerator-support.md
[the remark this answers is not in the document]
sriram-anyscale: I've asked a similar question to Chen below - essentially you can say you want a specific accelerator, or you don't care which accelerator you get. But you cannot say that you only want one of two accelerators (as in my example in the other comment - "I am OK with either of A100 or A10G").

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,399 @@
...
+    ...
+
+@ray.remote(num_tpus=1, accelerator_type="TPU-V4")
+def task():
+    ...
+```
+
+A single `ACCELERATOR` resource for the ability to schedule a task on any accelerator
+and an `accelerator_family` is also introduced to allow specifying the family of the accelerators to use.
+People feel this is less obvious or expclit than having a reosurce name per accelerator family
```
sriram-anyscale, line 88: typo (reosurce)

reps/2023-10-13-accelerator-support.md
sriram-anyscale: The H100 datasheet says this:

Note that the H100 GPUs are primarily built for executing
datacenter and edge compute workloads for AI, HPC, and data analytics, but not graphics
processing.

So blurs the distinction between GPU and other AI processors.

reps/2023-10-13-accelerator-support.md
jjyao: accelerator_type is just labels/custom resources that Ray adds for you automatically. For your case, you can just manually create your own label/custom resource like `A100_OR_A10G` and request it in your task.

reps/2023-10-13-accelerator-support.md
jjyao: We try to have the resource name match how these accelerators are being called by people.

> Is the distinction between GPU, TPU, HPU, etc. important?

Different accelerators are quite different: e.g. some supports fractional resource, some not. Also the user's code normally only works for one accelerator family since different accelerator family provides different programming model (e.g. Nvidia GPU uses cuda).
This means if we do `num_accelerators`, we also need to introduce an `accelerator_family` node label so uses can express the equivalent of `num_gpus=1` with `num_accelerators=1, accelerator_family="GPU"`

reps/2023-10-13-accelerator-support.md
jjyao: Later on, once we switch to label you can say `"accelerator_type": In("A100", "A10G")`

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,399 @@
...
+
+#### API
+Users will use accelerators through Ray resources. For GPU accelerators (e.g. Nvidia, AMD, Intel GPUs), they will use the `GPU` resource name. For other accelerators, they will each have their own resource name (e.g. TPU accelerator has resource name `TPU`).
+
+| Resource    | Accelerator               |
+| ----------- | ------------------------- |
+| GPU         | Nvidia, Intel, AMD, Apple |
+| TPU         | TPU                       |
+| HPU         | Habana Gaudi              |
+| neuron_cores| AWS Trainium, Inferentia  |
```
ArturNiederfahrenhorst, line 42: Can someone ELI5 what the difference between TPU, GPU and neuron_cores is? And what other _types_ of accelerators might be part of the triple dots?

My angle is, if these all do matrix multiplications, from a user perspective, why would I need to change my codebase to use another of these resources?

jjyao: Different accelerators have different programming models. E.g. Nvidia GPU uses cuda and TPU uses XLA so normally you need to change your codebase when you switch from one type of accelerator to another.

reps/2023-10-13-accelerator-support.md
rkooo567: Maybe we can allow syntactic sugar like accelerator_type="A100,A10"? Feel like this is more pythonic API

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,399 @@
...
+```
+
+### Kuberay
+
+To auto populate `rayStartParams` resources using pod resources, some code is needed.
+See https://github.com/ray-project/kuberay/blob/b7bc7ae8c983160bb700d6b8bb4014dda183ecea/ray-operator/controllers/ray/common/pod.go#L687 as an example of how GPU resource is populated.
+
+### Observability
+
+Currently we emit metrics about the physical usages of GPUs and also show them on Ray dashboard.
```
rkooo567, line 382: Can we make this as a part of `AcceleratorManager` class? We also need ways to emit metrics. 

We need: acclerator (TPU vs GPU), acclerator_type (A100, H100), utilization & memory. 

1. Dashboard cluster view should display live metrics.
2. Dashboard should emit metrics with device type (with different labels). 
3. ray status should correctly display devices.

reps/2023-10-13-accelerator-support.md
akshay-anyscale: there's also the proposed placement group alternatives feature which may be a better way of solving that

reps/2023-10-13-accelerator-support.md
jjyao: Yes, it will. I'm trying some prototype here: https://github.com/ray-project/ray/pull/40749.

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,399 @@
...
+  def get_resource_name() -> str:
+      """Get the name of the resource representing this accelerator family.
+
+      Returns:
+          The resource name: e.g., the resource name for Nvidia GPUs is "GPU"
+      """
+
+  @staticmethod
+  @abstractmethod
+  def get_visible_accelerator_ids_env_var() -> str:
```
ericl, line 159: Do we still need this given that get_current_process_visible_accelerator_ids and set_current_process_visible_accelerator_ids are available?

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,399 @@
...
+    scaling_config=ScalingConfig(
+        num_workers=4, 
+        resources_per_worker={"HPU": 1}
+    ),
+    torch_config=TorchConfig(backend="hccl")
+)
+```
+
+
+### Ray RLLib
```
ericl, line 348:
```suggestion
### Ray RLlib
```

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,399 @@
...
+Currently there is no plan to support other accelerators in RLLib.
```
ericl, line 349:
```suggestion
Currently there is no plan to support other accelerators in RLlib.
```

reps/2023-10-13-accelerator-support.md
```diff
@@ -0,0 +1,399 @@
...
+  @abstractmethod
+  def set_current_process_visible_accelerator_ids(ids: List[str]) -> None:
+      """Set the ids of accelerators of this family that are visible to the current process.
+
+      Args:
+          ids: The ids of visible accelerators of this family.
+      """
+
+  @staticmethod
+  def get_ec2_instance_num_accelerators(
```
ericl, line 230: Can we make this cloud agnostic?

reps/2023-10-13-accelerator-support.md
jjyao: Currently it's only used in some error messages

```
raise ValueError(
                    f"Attempting to start raylet with {num_accelerators} "
                    f"{accelerator_resource_name}, "
                    f"but {accelerator_manager.get_visible_accelerator_ids_env_var()} "
                    f"contains {visible_accelerator_ids}."
                )
```

reps/2023-10-13-accelerator-support.md
jjyao: Not really since this is used by autoscaler to auto fill the resources for each node type before the node is started so for different clouds, there are different ways (mostly just querying metadata for an instance type) to get those information.

ericl (commented):
LGTM

reps/2023-10-13-accelerator-support.md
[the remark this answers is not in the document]
anyscalesam: how would this work with mixed fleets? eg: a ray cluster with both nvidia, amd, and intel custom silicon?

reps/2023-10-13-accelerator-support.md
anyscalesam: what was the solution/conclusion that we landed on?

reps/2023-10-13-accelerator-support.md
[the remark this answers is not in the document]
jjyao: It should just work.

PR #48, opened 2023-12-04
RFC: Accelerated DAGs

stephanie-wang:


reps/2023-12-04-accelerated-dag.md
```diff
@@ -0,0 +1,360 @@
...
+workers = [Worker.remote() for worker in workers]
+# A task called on the CollectiveGroup will be gang-scheduled on each worker.
+pool = CollectiveGroup(workers)
+outputs = pool.allreduce.bind(inputs)
+```
+
+Workloads
+=========
+
+Possible validation workloads fall under two different categories:
```
ericl, line 273: It would be great to show the code for a few example workloads, for example, the scatter-gather vLLM example, and perhaps something more complex like Gpipe parallelism.

reps/2023-12-04-accelerated-dag.md
```diff
@@ -0,0 +1,360 @@
...
+
+*Vision for Ray Train in Ray 3.0: Ray Train is built on Ray's single-controller model, but with much finer-grained tasks and more control over the execution. The physical execution is as fast as SPMD.*
+
+The "accelerated DAGs" effort aims to keep Ray's flexibility and fault
+tolerance, but: (a) reduce control plane overheads, and (b) support
+specialized communication transports. The basic ideas are to: (a) reuse
+control plane decisions from past executions, (b) support
+application-defined transports such as
+[NCCL](https://developer.nvidia.com/nccl) and
+[UCX](https://openucx.org/). Ultimately, the goal is to
```
ericl, line 112: I think it's also worth calling out `libfabric` explicitly here.

reps/2023-12-04-accelerated-dag.md
```diff
@@ -0,0 +1,360 @@
...
+    pass
+  def recv(self, recv_meta: BufferMetadata) -> Buffer:
+    """Ideally async. Called by the receiver."""
+    pass
+```
+
+When an application-defined transport is used, we will use the default
+transport to synchronize between the sender and receiver, but we only
+send the BufferMetadata in along the default transport. This will act as
+a signal to begin the actual \`Transport.send\` and \`Transport.recv\`
```
rkooo567, line 247: Isn't this signal very high overhead (basically we will have default transport + custom transport overhead)? Is it necessary if the transport happens under the channel (as an output of each DAG node)?

ericl:
> Isn't this signal very high overhead (basically we will have default
transport + custom transport overhead)? Is it necessary if the transport
happens under the channel (as an output of each DAG node)?

Good point, especially since custom TCP/RDMA transports primarily improve
latency and not throughput.

stephanie-wang:
Another TODO: Add some more detail on interaction with load-balancing and autoscaling, ideally some psuedocode.

reps/2023-12-04-accelerated-dag.md
```diff
@@ -0,0 +1,360 @@
...
+
+
+### Advanced patterns
+|                                                                 | Key properties / requirements                                        | Goals                                                            |
+|-----------------------------------------------------------------|----------------------------------------------------------------------|------------------------------------------------------------------|
+| [Pipedream](https://arxiv.org/pdf/1806.03377.pdf) style pipeline-parallel distributed training (PP)     | Iterative P2P GPU communication                                      | Performance parity; Increase flexibility of partitioning scheme   |
+| vLLM pipeline parallelism on heterogeneous GPUs                 | Asymmetric compute                                                   | Reduce implementation burden                                     |
+| Fault-tolerant distributed serving                              | Resume execution w/o restarting everyone                             | Reduce downtime via greater recovery flexibility                 |
```
edoakes, line 331: As I understand the fault tolerance section above, failure handling will be done at the application layer by propagating the error back to the driver. In this case it seems that the entire DAG would need to be re-run. So how would the proposal help address this requirement?

It's also a little unclear to me how important this requirement is for most serving applications. Most often latencies are low (~seconds at most) and failures are rare, so needing to re-run the full inference is not a big problem.

reps/2023-12-04-accelerated-dag.md
```diff
@@ -0,0 +1,360 @@
...
+validation, and will select some advanced patterns as we are closer to
+release. Feedback on workload selection is welcome! The following
+categories are in rough order of technical complexity:
+
+### Common patterns: Microbenchmarks (simple distributed inference)
+
+|                                                                 | Key properties / requirements                                        | Goals                                                            |
+|-----------------------------------------------------------------|----------------------------------------------------------------------|------------------------------------------------------------------|
+| vLLM tensor parallelism                                         |                                                                      | Reduce Ray overheads                                             |
+| vLLM pipeline parallelism                                       | P2P/cross-node GPU communication                                     | Reduce (expected) Ray overheads; Validate cross-node performance  |
```
edoakes, line 314: My understanding from vLLM folks' benchmarks was that cross-node GPU communication overhead is too high (even if using optimized transports) so for now they don't have a desire to use it. I could be wrong though, this is hearsay.

edoakes (commented):
Overall the motivation makes a lot of sense. Right using the Ray API as the dataplane for ML workloads is a performance loss over other libraries and we should close that gap.

From a serving perspective, there are basically four categories of communication that happens (excluding the pattern used by vLLM where you use multiple GPUs for a single logical replica):

**(1)** Point-to-point communication passing small messages between two groups of actors (often colocated on the same node). This is the pattern for HTTP proxy -> replica communication and many user-defined applications doing basic model composition.

Requirements:

- Backpressure, load balancing (including adding/removing actors), pipelining.

Non-requirements:

- Dynamic control flow
- Passing output to a downstream actor.

**(2)** An extension of (1) where there is composition across multiple groups of actors (for example A calls B and passes the output to C). In some cases the output of (B) might be large or on a GPU, so passing it by reference may be helpful.

Requirements:

- Backpressure, load balancing (including adding/removing actors), pipelining.
- Passing outputs to a downstream actor.

Non-requirements:

- Dynamic control flow

**(3)** An extension of (1) and (2) where there is also dynamic control flow (i.e., users choose the downstream group of actors based on some input or intermediate output).

Requirements:

- Backpressure, load balancing (including adding/removing actors), pipelining.
- Passing outputs to a downstream actor.
- Dynamic control flow.

To me, it's clear that the faster transport layer could be beneficial to all of the above, particularly (1) where we mostly want fast message passing between two actors on the same node.

It's unclear how the optimized DAG as proposed could satisfy the requirements around load balancing and autoscaling across groups of actors. Would be interested to hear if you have any ideas there.

ericl:
> Point-to-point communication passing small messages between two groups of actors

Would the current proposal be viable for even this? As I understand this would require dynamic M:N communication, whereas the compiled dag is assuming a fully static communication topology between actors.

For example, if group 1 is a single actor dispatching work to group 2 actors, that would work (we could have multiple DAGs capturing "1->2" communication), but if group 1 had multiple dispatching actors then that would break the assumption that each group 2 actor is part of a single DAG.

rkooo567:
I have impression the custom/faster transport has to be decoupled from DAG. It is a requirement for DAG, but DAG shouldn't be a requirement to use custom/faster transport. I think it could be easily implemented with existing ray API if we just establish the channel when the first communication is initiated. And if it works with existing Ray API, we don't have to support ^ from core?

edoakes:
Yes you're correct @ericl. To optimize the basic case of M:N communication we essentially just want the fast transport layer. Whether we get that through the DAG abstraction (single-node DAG) or use the lower-level primitive directly doesn't really matter too much.

reps/2023-12-04-accelerated-dag.md
stephanie-wang: For tensor parallelism, yes, but I don't think that's the case for pipeline parallelism.

reps/2023-12-04-accelerated-dag.md
```diff
@@ -0,0 +1,360 @@
...
+allow us to keep current Ray Core system optimizations such as:
+
+1.  Overlapping communication with task execution.
+2.  Handle communication failures gracefully through retries.
+3.  Control communication concurrency / resource usage.
+
+Accelerated DAGs will ship with a default transport for objects passed
+along the DAG's edges. This will use shared memory if the
+sender/receiver are colocated, and shared memory/Ray Core for
+cross-node.
```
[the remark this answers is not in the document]
stephanie-wang: Yes, but not for a while probably. The initial focus is GPU communication.

reps/2023-12-04-accelerated-dag.md
stephanie-wang: Restarting here meaning the worker processes, not individual tasks.

reps/2023-12-04-accelerated-dag.md
```diff
@@ -0,0 +1,637 @@
...
+        return inp, gpu_tensors
+
+@ray.remote
+class Consumer:
+    def consume_all_sync(self, gpu_tensors):
+        pass
+
+with InputNode() as inp:
+    with InputGPUChannel() as gpu_channel:
+        _, tensors = producer.produce_all_sync.bind(inp.input)
```
ericl, line 481: This syntax is pretty strange, I'd expect something more like `tensors.set_device("gpu")` or `tensors.set_channel_impl(GPUChannel)`.

stephanie-wang: Hmm, `tensors.set_channel_impl` seems good. Mainly I want there to be a way to pass a specific Channel or Channel type, so `tensors.set_device("gpu")` seems less ideal.

zhe-thoughts (approved):
Vote on ray-committers just passed

PR #49, opened 2023-12-17
[REP] Virtual Cluster

jjyao:
https://docs.google.com/document/d/1TJ3jHWVGGviJOQYYlB9yNjUUG8Hn54hYsNqnUrY4L14/edit#heading=h.8f25wuyjzzer


reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,35 @@
+# Virtual Cluster
+
+## Summary
+
+Ray currently has the [physical cluster](https://docs.ray.io/en/releases-2.9.0/cluster/getting-started.html) concept and we proposes to add a new virtual cluster concept. A virtual cluster is a partition of the physical cluster and can be dynamically scaled at runtime. A physical cluster can be partitioned into multiple virtual clusters and each virtual cluster runs a single Ray job. Through this way, multiple jobs can share the cluster resources with isolation. Virtual cluster is a fundamental building block for multi-tenancy Ray.
```
ericl, line 5:
```suggestion
Ray currently has the [physical cluster](https://docs.ray.io/en/releases-2.9.0/cluster/getting-started.html) concept and we proposes to add a new virtual cluster concept. A virtual cluster is a partition of the physical cluster and can be dynamically scaled at runtime. A physical cluster can be partitioned into multiple virtual clusters and each virtual cluster runs a single Ray job. Through this way, multiple jobs can share the cluster resources with isolation. Virtual cluster is a fundamental building block for multi-tenant Ray.
```

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,35 @@
...
+
+<img src="https://user-images.githubusercontent.com/898023/291094699-35bac047-5844-4f2c-a794-17cd18e96219.png" alt="drawing" width="726"/>
+
+### General Motivation
+
+While users can run multiple Ray jobs within a single Ray cluster simultaneously, there is currently no sufficient isolation between jobs. Users also have no way to specify cross-job policies such as fairness. For example, if you run two Ray Tune jobs on the same cluster, they will both try to use all the cluster resources and compete for cluster resources in an ad-hoc mechanism, without notion of isolation, fairness, or priority. In order to properly support multi-tenancy, we need a machnism to share cluster resources between different jobs with isolation and certain cross-job policies. While [placement group](https://docs.ray.io/en/releases-2.9.0/ray-core/scheduling/placement-group.html) can solve some of the issues, its lack of nesting and autoscaling support makes it unusable for certain workloads. Virtual cluster is the mechanism we propose here and it supports isolation, nesting and autoscaling.
```
ericl, line 11:
```suggestion
While users can run multiple Ray jobs within a single Ray cluster simultaneously, there is currently no sufficient isolation between jobs. Users also have no way to specify cross-job policies such as fairness. For example, if you run two Ray Tune jobs on the same cluster, they will both try to use all the cluster resources and compete for cluster resources in an ad-hoc mechanism, without notion of isolation, fairness, or priority. In order to properly support multi-tenancy, we need a machnism to share cluster resources between different jobs with isolation and certain cross-job policies. While [placement group](https://docs.ray.io/en/releases-2.9.0/ray-core/scheduling/placement-group.html) can solve some of the issues, its lack of nesting and autoscaling support makes it unusable for certain workloads. Virtual cluster is the mechanism we propose here and it supports isolation, nesting, and autoscaling.
```

ericl, line 11: Could also add an example, "For instance, consider two nested Train jobs launched within a Tune grid sweep job, each of which is logically running on a "cluster" of its own."

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,398 @@
...
+    def max_resources()  # total after autoscaling to max, if known
+    def available_resources()  # like ray.available_resources()
+    def used_object_store_memory() # Object store memory used by objects in this virtual cluster
+    def nodes()  # like ray.nodes()
+    def state()  # like ray.util.state
+    def child_clusters() -> List[VirtualClusterInfo]
+    def parent_cluster() -> Optional[VirtualClusterInfo]
+
+class VirtualNodeInfo:
+    def node_id() -> str
```
stephanie-wang, line 248: Specify if this is virtual or physical node ID.

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,398 @@
...
+    def node_labels() -> Dict[str, str]
+    def used_resources() -> Dict[str, double]
+    def avail_resources() -> Dict[str, double]
+    def total_resources() -> Dict[str, double]
+    def parent_node() -> Optional[VirtualNodeInfo]
```
stephanie-wang, line 253: I have a feeling it's a bit less weird to just expose `.physical_node_id()`.

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,398 @@
...
+
+### Shepherd of the Proposal
+
+@ericl
+
+## Design
+
+With the introduction of virtual clusters, every Ray job runs in its own virtual cluster and only has access to resources inside that virtual cluster. Each virtual cluster has a spec that defines the min and max resources of the cluster. Min resources are minimal resources required for the job to run and they are atomically reserved for gang scheduling. If min resources cannot be reserved when there are not enough available resources, the job will be queued. With job queueing, we can implement different policies such as FIFO or priority-based queueing. Max resources are the autoscaling limit of the virtual cluster and the maximal resources can be used by the job.
+
+Virtual clusters can be nested and a Ray job can create sub-clusters to isolate separate parts of its application workload. For example, a Tune grid sweep job can create a sub-cluster for each of its nested Train workload. These possibly nested virtual clusters form a tree where the root is the entire physical cluster.
```
stephanie-wang, line 31: I think it's a bit strange that the root is a physical cluster. Shouldn't it be another virtual cluster?

jjyao: You can think of it's a virtual cluster that represents the entire physical cluster: each virtual node represents a physical node.

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,398 @@
...
+
+  // If specified, ensure we have at least this min amount
+  // of resources before starting the cluster.
+  // If not specified, the default value is 0.
+  map<string, double> flexible_resource_min
+
+  // If specified, limit the consumption of these resources to
+  // the specified values.
+  // If not specified, the default value is infinite.
+  map<string, double> flexible_resource_max
```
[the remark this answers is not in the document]
jjyao: I think the strategy is within flexible resources: Ray will create as few flexible virtual nodes as possible for the given amount of flexible resources.

Do we really need a strategy between flexible nodes and fixed size nodes?

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,398 @@
...
+  // Additional labels that the
+  // virtual node has in addition to
+  // those inherited from the parent node.
+  map<string, string> labels
+}
+
+enum SchedulingPolicy {
+  PACK
+  SPREAD
+  STRICT_SPREAD
```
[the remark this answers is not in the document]
jjyao: This won't. This only supports pg strategies.

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,452 @@
...
+  // Flexible resources are stored in flexible virtual nodes
+  // and Ray will use as few flexible virtual nodes as possible
+  // for the given amount of flexible resources to minimize fragmentation.
+  // In other words, Ray prefers to scale up a flexible virtual node
+  // than adding a new flexible virtual node.
+
+  // If specified, ensure we have at least this min amount
+  // of resources before starting the cluster.
+  // If not specified, the default value is 0.
+  map<string, double> flexible_resource_min
```
ericl, line 57:
```suggestion
  map<string, double> flex_resource_min
```

nit: "flexible" is too long/hard to spell, prefer to abbreviate in these cases

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,452 @@
...
+actors = []
+for i in range(2):
+  actors.append(Actor.options(num_gpus=1, node_labels={"bundle_index": In(str(i))}).remote())
+```
+
+##### Tune + Dataset
+
+```python
+# Tune can run 2 trails (each trail runs two 1 GPU trainers) in parallel and Dataset can use 1-10 CPUs
+ray.init(virtual_cluster=VirtualCluster(fixed_size_nodes=[FixedSizeNodes(nodes=[FixedSizeNode(resources={"GPU": 1}), FixedSizeNode(resources={"GPU": 1})], scheduling_policy=PACK), FixedSizeNodes(nodes=[FixedSizeNode(resources={"GPU": 1}), FixedSizeNode(resources={"GPU": 1})], scheduling_policy=PACK)], flexible_resource_min={"CPU": 1}, flexible_resource_max={"CPU": 10}))
```
ericl, line 226: nit: this is hard to read in the rendered markdown due to the line wrapping

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,452 @@
...
+
+Virtual nodes of a virtual cluster can be either fixed size or flexible/resizable. For a single virtual cluster, there can be multiple fixed-size virtual nodes but at most one flexible virtual node on a single physical node. The upscaling of a virtual cluster can be achieved by adding new virtual nodes or scaling up an existing flexible virtual node.
+
+### API
+
+#### Virtual Cluster Spec
+
+```
+message VirtualCluster {
+  // A virtual cluster consits of flexible resources and fixed size resources.
```
ericl, line 45: Should add `map<string, string> labels` that applies to both flex and fixed nodes.

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,452 @@
...
+```
+
+```python
+# Create a placement group with two bundles that are strict spreaded
+pg = placement_group([{"GPU": 4}, {"GPU": 1}], strategy="STRICT_SPREAD")
+# Run the actor using the bundle 0 resources
+Actor.options(scheduling_strategy=PlacementGroupSchedulingStrategy(placement_group=pg, placement_group_bundle_index=0)).remote()
+
+# Create a virtual cluster with two virtual nodes that are strict spreaded
+vc = VirtualCluster(fixed_size_nodes=[FixedSizeNodes(nodes=[FixedSizeNode(resources={"GPU": 4}, labels={"bundle_index": "0"}), FixedSizeNode(resources={"GPU": 1}, labels={"bundle_index": "1"})], scheduling_policy=STRICT_SPREAD)])
```
ericl, line 189: Nice, I like how we can get rid of the explicit bundle indexing in favor of labels.

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,452 @@
...
+  remote_nodes: {"raylet2": [Node(total_resources={"CPU": 2}, available_resources={"CPU": 1}, labels={"ray.io/vnode_id": "raylet2", "ray.io/vcluster_id": "physical_cluster_id"}), Node(total_resources={"CPU": 1}, available_resources={"CPU": 0}, labels={"ray.io/vnode_id": "vnode2", "ray.io/vcluster_id": "vcluster1"})]}
+
+Raylet2:
+  local_nodes: [Node(total_resources={"CPU": 2}, available_resources={"CPU": 1}, labels={"ray.io/vnode_id": "raylet2", "ray.io/vcluster_id": "physical_cluster_id"}), Node(total_resources={"CPU": 1}, available_resources={"CPU": 0}, labels={"ray.io/vnode_id": "vnode2", "ray.io/vcluster_id": "vcluster1"})]
+  remote_nodes: {"raylet1": [Node(total_resources={"CPU": 2}, available_resources={"CPU": 0}, labels={"ray.io/vnode_id": "raylet1", "ray.io/vcluster_id": "physical_cluster_id"}), Node(total_resources={"CPU": 2}, available_resources={"CPU": 0}, labels={"ray.io/vnode_id": "vnode1", "ray.io/vcluster_id": "vcluster1"})]}
+```
+
+Eventually when the job finishes, newly added nodes will be idle terminated.
```
ericl, line 446: Can you also add a subsection on handling the fairness/starvation issue @stephanie-wang brought up?

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,452 @@
...
+ray.init(virtual_cluster=VirtualCluster(fixed_size_nodes=[FixedSizeNodes(nodes=[FixedSizeNode(resources={"GPU": 1}, labels={"bundle_index": "0"}), FixedSizeNode(resources={"GPU": 1}, labels={"bundle_index": "1"})], scheduling_policy=STRICT_SPREAD)]))
+actors = []
+for i in range(2):
+  actors.append(Actor.options(num_gpus=1, node_labels={"bundle_index": In(str(i))}).remote())
+```
+
+##### Tune + Dataset
+
+```python
+# Tune can run 2 trails (each trail runs two 1 GPU trainers) in parallel and Dataset can use 1-10 CPUs
```
ericl, line 225:
```suggestion
# Tune can run 2 trials (each trial runs two 1 GPU trainers) in parallel and Dataset can use 1-10 CPUs
```

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,452 @@
...
+```python
+# Default virtual cluster
+# The job can use up to the entire physical cluster resources.
+ray.init()
+
+# The job can use at least 1 CPU and at most 8 CPUs.
+ray.init(virtual_cluster=VirtualCluster(flexible_resource_min={"CPU": 1}, flexible_resource_max={"CPU": 8}))
+
+# The job uses 1 A100 GPU.
+ray.init(virtual_cluster=VirtualCluster(fixed_size_nodes=[FixedSizeNode(resources={"GPU": 1}, parent_node_selector={"accelerator_type": In("A100")})]))
```
scv119, line 150: what happens if there is no such node exactly matches the fixed size node resource? will we round up, shard the node, or fail?

reps/2023-12-14-virtual-cluster.md
scv119: should we just name it total_resource_min, total_resource_max?

reps/2023-12-14-virtual-cluster.md
scv119: a follow up question is what happens if the required_nodes conflicts with flex_resource_max?

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,452 @@
...
+  // If not specified, the default value is infinite.
+  map<string, double> flexible_resource_max
+
+  // == Fixed size resources ==
+  // Fixed sized virtual nodes to request, e.g. {"GPU": 4}.
+  // These virtual node resources are part of the min resources
+  // together with flexible_resource_min
+  // that will be atomically reserved when the
+  // virtual cluster is created.
+  repeated FixedSizeNodes fixed_size_nodes
```
scv119, line 70: would min_nodes be a better name?

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,452 @@
...
+
+<img src="https://user-images.githubusercontent.com/898023/291710281-41d5b172-95ae-4134-8fab-a6da6c08701a.png" alt="drawing" width="729"/>
+
+### General Motivation
+
+While users can run multiple Ray jobs within a single Ray cluster simultaneously, there is currently no sufficient isolation between jobs. Users also have no way to specify cross-job policies such as fairness. For example, if you run two Ray Tune jobs on the same cluster, they will both try to use all the cluster resources and compete for cluster resources in an ad-hoc mechanism, without notion of isolation, fairness, or priority. In order to properly support multi-tenancy, we need a machnism to share cluster resources between different jobs with isolation and certain cross-job policies. While [placement group](https://docs.ray.io/en/releases-2.9.0/ray-core/scheduling/placement-group.html) can solve some of the issues, its lack of nesting and autoscaling support makes it unusable for certain workloads. Virtual cluster is the mechanism we propose here and it supports isolation, nesting, and autoscaling.
```
zhe-thoughts, line 11: We should add a concrete use case to explain why nesting is useful.

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,452 @@
+# Virtual Cluster
+
+## Summary
+
+In addition to the existing [physical cluster](https://docs.ray.io/en/releases-2.9.0/cluster/getting-started.html) and physical node concepts, we propose to add new virtual cluster and virtual node concepts. A virtual node is a partition of a single physical node and it has resources and node labels just like the physical node. A virtual cluster made up of virtual nodes is a partition of the physical cluster and can be dynamically scaled at runtime. A physical cluster can be partitioned into multiple virtual clusters and each virtual cluster runs a single Ray job. Through this way, multiple jobs can share a physical cluster with logical resources isolation. Virtual cluster is a fundamental building block for multi-tenant Ray.
```
zhe-thoughts, line 5: Do we plan to support different "environments" for different virtual clusters within the same physical cluster? E.g. different Python versions?

reps/2023-12-14-virtual-cluster.md
jjyao: > should we just name it min_resources, max_resources?

The semantics is different. `flex_resource_min` doesn't include fixed node resources while `min_resources` should include everything.

reps/2023-12-14-virtual-cluster.md
jjyao: What's the use case for adding labels to flex nodes?

reps/2023-12-14-virtual-cluster.md
jjyao: A physical node can be split into multiple fixed size virtual nodes.

> a follow up question is what happens if the required_nodes conflicts with flex_resource_max?

flex_resource_max doesn't include fixed size node resources. Total virtual cluster resource is flex resource + fixed resource.

reps/2023-12-14-virtual-cluster.md
ericl: You need labels in order to schedule tasks within an specific virtual cluster in the first place, right? This is also for consistency with the physical node interface.

reps/2023-12-14-virtual-cluster.md
jjyao: This is out of scope for virtual cluster but potentially in scope for multi-tenant Ray.

reps/2023-12-14-virtual-cluster.md
jjyao: Yea, flex nodes also have system labels (e.g. vc_id). I'm just not sure whether we need to expose a way for users to set custom labels for flex nodes. 

So you are suggesting we have `map<string, string> cluster_level_custom_labels` that applies to all virtual nodes of a virtual cluster. And later on fixed nodes can also set additional custom labels individually?

Or we have `map<string, string> flex_nodes_custom_labels` that only applies to flex nodes.

ericl: I see, if we have automatically added labels for vc_id then I agree this is more of a P1.

reps/2023-12-14-virtual-cluster.md
scv119: would flex_resource_max -> additional_resource make sense?

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,807 @@
...
+
+<img src="https://user-images.githubusercontent.com/898023/291710281-41d5b172-95ae-4134-8fab-a6da6c08701a.png" alt="drawing" width="729"/>
+
+### General Motivation
+
+While users can run multiple Ray jobs within a single Ray cluster simultaneously, there is currently no sufficient isolation between jobs. Users also have no way to specify cross-job policies such as fairness. For example, if you run two Ray Tune jobs on the same cluster, they will both try to use all the cluster resources and compete for cluster resources in an ad-hoc mechanism, without notion of isolation, fairness, or priority. In order to properly support multi-tenancy, we need a machnism to share cluster resources between different jobs with isolation and certain cross-job policies. While [placement group](https://docs.ray.io/en/releases-2.9.0/ray-core/scheduling/placement-group.html) can solve some of the issues, its lack of nesting and autoscaling support makes it unusable for certain workloads. Virtual cluster is the mechanism we propose here and it supports (logical resources) isolation, nesting, and autoscaling.
```
kadisi, line 11: Is it enough for a Virtual cluster to support only logical isolation? Whether it makes more sense for virtual cluster to support real resource isolation？

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,807 @@
...
+  # Run the actor using the first virtual node resources
+  Actor.options(node_labels={"bundle_index": In("0")}).remote()
+```
+
+Here we have two options for the API:
+
+1. Deprecate the placement group API and use virtual cluster API directly.
+2. Keep the placement group API and only change the internal implementation to use virtual cluster.
+
+#### Examples
```
justinvyu, line 228: Another example to consider:

#### ML platform with priority-based pre-emption of distributed training workers

```
Overall Ray Cluster: 16 GPUs, 128 CPUs
    Job 0 (P0): min_gpus=4, max_gpus=10, cpus=32
        Worker * 10: each with 1 GPU
    Job 1 (P1): min_gpus=2, max_gpus=8, cpus=1
        Worker * 3: each with 1 GPU
    Job 2 (P1): min_gpus=1, max_gpus=4, cpus=1
        Worker * 3: each with 1 GPU
    Job 3 (P2): min_gpus=1
        Pending for something to free up. 
```

It might make more sense to switch these priorities to GPU time in %. This arbitrary upscale/downscale depends on elastic distributed training from Ray Train / another library.

This might be a little too much scope for the current REP, but I would like a future REP to enable this kind of pre-emption policy for each "parent" virtual cluster. The parent VirtualCluster should be able to drain some nodes of the child clusters to reclaim resources, as a form of "hard downscaling", rather than just waiting for task completion.

justinvyu (commented):
🚀

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,807 @@
...
+        FixedSizeNodes(
+            nodes=[
+                FixedSizeNode(resources={"GPU": 1}),
+                FixedSizeNode(resources={"GPU": 1})
+            ],
+            scheduling_policy=PACK
+        )
+    ],
+    flex_resource_min={"CPU": 1},
+    flex_resource_max={"CPU": 10}))
```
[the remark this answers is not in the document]
ericl: Trial 1: {fixed_nodes={2 GPU nodes}, flex={2..20 CPUs}}
Trial 2: {fixed_nodes={2 GPU nodes}, flex={2..20 CPUs}}

For the SplitCoordinator, I'm assuming you mean there is some separate preprocessing job that's run here before kicking off the training trials? In that case, you can put that in a another VC with {flex={1-10 CPU, 1-4 GPU}}.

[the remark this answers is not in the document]
justinvyu: The `SplitCoordinator` I'm mentioning is the thing executing the ray data + ray train streaming execution that gets run in the background during the training execution.

https://github.com/ray-project/ray/blob/master/python/ray/data/_internal/iterator/stream_split_iterator.py#L47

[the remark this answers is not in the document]
ericl: I see, then those should be part of the trial cluster, so this would be a reasonable config:

Trial 1: {fixed_nodes={2 GPU nodes}, flex={2..20 CPUs}}
Trial 2: {fixed_nodes={2 GPU nodes}, flex={2..20 CPUs}}

For more predictable performance you'd probably also want to assign a fixed number of CPUs instead of allowing 2..20 for flex, so probably `{fixed_nodes={2 GPU nodes}, flex={10 CPUs}}` would be the way to go.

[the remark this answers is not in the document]
ericl:
```
The Ray Data components are currently detached from the Ray Train/Tune placement group. How does [placement_group_capture_child_tasks=None](https://docs.ray.io/en/latest/ray-core/scheduling/placement-group.html#advanced-child-tasks-and-actors) translate to VirtualClusters?
```

To answer this question directly, by design we don't need to support detached resources for virtual clusters. The data tasks can use flex resources instead.

larrylian:
@jjyao  
The two prerequisite tasks that need to be completed are:
1. Move the NodeLabelSchedulingStrategy to node_label.
2. Support dynamic Node Label.
Should we consider starting the design for these now?

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,807 @@
...
+    fixed_size_nodes=[
+        FixedSizeNodes(
+            nodes=[
+                FixedSizeNode(resources={"GPU": 4}, labels={"bundle_index": "0"}),
+                FixedSizeNode(resources={"GPU": 1}, labels={"bundle_index": "1"})
+            ],
+            scheduling_policy=STRICT_SPREAD
+        )
+    ])
+with vc:
```
larrylian, line 218: For nested virtualcluster case:
1. Actor 1 create vc_1 , Actor 2 want to create actor_3 in vc_1. This case actor 2 don't have vc_1 handle.  so can't use `with vc:`
So  i prefer virtual cluster hase name , options has a  `virtual_cluster` param.
```
Actor.options(virtual_cluster="vc_1").rmeote()
```

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,807 @@
...
+  // Flexible resources are stored in flexible virtual nodes
+  // and Ray will use as few flexible virtual nodes as possible
+  // for the given amount of flexible resources to minimize fragmentation.
+  // In other words, Ray prefers to scale up a flexible virtual node
+  // than adding a new flexible virtual node.
+
+  // If specified, ensure we have at least this min amount
+  // of resources before starting the cluster.
+  // If not specified, the default value is 0.
+  map<string, double> flex_resource_min
```
MissiontoMars, line 57: Can virtual nodes use custom resources on physical nodes in addition to CPU/GPU resources?

larrylian:
Hi, thanks. I will reply to you soon when I'm back.

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,807 @@
...
+    flex_resource_min={"CPU": 1}, flex_resource_max={"CPU": 8}))
+
+# The job uses 1 A100 GPU.
+ray.init(virtual_cluster=VirtualCluster(
+    fixed_size_nodes=[
+        FixedSizeNode(
+            resources={"GPU": 1},
+            parent_node_selector={"accelerator_type": In("A100")}
+        )
+    ]))
```
MissiontoMars, line 157: Earlier, job queuing was mentioned, FIFO/priority-based queueing, which means that multiple jobs can run simultaneously on a virtual cluster. In this case, how can multiple ray.init calls specify the use of the same virtual cluster?

reps/2023-12-14-virtual-cluster.md
```diff
@@ -0,0 +1,807 @@
...
+
+<img src="https://user-images.githubusercontent.com/898023/291139618-0be11470-db09-466d-8c2c-37b9e5b3765c.png" alt="drawing" width="508"/>
+
+Virtual clusters with different min and max resources are autoscalable. When scaling up, virtual clusters will try to borrow more resources from their parent clusters. If their parents have available resources, the scaling up is instant. Otherwise, parents will try to borrow resources from their parents recursively and eventually this may cause the upscale of the physical cluster. When scaling down, virtual clusters will return resources to their parents recursively and eventually this may cause the downscale of the physical cluster. To ensure fairness and avoid starvation, the borrowed resources are associated with leases. Once leases expire and the tasks or actors that are currently using these resources finish, the borrowed resources will be returned back to their parent clusters so that parent clusters can lend these resources to potentially other child clusters that also need to be upscaled.
```
ArturNiederfahrenhorst, line 35: It's clear to me how one layer of virtualization would help share physical clusters with better isolation, but what motivates multiple layers?

edoakes:

PR #51, opened 2024-05-21
REP: Kubernetes Native Ray Authentication

andrewsykim:
Add REP for supporting authentication for Ray clusters created by KubeRay using the Kubernetes TokenReview API

### DIFF
--- a/reps/2024-05-21-kuberay-authentication.md
+++ b/reps/2024-05-21-kuberay-authentication.md
@@ -1,0 +1,159 @@
+# Kubernetes Native Ray Authentication
+
+Ray, in its default configuration, lacks robust security and authentication measures.
+Its deployment on Kubernetes provides some level of protection by leveraging RBAC for access control of RayCluster resources.
+However, once provisioned, a RayCluster remains vulnerable to unauthorized access from anyone with network connectivity to the Ray head node.
+
+This proposal introduces a Kubernetes aware sidecar proxy for authenticating external access to the Ray head.
+It will leverage the existing Kubernetes authentication system to grant tokens used to securely access Ray head endpoints.
+This simplifies Ray authentication by centralizing management of tokens and users within Kubernetes.
+
+## Authentication Scope
+
+While there is a large surface area of a Ray cluster that could benefit from authenticated access, this proposal focuses specifically on securing Ray head endpoints that are frequently accessed externally.
+This includes enforcing authentication for both the dashboard server and the interactive client server. Securing access from internal clients like the Raylet to the GCS server is not addressed in this proposal
+as network policies are typically sufficient to protect this communication.
+
+## Sidecar Authentication Proxy
+
+To enforce authenticated access to specific Ray head ports, the KubeRay operator will deploy a sidecar container alongside the Ray head pod.
+This sidecar will function as a reverse proxy, validating authentication tokens for requests to the dashboard port (8265) and the interactive client port (10001).
+Traffic to other ports will pass through the sidecar proxy without requiring authentication.
+
+The KubeRay operator will be changed in the following ways:
+1. The Ray head container will bind to localhost only
+2. A reverse proxy sidecar will run alongside the Ray head container
+3. A ServiceAccount is created per RayCluster resource
+
+The sidecar ensures only requests with the correct authorization tokens can access the Ray head. More details on how tokens are authenticated below.
+For the MVP, the sidecar will be implemented using Envoy configured with an [External Authorizer](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/ext_authz_filter).
+
+## Access Control with TokenReview API
+
+The sidecar authenticator will use the TokenReview API to authenticate tokens passed into the authorization headers of each request.
+Tokens can be created from Kubernetes ServiceAccounts or an external identity provider as long as the Kubernetes API Server is configured with an external authorizer.
+Tokens from Kubernetes Service Accounts are required to specify a unique audience for the cluster. For now the placeholder audience is "ray.io/cluster/<cluster-name>".
+Each RayCluster will be provisioned a default ServiceAccount that KubeRay Operator will use when authenticating to the Ray head (specifically for RayJob and RayService).
+Users can use the default ServiceAccount in the absence of external identity providers, but using external identity providers is strongly recommended.
+
+Example of a TokenReview request that contains a token:
+```
+apiVersion: authentication.k8s.io/v1
+kind: TokenReview
+spec:
+  token: <your_token_here>
+  audiences:
+  - "ray.io/cluster/<cluster-name>"
+```
+
+Example of a TokenReview response with user information about the token:
+```
+apiVersion: authentication.k8s.io/v1
+kind: TokenReview
+status:
+  authenticated: true
+  user:
+    username: system:serviceaccount:default:ray-cluster
+    uid: <some_uid>
+    groups: ["system:serviceaccounts", "system:serviceaccounts:default"]
+  audiences:
+  - "ray.io/cluster/<cluster-name>"
+```
+
+## User to Ray Head Authentication
+
+When submitting jobs using the Ray job CLI:
+```
+export RAY_JOB_HEADERS="Authorization: Bearer $(kubectl create token ray-cluster)"
+OR
+export RAY_JOB_HEADERS="Authorization: Bearer $(gcloud auth print-access-token)" # example external identity provider
+
+ray job submit...
+```
+
+When using Ray interactive client:
+```
+export RAY_CLUSTER_TOKEN=$(kubectl create token ray-cluster)
+OR
+export RAY_CLUSTER_TOKEN=$(gcloud auth print-access-token) # example external identity provider
+------------------------------------------------------------
+
+import os
+import ray
+import base64
+
+def get_metadata():
+    """
+    Get authentication header metadata for ray.util.connect
+    """
+    headers = {"Authorization": f"Bearer {os.environ["RAY_CLUSTER_TOKEN"]}"}
+    return [(key.lower(), value) for key, value in headers.items()]
+
+ray.init(
+    "ray-cluster:443",
+    metadata=get_metadata(),
+    secure=True,
+)
+```
+
+## RayCluster API Changes
+
+Enabling authentication should be optional per RayCluster. Two new fields `enableAuthentication` and `authenticationOptions` will be added to the RayCluster spec to enable this capability and configure allowed principals.
+If no allowed principals are specified, it will default to a ServiceAccount with the same namespace and name as the RayCluster.
+
+```
+apiVersion: ray.io/v1
+kind: RayCluster
+metdata:
+  ...
+spec:
+  enableAuthentication: true
+  authenticationOptions:
+    allowedPrincipals:
+    - my-user@example.com   # example user
+    - system:serviceaccount:default:ray-cluster  # example service account
+    - system:authenticated  # example group
+```
+
+## Dynamic Access Control with Kubernetes RBAC and the SubjectAccessReview API
+
+Beyond token authentication via the TokenReview API, Kubernetes Role-Based Access Control (RBAC) provides a dynamic mechanism to manage which principals (users or service accounts)
+have access to Ray clusters. This can be achieved by introducing a custom verb and leveraging the SubjectAccessReview API.
+
+A custom verb `admin` can be used in Roles that reference the `rayclusters` resource:
+```
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: ray-admins
+  namespace: my-team
+rules:
+- apiGroups: ["ray.io"]
+  resources: ["rayclusters"]
+  verbs: ["admin"]
+```
+
+Role bindings to this role can grant users access to the Ray cluster. The auth proxy can use the `SubjectAccessReview` API to verify if
+the authenticated user also has admin access to the targetted Ray cluster.
+
+Below is an example `SubjectAccessReview` request that would be sent from the auth proxy:
+```
+apiVersion: authorization.k8s.io/v1
+kind: SubjectAccessReview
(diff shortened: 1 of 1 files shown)


reps/2024-05-21-kuberay-authentication.md
```diff
@@ -0,0 +1,119 @@
...
+3. A ServiceAccount is created per RayCluster resource
+
+The sidecar ensures only requests with the correct authorization tokens can access the Ray head. More details on how tokens are authenticated below.
+For the MVP, the sidecar will be implemented using Envoy configured with an [External Authorizer](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/ext_authz_filter).
+
+## Access Control with TokenReview API
+
+The sidecar authenticator will use the TokenReview API to authenticate tokens passed into the authorization headers of each request.
+Tokens can be created from Kubernetes ServiceAccounts or an external identity provider as long as the Kubernetes API Server is configured with an external authorizer.
+Each RayCluster will be provisioned a default ServiceAccount that KubeRay Operator will use when authenticating to the Ray head (specifically for RayJob and RayService).
```
astefanutti, line 35: It may be useful for the implementation phase, to explicitly mention here the case of RayJobs with submission mode set to `K8sJobMode`, for which the Ray CLI command executed in the submission Job will have to be authenticated as well.

andrewsykim: Good point, will add that!

reps/2024-05-21-kuberay-authentication.md
```diff
@@ -0,0 +1,119 @@
...
+This sidecar will function as a reverse proxy, validating authentication tokens for requests to the dashboard port (8265) and the interactive client port (10001).
+Traffic to other ports will pass through the sidecar proxy without requiring authentication.
+
+The KubeRay operator will be changed in the following ways:
+1. The Ray head container will bind to localhost only
+2. A reverse proxy sidecar will run alongside the Ray head container
+3. A ServiceAccount is created per RayCluster resource
+
+The sidecar ensures only requests with the correct authorization tokens can access the Ray head. More details on how tokens are authenticated below.
+For the MVP, the sidecar will be implemented using Envoy configured with an [External Authorizer](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/ext_authz_filter).
```
kevin85421, line 29: I also chatted with some users about this project. They have their own internal authentication solution. We should make the interface extensible and not be too opinionated about Envoy.

kevin85421:
This PR looks good to me. Maybe we can organize a community design review meeting for this REP.

kevin85421:
cc @thomasdesr would you mind reviewing this PR? Thanks!

reps/2024-05-21-kuberay-authentication.md
```diff
@@ -0,0 +1,157 @@
...
+metdata:
+  ...
+spec:
+  enableAuthentication: true
+  authenticationOptions:
+    ports:
+    - 8265
+    - 10001
+    allowedPrincipals:
+    - my-user@example.com   # example user
```
rueian, line 111: Should this `authenticationOptions` support specifying different sets of `allowedPrincipals`  for different ports?

For example, as a user, I may want to keep the 10001 port private to me but open 8265 to others.

rueian: Yes, as a cluster owner, I should be able to access the client and dashboard ports. However, I may want others to help me monitor the cluster but they should not run any ray programs on it.

andrewsykim: The problem is that the dashboard port also grants access to submit jobs and run programs right? So isolating one or the other doesn't actually prevent users from running Ray programs

andrewsykim: It's a good point, I had initially put ports there because I thought users might want to configure it, but I think it's better to remove it for now and add it later if users ask for it

reps/2024-05-21-kuberay-authentication.md
```diff
@@ -0,0 +1,157 @@
...
+If no allowed principals are specified, it will default to a ServiceAccount with the same namespace and name as the RayCluster.
+
+```
+apiVersion: ray.io/v1
+kind: RayCluster
+metdata:
+  ...
+spec:
+  enableAuthentication: true
+  authenticationOptions:
```
astefanutti, line 106: In some organisations, RayCluster resources will be created by the "data scientist" / end-user persona, while the security aspect will be owned by the platform administrator. The current proposal makes it difficult to enforce that model with Kubernetes RBAC.

The API surface of that new feature should ideally follow usual API graduation. Including it into the RayCluster API, that's v1, will make it difficult to iterate on it and make it evolve, by complying to v1 guarantees.

The RayCluster API has already a large surface. Could the proposal explore the solution space, and research a more modular approach for the configurability aspect, that'd be more symmetrical to the sidecar implementation proposal, i.e. non intrusive, and generally try to keep the single-responsibility principle for the RayCluster API as much as possible.

andrewsykim: > In some organisations, RayCluster resources will be created by the "data scientist" / end-user persona, while the security aspect will be owned by the platform administrator. The current proposal makes it difficult to enforce that model with Kubernetes RBAC.

My assumption is that a platform admin can grant role / rolebindings that give data scientists namespace-wide access to RayClusters (both creating them and accessing them) so they don't need to manage security on a per-cluster basis. You can also make this a ClusterRole to give access to all RayClusters in the cluster. In organizations where data scientists also create RayClusters, they also have the option to by-pass the RBAC check with the fixed list of allowed users (`allowedPrincipals`).

> The API surface of that new feature should ideally follow usual API graduation. Including it into the RayCluster API, that's v1, will make it difficult to iterate on it and make it evolve, by complying to v1 guarantees.

I think this is more of a concern around feature gates and gating changes to v1 APIs that might not be mature yet. I will try to explore adding Kubernetes-like feature gates to KubeRay as we implement this. FYI @kevin85421 

> The RayCluster API has already a large surface. Could the proposal explore the solution space, and research a more modular approach for the configurability aspect, that'd be more symmetrical to the sidecar implementation proposal, i.e. non intrusive, and generally try to keep the single-responsibility principle for the RayCluster API as much as possible.

Let me give this more thought and let's revisit during code review? In theory the addition of a sidecar is already modular because RayCluster lets you override the entire pod template. So my initial thinking is that enabling authentication in KubeRay should default to a solution that will work for most users out of the box (with something like kube-rbac-proxy)

reps/2024-05-21-kuberay-authentication.md
andrewsykim: I removed the option to specify ports

reps/2024-05-21-kuberay-authentication.md
```diff
@@ -0,0 +1,154 @@
...
+2. A reverse proxy sidecar will run alongside the Ray head container
+3. A ServiceAccount is created per RayCluster resource
+
+The sidecar ensures only requests with the correct authorization tokens can access the Ray head. More details on how tokens are authenticated below.
+For the MVP, the sidecar will be implemented using Envoy configured with an [External Authorizer](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/ext_authz_filter).
+
+## Access Control with TokenReview API
+
+The sidecar authenticator will use the TokenReview API to authenticate tokens passed into the authorization headers of each request.
+Tokens can be created from Kubernetes ServiceAccounts or an external identity provider as long as the Kubernetes API Server is configured with an external authorizer.
```
mtaufen, line 34: For KSA tokens, the sidecar should require them to have audiences that do not intersect with the audiences accepted by kube-apiserver. We don't want KSA tokens that grant API access included in requests destined for anywhere other than kube-apiserver. Note that audiences can be specified for mounted tokens using projected volumes: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#serviceaccount-token-volume-projection

reps/2024-05-21-kuberay-authentication.md
```diff
@@ -0,0 +1,154 @@
...
+Each RayCluster will be provisioned a default ServiceAccount that KubeRay Operator will use when authenticating to the Ray head (specifically for RayJob and RayService).
+Users can use the default ServiceAccount in the absence of external identity providers.
+
+Example of a TokenReview request that contains a token:
+```
+apiVersion: authentication.k8s.io/v1
+kind: TokenReview
+spec:
+  token: <your_token_here>
```
liggitt, line 43: +1 to requiring a ray-specific audience here

reps/2024-05-21-kuberay-authentication.md
```diff
@@ -0,0 +1,154 @@
...
+```
+
+Example of a TokenReview response with user information about the token:
+```
+apiVersion: authentication.k8s.io/v1
+kind: TokenReview
+status:
+  authenticated: true
```
liggitt, line 51: the tokenreview response status must contain the audience the token is good for, and should be verified to intersect with the audience requested in the tokenreview spec

reps/2024-05-21-kuberay-authentication.md
andrewsykim: thanks, added a note about KSA requiring audiences

reps/2024-05-21-kuberay-authentication.md
andrewsykim: thanks, added a note about KSA requiring audiences

reps/2024-05-21-kuberay-authentication.md
andrewsykim: thanks, updated the status example to include an audience that will be verified with the audience in the spec

reps/2024-05-21-kuberay-authentication.md
```diff
@@ -0,0 +1,159 @@
...
+    groups: ["system:serviceaccounts", "system:serviceaccounts:default"]
+  audiences:
+  - "ray.io"
+```
+
+## User to Ray Head Authentication
+
+When submitting jobs using the Ray job CLI:
+```
+export RAY_JOB_HEADERS="Authorization: Bearer $(kubectl create token ray-cluster)"
```
ahmedtd, line 67: The user will need to be given impersonation permissions for the service account `default/ray-cluster`.  If multiple users are using the cluster, then all their access will be laundered through this service account.

The recommended usage should probably be the one from line 69, where the user authenticates to the ray cluster using the same credential they use to authenticate to the Kubernetes cluster.

reps/2024-05-21-kuberay-authentication.md
```diff
@@ -0,0 +1,159 @@
...
+OR
+export RAY_JOB_HEADERS="Authorization: Bearer $(gcloud auth print-access-token)" # example external identity provider
```
ahmedtd, line 69: I was surprised that this would work, but it does seem like TokenReview will understand external tokens by calling the Webhook Token Authenticator, if one is configured.

We'll need to think about how to make this mesh with the audiences from above.  GCP (and presumably AWS, Azure, etc) access tokens don't carry scopes, so they won't pass the scope check by default.

reps/2024-05-21-kuberay-authentication.md
andrewsykim: > The recommended usage should probably be the one from line 69, where the user authenticates to the ray cluster using the same credential they use to authenticate to the Kubernetes cluster.

Yes, we will likely recommend this for most cases, the KSA case is only supported for two cases:
1) For KubeRay operator itself that needs to connect to Ray clusters
2) For clusters that don't have external identity providers

reps/2024-05-21-kuberay-authentication.md
```diff
@@ -0,0 +1,159 @@
...
+3. A ServiceAccount is created per RayCluster resource
+
+The sidecar ensures only requests with the correct authorization tokens can access the Ray head. More details on how tokens are authenticated below.
+For the MVP, the sidecar will be implemented using Envoy configured with an [External Authorizer](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/ext_authz_filter).
+
+## Access Control with TokenReview API
+
+The sidecar authenticator will use the TokenReview API to authenticate tokens passed into the authorization headers of each request.
+Tokens can be created from Kubernetes ServiceAccounts or an external identity provider as long as the Kubernetes API Server is configured with an external authorizer.
+Tokens from Kubernetes Service Accounts are required to specify an audience. For now the placeholder audience is "ray.io".
```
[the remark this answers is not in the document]
andrewsykim: good point, updated to require unique auidences

reps/2024-05-21-kuberay-authentication.md
andrewsykim: > I was surprised that this would work, but it does seem like TokenReview will understand external tokens by calling the Webhook Token Authenticator, if one is configured.

I was also pleasantly surprised as well :) 

> We'll need to think about how to make this mesh with the audiences from above. GCP (and presumably AWS, Azure, etc) access tokens don't carry scopes, so they won't pass the scope check by default.

From my testing w/ GKE, the returned audience is something like:
```
audiences:
- https://container.googleapis.com/v1/projects/<project>/locations/us-east4-c/clusters/<cluster>
```

Perhaps we need to specify a list of allowed audiences in the RayCluster if we are required to verify the audience.

ahmedtd: OK, yes, that seems like it would work.

reps/2024-05-21-kuberay-authentication.md
kevin85421: We can discuss more during code review about the API.

kevin85421 (approved):

zhe-thoughts:
We had an in-depth review meeting on this feature on June 4. The vote on ray-committers has also just passed. I'm going to merge this REP

Thanks @andrewsykim for the great contribution and @kevin85421 for shepherding 

Screenshot 2024-07-22 at 2 09 56 PM

PR #55, opened 2024-08-06
[REP] Ray Export API.

MissiontoMars:


reps/2024-06-13-export-api.md
```diff
@@ -0,0 +1,141 @@
+## Summary
+### General Motivation
+
+In the current design of Ray, the way to export various states in the Ray cluster are inconsistent.
```
nikitavemuri, line 4: I believe the State API is typically recommended for end users to fetch current state info of a running Ray cluster, but the state API gets data from multiple sources depending on the resource type. 

The export API will provide another way to fetch state data that works well for large scale clusters or can be saved to query after the cluster is terminated, but I don’t see disjoint interfaces for state data as the main motivation for this feature.

reps/2024-06-13-export-api.md
```diff
@@ -0,0 +1,141 @@
...
+For example, the state of the Actor is broadcast through GCS pubsub, and to obtain the state change of the node,
+it is necessary to query rpc service (NodeInfoGcsService). For the jobs submitted through job submission,
+there is no way to expose the state. The high-level libraries on top of Ray also don't have a unified export way.
+E.g. RayServe and RayData collect the states through their own StateActor and report to the Dashboard respectively.
+
+It is very difficult to obtain these ray basic states outside the Ray cluster. If a unified export API can be defined,
```
nikitavemuri, line 10: Would be good to mention it’s difficult to view these states when (1) scale of data exceeds current dashboard API limits (2) after the cluster is terminated

reps/2024-06-13-export-api.md
```diff
@@ -0,0 +1,141 @@
...
+we can achieve the observable ability independent of the Ray cluster. The most typical scenario is to build the Ray history server.
+
+#### Key requirements:
+- Need to expose all necessary ray states for basic observability, tasks/actor/jobs/nodes.
```
nikitavemuri, line 14: A good initial use case of this data is rebuilding the dashboard APIs, so we should make sure the schema for each resource contains all required fields. Timestamps should also be included so the reader can reconstruct a history of state changes.

reps/2024-06-13-export-api.md
```diff
@@ -0,0 +1,141 @@
...
+- The states export should not increase the load of the ray cluster.
```
nikitavemuri, line 15: Some load will be added if we are writing this data to file, but we should be able to limit this impact (eg: max data stored in memory, not blocking any control logic)

reps/2024-06-13-export-api.md
```diff
@@ -0,0 +1,141 @@
...
+- Export states streamingly rather than fetching it directly from ray cluster. To obtain the current status of the ray cluster at one time,
+  you need to use [the state observability api](https://github.com/ray-project/enhancements/blob/main/reps/2022-04-21-state-observability-apis.md "the state observability api") . 
+- Certain states can be exported respectively.
+- Friendly to all types of users (especially cloud vendors), easy to deploy and use, without modifying Ray.
+
```
nikitavemuri, line 20: Let’s also add the requirements
- It should be possible to disable the export API for security and performance
- The data should be exported from as close as possible to the source where it is generated. The export API should have minimal overhead and we should not need to unnecessarily move data to other nodes before it is exported.

reps/2024-06-13-export-api.md
```diff
@@ -0,0 +1,141 @@
...
+It is very difficult to obtain these ray basic states outside the Ray cluster. If a unified export API can be defined,
+we can achieve the observable ability independent of the Ray cluster. The most typical scenario is to build the Ray history server.
+
+#### Key requirements:
+- Need to expose all necessary ray states for basic observability, tasks/actor/jobs/nodes.
+- The states export should not increase the load of the ray cluster.
+- Export states streamingly rather than fetching it directly from ray cluster. To obtain the current status of the ray cluster at one time,
+  you need to use [the state observability api](https://github.com/ray-project/enhancements/blob/main/reps/2022-04-21-state-observability-apis.md "the state observability api") . 
+- Certain states can be exported respectively.
```
nikitavemuri, line 18: I didn’t quite get this point, could you clarify?

reps/2024-06-13-export-api.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Events to Export and How
+#### Events to Export
+- Core
+Tasks/Actors/Objects/Jobs/Nodes/Placement groups, including its meta data and current status when event emits.
+
+|  Event            | Event Source          | When to export                                                                    | Format example(maybe a file link)|
+|  ----             | ----                  | ----                                                                              | ----                             |
+| Tasks             | CoreWorker and raylet | When coreworker adds task event to<br>ray::core::worker::TaskEventBuffer          |          |
+| Actors            | GCS                   | GcsPublisher::PublishActor<br>publish actor status through<br>GCS_ACTOR_CHANNEL   |          |
```
nikitavemuri, line 65: Schema for actors: https://docs.google.com/document/d/1_7Z81KtmjA3IAEeZqEuxvw6ur5acyU-CrofQ6rFEmTg/edit

reps/2024-06-13-export-api.md
```diff
@@ -0,0 +1,141 @@
...
+| Objects           | CoreWorker and raylet | None                                                                              |          |
+| Jobs              | JobManager            | When JobInfoStorageClient.put_status called                                       |          |
```
nikitavemuri, line 67: Schema for jobs: https://docs.google.com/document/d/1upQRU-f8WgVH_NWBmeJyegyOSJwNDPPqnl1cCGpmiGo/edit?usp=sharing

reps/2024-06-13-export-api.md
```diff
@@ -0,0 +1,141 @@
...
+| Nodes             | GCS                   | GcsNodeManager::HandleXXXNode                                                     |          |
```
nikitavemuri, line 68: Schema for nodes: https://docs.google.com/document/d/1qjoF51h2oUN2sr_MtPnovbNFZYZrh3WLNR_P0HrUuOI/edit?usp=sharing

reps/2024-06-13-export-api.md
```diff
@@ -0,0 +1,141 @@
...
+| Placement groups  | GCS                   | GcsPlacementGroupManager::HandleXXXPlacementGroup                                 |          |
+
+
+- RayServe
```
nikitavemuri, line 72: Let's mark objects, placement groups, Ray Serve, and Ray Data as P1s. We can come up with the schema for these after the initial P0 core resources

reps/2024-06-13-export-api.md
```diff
@@ -0,0 +1,141 @@
...
+
+
+#### How to export
+- Generate event
+We propose to use the filesystem based export interface. The event source can just write the
+event into the file of this node and then return. The export of the event can be the responsibility of the agent on the node. We know that there will be multiple worker processes running multiple actors or tasks on each Ray node. Events generated by multiple worker processes are written to multiple event files, not just one, in order to avoid competition among multiple worker processes or possible file writing errors.
+
+
+- Export event
+We propose to use pubsub for event export, because pubsub is easier to decouple systems,
```
nikitavemuri, line 94: My opinion is we can limit this REP to generating the events and let users implement exporting the events to an external system. We can share recommendations on how to do this (eg: log ingestion using Vector which can send data to various sinks per the use case), but I think the pub sub event export isn’t necessary for this version to keep the interface simple and flexible

reps/2024-06-13-export-api.md
MissiontoMars: Make sense. I mentioned it in the Key requirements: 
To obtain the current status of the ray cluster at one time, you need to use state api.

reps/2024-06-13-export-api.md
MissiontoMars: @nikitavemuri RayCores (actors/tasks/nodes), RayData, and RayServe-related states can be selectively exported. For example, if a user is only interested in the state of RayServe and not in the state of RayCore, then the export of RayCore state can be disabled to reduce overhead.

reps/2024-06-13-export-api.md
MissiontoMars: LGTM.

reps/2024-06-13-export-api.md
MissiontoMars: Let me try to understand. Do you mean that we can use the data format of the dashboard api to correct the completeness of the exported data by the export api? For example, exported state is published to gcs, and dashboard gets data from gcs.

reps/2024-06-13-export-api.md
MissiontoMars: LGTM.

reps/2024-06-13-export-api.md
MissiontoMars: Or instead, we give some possible export methods (such as gcs or vector) as a suggestion, not a proposal?

reps/2024-06-13-export-api.md
```diff
@@ -0,0 +1,153 @@
...
+For example, the state of the Actor is broadcast through GCS pubsub, and to obtain the state change of the node,
+it is necessary to query rpc service (NodeInfoGcsService). For the jobs submitted through job submission,
+there is no way to expose the state. The high-level libraries on top of Ray also don't have a unified export way.
+E.g. RayServe and RayData collect the states through their own StateActor and report to the Dashboard respectively.
+
+It is very difficult to obtain these ray basic states outside the Ray cluster. This issue is reflected in two aspects:
+1. Scale of data exceeds current dashboard API limits.
+2. After the Ray cluster terminates, all status data is lost.
+If a unified export API can be defined,
+we can achieve the observable ability independent of the Ray cluster. The most typical scenario is to build the Ray history server.
```
nikitavemuri, line 14: I would re-organize this to something like 
```suggestion
There are several major improvements requested by users about the Ray Dashboard
1. Currently, all state data is lost when the Ray cluster terminates and users are looking for persistence of this data.
2. The Ray dashboard has scalability limits on the amount of data stored and returned. Advanced users would benefit from greater scalability.
3. There is no unified and accurate way to get state events pushed on state change across various resource types. The State and dashboard APIs support a pull model, where it is possible to miss some events. There is also not a consistent way across resource types to get this information directly from components that generate it.

A unified export API could serve as a base to address these issues by allowing state data to be exported and processed separately. This would ensure that observability features remain functional independent of the Ray cluster.
```

nikitavemuri: > to obtain the state change of the node, it is necessary to query rpc service (NodeInfoGcsService)

Just curious, do you or other users already query these GCS services directly to get data?

nikitavemuri (commented):
Mostly rewording and restructuring comments

nikitavemuri:
> I have two questions:
> 
> * It looks like that users can not get a completed history server after the implementation of this REP. Do you have any plans to support a default history server including UI?

Yeah, we wanted to open source the export API as the first step for various dashboard persistence and scalability features. We are looking into prototyping a history server to better understand the design options. 

One open question for a full history server is choosing the right backend and how to interface with it. We need to choose a light weight backend, or we should probably support plugging into multiple backends.

> * What's the status of this REP? Do you have any plans for the implementation of this REP?

Yes, I have an initial working version of this in review, but we need to still verify if there are any performance regressions

edoakes:

PR #56, opened 2024-10-17
[WIP][REP][Serve] Add proposal for API allowing user-defined autoscaling and scheduling algorithms

arcyleung:
This REP aims to extend the existing Ray serve.deployment functionality for users to define their own autoscaling and scheduling policy. The existing policies are request-queue-length scaling and power-of-2 scheduling, while the configs themselves can be modified (ie. target_ongoing_requests) the policy's algorithm and monitored heuristics currently cannot be changed by the user. For instance, if the user wants to specify scaling by number of SLA violations detected to proactively scale deployments, they cannot use a request queue length heuristic when each request has variable latency.

### DIFF
--- a/reps/2024-10-03-user-defined-scheduling-scaling.md
+++ b/reps/2024-10-03-user-defined-scheduling-scaling.md
@@ -1,0 +1,43 @@
+# User-defined scheduling and scaling policies for Ray Serve deployments
+
+## Summary
+Provide support for user-defined scheduling and scaling policies
+
+## Motivation
+
+### General Motivation
+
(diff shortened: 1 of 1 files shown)


abrarsheikh:
this got implemented, but with a different API structure 
1. custom autoscaling https://docs.ray.io/en/latest/serve/advanced-guides/advanced-autoscaling.html#custom-autoscaling-policies
2. custom request routing https://docs.ray.io/en/latest/serve/advanced-guides/custom-request-router.html

Hence closing the PR for REP.

PR #57, opened 2024-10-18
Train + Tune API Revamp

justinvyu:
## Summary

Ray Tune and Ray Train have been tightly coupled since Ray 2.0, when Ray Tune became the common execution engine for both libraries.

Ray Train execution invokes Tune’s execution logic under the hood, which leads to a complex, layered system.
The original intention behind this was to increase the interoperability of the two libraries, but the dependency of Ray Train on Ray Tune has led to many usability and stability issues, and it has stalled feature development.

The goal of these changes is to improve:
* **Usability** by mostly keeping feature parity, while introducing more intuitive APIs in place of inherited Ray Tune APIs that do not fit in the context of Ray Train.
* **Extensibility** by introducing more modular execution components that can be more easily customized.
* **Interoperability** between Ray Tune and Ray Train so that using both together is more natural and still provides separation between the two libraries.
* **Testability** with proper unit tests rather than hundreds of mini end to end tests.


hongpeng-guo (dismissed):
This is legendary! Thanks for putting everything together.

matthewdeng (approved):

hongpeng-guo (approved):

raulchen (approved):

reps/2024-10-18-train-tune-api-revamp/2024-10-18-train-tune-api-revamp.md
```diff
@@ -0,0 +1,1194 @@
...
+    parser.add_argument("run_id", type=str)
+    args = parser.parse_args()
+
+    unique_run_name = args.run_id  # some UNIQUE job id
+    storage_path = "s3://my/bucket"
+
+    trainer = TorchTrainer(
+        train_fn_per_worker,
+        ...,
+        run_config=ray.train.RunConfig(
```
robertnishihara, line 892: A few questions

1. If there is nothing to restore, e.g., this is the first time the job is submitted, then `unique_run_name` will be `None` and the `TorchTrainer` will see that there is no checkpoint to load from and will not load anything?
2. What's the reason for removing the explicit `restore` call?
3. Why do I need the storage info in `RunConfig`? Couldn't that all just go into `get_checkpoint`?

reps/2024-10-18-train-tune-api-revamp/2024-10-18-train-tune-api-revamp.md
```diff
@@ -0,0 +1,1194 @@
...
+![Train V2 Architecture](train_architecture.png)
+
+### Controller
+
+The **Controller** is the central orchestrator that manages the distributed execution of training jobs. Its main responsibilities include:
+
+* **Execution Control**:
+  * It triggers the distributed training function across the workers in the `WorkerGroup` and shuts down the workers when training has either finished or failed.
+* **Polling & Health Checking**:
+  * The Controller regularly polls the workers to assess their health and training progress. Based on the worker statuses, it can handle errors and make scaling decisions.
```
robertnishihara, line 557: This looks awesome. A few questions.
Training health monitoring is becoming more and more important and can be quite complex and compute intensive (not just looking at lots of metrics, but also running evals). Where does this logic live?

If the health monitoring takes a while to run but eventually decides that we need to restart from an earlier checkpoint, how does that get implemented?

reps/2024-10-18-train-tune-api-revamp/2024-10-18-train-tune-api-revamp.md
```diff
@@ -0,0 +1,1194 @@
...
+
+#### `resume_from_checkpoint`
+
+❌ `Trainer(resume_from_checkpoint)` was a common source of confusion and will be deprecated. `resume_from_checkpoint` is meant to provide a starting checkpoint for a new run, rather than restore all state from the previous run. See the following example for a more intuitive way to accomplish the same thing:
+
+```py
+def train_fn_per_worker(config: dict):
+    # Equivalent behavior that is explicit and more flexible.
+    checkpoint = (
+        ray.train.get_checkpoint()
```
robertnishihara, line 918: Minor, but isn't `config.get("resume_from_checkpoint")` enough? Why do you also need `ray.train.get_checkpoint()`?

reps/2024-10-18-train-tune-api-revamp/2024-10-18-train-tune-api-revamp.md
justinvyu: > Where does this logic live?

This logic lives in the Train driver right now, where the controller actor periodically pings the worker actor tasks and handles any errors raised by those health-check tasks.

> If the health monitoring takes a while to run but eventually decides that we need to restart from an earlier checkpoint, how does that get implemented?

Health-checks are every few seconds at the moment, so the decisions happen pretty quickly after Ray detects that a node/actor has died. We also handle more extreme edge-cases where the underlying RPC times out randomly, in which case we'll wait for some configurable time until restarting from the latest checkpoint.

> (not just looking at lots of metrics, but also running evals)
Oh, maybe this part was a bit misleading. At the moment, the "health and training progress" just refers to the underlying actor/node health and reported checkpoints, rather than the health of the training jobs (we do NOT implement anything to handle how well things are converging, how validation metrics are doing, etc).

reps/2024-10-18-train-tune-api-revamp/2024-10-18-train-tune-api-revamp.md
justinvyu: `ray.train.get_checkpoint` is still needed if you want to support fault tolerance, in addition to a checkpoint to start fine-tuning from initially. Otherwise, if a node fails, the training progress always gets reset to the initial checkpoint, throwing away all finetuning progress.

PR #58, opened 2024-12-19
REP: KubeRay RayService Incremental Rollout

ryanaoleary:
Add REP for RayService incremental upgrade strategy using `max_surge_percent` and Gateway API.


ryanaoleary:
cc: @kevin85421 @andrewsykim

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+Example RayService manifest:
+```sh
+apiVersion: ray.io/v1
+kind: RayService
+metadata:
+  name: example-rayservice
+spec:
+  upgradeStrategy: "NewCluster"
```
andrewsykim, line 67: It's not too late to change this API, should we change this now before v1.3 so it'll play nicely with the new fields you're proposing here?
```
spec:
  upgradeStrategy:
    type: "NewCluster"
    targetCapacity: 50
    maxSurgePercent: 50
...
```

andrewsykim: At least for 1.3, we can just support
```
spec:
  upgradeStrategy:
    type: NewCluster
```

So it's easy to add new fields in v1.4 that are compatible with the `upgradeStrategy`

ryanaoleary: a5ea9db, API change: https://github.com/ray-project/kuberay/pull/2678

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,125 @@
...
+    ...
+```
+
+#### Key requirements:
+- If users want to use the original blue/green deployment, RayService should work regardless of whether the K8s cluster has Gateway-related CRDs
+- If users specify both `target_capacity` and `max_surge_percent` Ray should validate whether there is a conflict in the values before proceeding
+
+### Gateway API
+
+Kubernetes Gateway API is a set of resources designed to manage HTTP(S), TCP, and other network traffic for Kubernetes clusters. It provides a more flexible and extensible alternative to the traditional Ingress API, offering better support for service mesh integrations, routing policies, and multi-cluster configurations. Gateway API already has a well-defined interface for traffic splitting, so KubeRay would not need to implement backends for different Ingress controllers.
```
kevin85421, line 82: which persona is responsible for which Gateway resources? For example, `GatewayClass` is created / maintained by the K8s cluster admin.

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,125 @@
...
+
+| API |  Use   | Status |  K8s version  |
+| ------- | ------ | ---------- | ------------- |
+|  v1.GatewayClass | Defines a Gateway Cluster level resource | GA (v0.5+) | v1.24+ |
+|  v1.Gateway | Infrastructure that binds Listeners to a set of IP addresses | GA (v0.5+) | v1.24+ |
+|  v1.HTTPRoute | Provides a way to route HTTP requests | GA (v0.5+) | v1.24+ |
+|  v1.GRPCRoute | Specifies routing behavior of gRPC requests from a Gateway listener to an API object | GA (v1.1+) | v1.25+ |
+
+### Example Upgrade Process
+
```
kevin85421, line 94: how does the Ray Serve API / KubeRay API / K8s Gateway API interact with each other? For example, what's actually happening under the hood for "The KubeRay controller switches `target_capacity` percent of the requests to the new RayCluster once it is ready."?

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,125 @@
...
+4. KubeRay scales down the old RayCluster according to the `target capacity` and `max_surge_percent`.
+5. Users then increase the `target_capacity` and repeat the above steps until the `target_capacity` reaches 1.
+
+Automatic Upgrade:
+1. The user only specifies `max_surge_percent`.
+2. The KubeRay controller creates a Gateway and Routes for the RayService instance, if this is not possible, the controller defaults to a blue/green upgrade strategy.
+3. The upgraded RayCluster scales up to (1 + `max_surge_percent` - `target_capacity_old`), and the old RayCluster decreases its `target_capacity` until `target_capacity_new` plus `target_capacity_old` equals 1.
+
+## Compatibility, Deprecation, and Migration Plan
+
```
kevin85421, line 110: summarize the limitation / assumption for the incremental upgrade. For example, the new RayCluster should have the capacity to handle the same or more requests comparing to the old RayCluster. If users don't follow it, some requests may be dropped.

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,125 @@
...
+Kubernetes Gateway API is a set of resources designed to manage HTTP(S), TCP, and other network traffic for Kubernetes clusters. It provides a more flexible and extensible alternative to the traditional Ingress API, offering better support for service mesh integrations, routing policies, and multi-cluster configurations. Gateway API already has a well-defined interface for traffic splitting, so KubeRay would not need to implement backends for different Ingress controllers.
+
+Gateway APIs:
+
+| API |  Use   | Status |  K8s version  |
+| ------- | ------ | ---------- | ------------- |
+|  v1.GatewayClass | Defines a Gateway Cluster level resource | GA (v0.5+) | v1.24+ |
+|  v1.Gateway | Infrastructure that binds Listeners to a set of IP addresses | GA (v0.5+) | v1.24+ |
+|  v1.HTTPRoute | Provides a way to route HTTP requests | GA (v0.5+) | v1.24+ |
```
kevin85421, line 90: Add examples of HTTPRoute that KubeRay needs to create https://gateway-api.sigs.k8s.io/api-types/httproute/.

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,125 @@
+# Support RayService Incremental Upgrades
+
+## General Motivation
+
+Currently, during a zero-downtime upgrade on a RayService, the KubeRay operator temporarily creates a pending RayCluster and waits for it to become ready. Once ready, it switches all traffic to the new cluster and then terminates the old one. That is, both RayCluster custom resources require 100% computing resources. However, most users do not have access to such extensive resources, particularly in the case of LLM use cases where many users [lack sufficient high-end GPUs](https://github.com/ray-project/kuberay/issues/1476).
+
+It'd be useful to support an API to incrementally upgrade the RayService, scaling a new RayCluster to handle only a % capacity of the total traffic to the RayService in order to avoid delays in the upgrade process due to resource constraints. In order to enable this, we propose the following APIs and integration with [Gateway](https://github.com/kubernetes-sigs/gateway-api).
```
JoshKarpel, line 7: An additional benefit to discuss is that a smooth traffic migration (instead of abrupt like it is now) means that autoscaling has a chance to kick in on the new cluster (assuming you configure the timescales appropriately) - right now, one problem we face sometimes is that if you deploy under high load, the new cluster doesn't actually know about the high load until it sees it, but if you don't have a high initial replica count for whatever reason, you might get a service disruption, at least in terms of latency.

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,125 @@
...
+apiVersion: ray.io/v1
+kind: RayService
+metadata:
+  name: example-rayservice
+spec:
+  upgradeStrategy: "NewCluster"
+  upgradeSpec:
+    type: "NewCluster"
+    targetCapacity: 50
+    maxSurgePercent: 50
```
JoshKarpel, line 71: Hmmm, given the description above, I was expecting to see something like a timescale here, e.g., it migrates X% of the traffic every N minutes. Could you talk more about what exactly the timeline is during a deploy?

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,125 @@
...
+Automatic Upgrade:
+1. The user only specifies `max_surge_percent`.
+2. The KubeRay controller creates a Gateway and Routes for the RayService instance, if this is not possible, the controller defaults to a blue/green upgrade strategy.
+3. The upgraded RayCluster scales up to (1 + `max_surge_percent` - `target_capacity_old`), and the old RayCluster decreases its `target_capacity` until `target_capacity_new` plus `target_capacity_old` equals 1.
+
+## Compatibility, Deprecation, and Migration Plan
+
+### Compatibility
+
+Below is a compatibility matrix for the proposed supported versions of Kubernetes, KubeRay, and Gateway API:
```
JoshKarpel, line 113: This is more about the documentation for this, but it would be nice to add information about what users of various k8s setups might need to do to make sure this works for them - e.g., we use Istio, which it looks like does support the Gateway API https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/ , and it might be nice to have a "how to make this work with Istio/whatever other popular k8s networking layers people are using" section(s) in the docs for this feature.

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: I was thinking that `target_capacity` % of traffic would be switched over immediately once the new Pods in the upgraded RayCluster replicas were marked as ready. Does this fit your idea of the use-case? We could also add a `minReadySeconds` field to specify the delay before switching over traffic to a worker that is newly marked as ready.

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: I think that's a good idea, I'll add a mention in the REP of the documentation plan for this feature and its compatibility.

reps/2024-12-4-ray-service-incr-upgrade.md
JoshKarpel: Ah, ok, I get it - and I see that I missed https://github.com/ray-project/enhancements/pull/58/files#diff-931e7727c347f0773468d4e7398fa53a04a143bbf8a24bfbc2a99e66d5f23275R95-R107 below which explains this!

A key interaction that I'm concerned about here is Ray Serve autoscaling - see my comment here https://github.com/ray-project/enhancements/pull/58/files#r1913488877 .

The thing about Ray Serve autoscaling is that the timescales it operates on are user-configurable, and therefore the time to stabilize depends on those parameters. If running a cluster at far above `initial` replica counts, an instantaneous transition with a large `target_capacity` could be disastrous, because the new cluster hasn't seen any traffic yet and then will suddenly get an instantaneous traffic burst that it's not scaled up for. This actually happens to us all the time in our current setup - we keep `initial` replica counts low so that we have enough resources to spin up the new cluster, but when the swap happens, that means we don't have enough replicas to handle the traffic!

This is just some quick late-night brainstorming, but what I'd love to see is for KubeRay to manage a gradual traffic migration from one cluster to the other, with the rate controlled by config, so that traffic smoothly swaps from one cluster to the other. For example, suppose you estimate that it takes ~1 minute for autoscaling to stabilize given your settings (target concurrency, how long it takes to start a deployment replica, etc.; 1 minute is probably very short for real systems, ours can take 5-10 minutes if we need to autoscale new k8s nodes). Your config might be like
```
rate: 0.05
interval: 60
```
i.e., swap 5% of traffic every 60 seconds.

The timeline would look like this:
- KubeRay decides we need a new Ray cluster for whatever reason.
- The new cluster starts up, Serve apps are deployed with `target_capacity = 0.05 * 1 = 0.05` (presumably all min/initial/max replica counts are min 1?)
- Once all Serve apps are ready, KubeRay uses Gateway to move 5% of traffic to the new cluster, then waits 60 seconds (one interval). This interval gives the new cluster time to react to the traffic (autoscaling new replicas, populating caches, whatever is needed to stabilize at 5% traffic - it's up to the user to estimate how long this would take, though one could image a very fancy system where the cluster actually tells KubeRay it's ready, maybe once all Serve deployment replica counts are stable for some time?).
- KubeRay changes the `target_capacity` to `0.05 * 2 = 0.1`, then moves another 5% of traffic, so that 10% of traffic is going to the new cluster and 90% to the old. The `target_capacity` of the old cluster is changed to `1 - (0.05 * 1) = 0.95` (i.e., it is one `rate` tick behind). Then it waits another interval.
- Repeat until all traffic is migrated, then shut down the old cluster.

The effective max surge in this case is equal to the `rate`, and users can tune the rate and interval depending on their configuration. `rate=1` is equivalent to the existing behavior of swapping all traffic once the new cluster is ready.

The current state of the system (how many intervals have passed) could be stored in the `RayService` CRD so that KubeRay could recover from a crash mid-roll.

This is obviously more complex than what's being proposed here, but it lets you move the `rate = max_surge` closer to `0` without making the deploy unstable.

trevenrawr: Howdy, folks!  Late to the discussion, but here to drop a +1 on Josh's "gradual traffic" thoughts.  I may be missing something (again, late to the party 😅) but it seems to me that if traffic were gradually transitioned, there wouldn't be a need for `target_capacity` or `max_surge_percent` at all as they become metrics by which to measure a transition rather than declarations of how the transition should progress.

I'm imagining something like the following:
- New cluster comes up with `min_replicas` (or `initial_replicas` if configured) for all deployments - no change
- KubeRay starts to migrate traffic as per Josh's proposal; every `interval` seconds bump the traffic fraction to the new cluster by `rate` amount (lots of enhancement options here too, e.g., have KubeRay wait for `interval` seconds, then continue waiting until any pending worker pods in the new cluster are ready before continuing to the next tick)
- In the original cluster, traffic drains and replicas/workers spin down automatically
- In the new cluster, traffic ramps and replicas/workers spin up automatically
- After the transition period, the original cluster should be down to `min_replicas` and the new cluster should be up to the number of replicas necessary to serve traffic

Of course there are some potential mis-configuration possibilities here.  A few quick examples:
1. User configures transition to occur too quickly: status-quo
2. User has very high `initial_replica` count but traffic to new cluster doesn't require that many replicas any more so there's replica thrash during the transition
3. Replicas downscale slower than the transition occurs, meaning the original cluster doesn't drain during the transition (high `max_surge_percent`)

...but all of these are fully under the user's control, which means that they can decide what settings are appropriate for their workloads.
1. User can configure a longer transition time
2. User can remove `initial_replicas` entirely so new clusters are minimally viable and rely on the traffic ramp to warm them up
3. User can expedite downscaling or extend transition time; docs can recommend an `interval` relative to downscale delay configuration.

The real benefit here is that the transition mechanism is simple: just slowly migrate traffic and let each cluster behave as it would if it were running in a vacuum.  Or, to think of it another way, it's akin to a [reversible process](https://en.wikipedia.org/wiki/Reversible_process_(thermodynamics)) in thermodynamics.

One downside is that it could be slow to transition fully, but waiting for a new cluster to acquire enough resources to satisfy large `initial_replicas` counts can also take a long time and this would enable the transition to begin sooner.  (And may also enable nodes dropping out of the original cluster to be immediately taken up by the new one.)

JoshKarpel: Very elegant @trevenrawr ! And I think that the way you describe makes it clear that this behavior is actually still quite simple, especially since it seems like the `target_capacity` behavior isn't required.

ryanaoleary: This solution makes sense to me, essentially allowing the gradual shift in traffic flow to dictate the capacity of each cluster rather than having the user set the `target_capacity` themselves, wait until the replicas are ready, and then switch the traffic. 

For the way it's currently described in the REP, I think `target_capacity` would involve automatically setting the initial replicas of the upgraded cluster to the amount required to handle the load to the new cluster, so there shouldn't actually be a warm-up time in terms of marking the upgraded cluster as ready (i.e. `replicas` have all scaled up) and it being able to handle `target_capacity` traffic from the `HttpRoute`. I think that would be the case only if we implemented the upgraded cluster to start with `minReplicas`, unless there's more "warm-up" operations that I'm missing besides just scaling up the required number of replicas. Either way, the time to transition fully should be about the same for both options.

I still see some use in retaining `target_capacity` and `max_surge_percent` in terms of capping the amount of traffic that can transition to the new cluster. The use-case I'm imagining is if the user wants to test some change and initiates an incremental upgrade such that a certain % of their traffic switches to the new cluster, without completely scaling down the old cluster. Another scenario could be variable wait times in provisioning the nodes when transitioning to the new cluster at `rate` + `interval` could leave a large number of pending Pods while the old cluster downscales, potentially leading to a disruption in traffic. Though as you mentioned, issues like this could be ameliorated by having KubeRay wait until all pending worker Pods are running.

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: A smooth traffic migration sounds good to me, I can mention that in the REP. The current proposal is more "replica" driven with `target_capacity` dictating the initial replicas to scale up on the new cluster, and then switching that same % of traffic to the new cluster once all replicas are marked as ready. I think in terms of latency both options should be the same (i.e. switching traffic and then scaling from `minReplicas` or scaling up the # of replicas dictated by `target_capacity` and then splitting the traffic), but the former option might be a more clear upgrade path for users.

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: [e0a3c19](https://github.com/ray-project/enhancements/pull/58/commits/e0a3c193babba2abaaf8865a9281a14cafba923f)

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: Per some offline discussion with @andrewsykim, there's an issue with the above strategy for incremental upgrades on a static RayService. In this case, `target_capacity` and/or `max_surge_percent` would be required as we could not rely on traffic splitting to drive autoscaling in the upgraded RayCluster to the required number of replicas. The current proposal is based on the existing Kubernetes Deployment rolling upgrade strategy of [maxSurge](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#max-surge).

One thing to also note is that, with the `Automatic` path in the REP (i.e. when only `max_surge_percent` is specified) `target_capacity` will default to `100` and the cluster will switch over traffic at the rate that the new replicas can be scaled up. While this isn't exposed as a straightforward `rate` to the user, it still ensures a gradual migration of traffic to the upgraded cluster without additional latency or the risk of dropped traffic.

I think the main point to consider between these two options is whether it's better to drive the upgrade by splitting traffic (I guess only for RayServices with autoscaling enabled) or by scaling up replicas and then switching traffic over.

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: At a high level I think it would work as follows:
1. K8s cluster admin installs Gateway CRDs and Gateway controller that is compatible with their infrastructure backend.
2. When a RayService with a `RayServiceUpgradeType` of `IncrementalCluster` is created, the KubeRay controller validates the values of `target_capacity` and `max_surge_percent` and then checks whether the Gateway CRDs and controller exists using the Kubernetes python client. If the controller doesn't exist, defaults to a blue/green deployment.
3. KubeRay controller then checks the [Serve config files](https://docs.ray.io/en/releases-2.7.0/serve/production-guide/config.html#serve-config-files-serve-build) and creates a `Gateway`.

Example Gateway:
```
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: <ray-service-name>-gateway
spec:
  gatewayClassName: <ray-service-name>-gateway
  listeners:
  - name: http
    protocol: HTTP
    port: http_options.port // defaults to 8000
 - name: grpc // can be ommitted if grpc_servicer_functions is empty
    protocol: GRPc
    port: grpc_options.port // defaults to 9000
```

4. KubeRay controller creates the new cluster with initial `replicas` = `min(target_capacity, max_surge_percent)`/100 * `total_ray_cluster_replicas`. 

5. Once the new RayCluster is ready, the KubeRay controller creates `HTTPRoute`s and `GRPcRoute`s with settings according to the serve config and `target_capacity` weight for the new RayCluster head service.

Example HTTPRoute:
```
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: httproute-<gateway-name>
spec:
  parentRefs:
  - kind: Gateway
    name: <gateway-name>
  hostnames:
  - "ray.example.com"
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: <route_prefix> // set according to serve config
// add more paths according to `applications` in serve config
    - backendRefs:
        - name: <head-svc-of-original-ray-cluster>
          weight: 1-<target-capacity>
          port: http_options.port // defaults to 8000
       - name: <head-svc-of-upgraded-ray-cluster>
          weight: target-capacity
          port: http_options.port // defaults to 8000
```

Example GRPcRoute:
```
apiVersion: gateway.networking.k8s.io/v1
kind: GRPCRoute
metadata:
  name: grpcroute-<gateway-name>
spec:
  parentRefs:
  - kind: Gateway
    name: <gateway-name>
  hostnames:
  - "ray.example.com"
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: <route_prefix> // set according to serve config
// add more paths according to `applications` in serve config
    - backendRefs:
        - name: <head-svc-of-original-ray-cluster>
          weight: 1-<target-capacity>
          port: grpc_options.port // defaults to 9000
       - name: <head-svc-of-upgraded-ray-cluster>
          weight: target-capacity
          port: grpc_options.port // defaults to 9000
```

6. While `target_capacity` of the new RayCluster is less than `target_capacity` set in the RayService `UpgradeSpec`, KubeRay controller loops repeating steps 4 and 5, updating the upgraded RayCluster replicas to `min(target_capacity, max_surge_percent)`/100 * `total_ray_cluster_replicas`, waiting for the new replicas to become ready, and updating the weights on the Gateway API Routes to `target_capacity` and `1-target_capacity` for the new and old RayClusters respectively.

7. Once `target_capacity` of the new RayCluster is equal to 1, KubeRay controller removes `Gateway` and routes. 

The KubeRay API interacts with the RayServe API by checking the serve config when configuring the Gateway and routes, and with Gateway API when creating those resources and then cleaning them up. I'm working on putting code snippets from my prototype in the REP which will follow the above design.

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: I put some in https://github.com/ray-project/enhancements/pull/58#discussion_r1938000414 and I'll copy those into the doc.

reps/2024-12-4-ray-service-incr-upgrade.md
kevin85421: I briefly read through this thread. My opinion is that `target_capacity` is still needed. 

Without `target_capacity`, KubeRay switches traffic before the RayCluster has enough capacity to handle the number of requests. Ray autoscaling is complicated, and the overhead is unpredictable. In the worst case, it requires Ray Serve to scale up a new serve replica, Ray Autoscaler and KubeRay to scale up a new Ray Pod, and the K8s Autoscaler to scale up a new K8s node. Each part in this calling chain that goes wrong may cause requests to be dropped.

That's why I prefer to switch traffic only when the RayCluster has enough capacity to handle the number of requests.

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: [d543860](https://github.com/ray-project/enhancements/pull/58/commits/d5438601abd873bcd794ea559bed53c507f64bbe)

reps/2024-12-4-ray-service-incr-upgrade.md
kevin85421: > An additional benefit to discuss is that a smooth traffic migration (instead of abrupt like it is now) means that autoscaling has a chance to kick in on the new cluster (assuming you configure the timescales appropriately)

This contradicts the incremental upgrade assumption. The current assumption is that:

* The new cluster's stable state should have equal or greater capacity than the old RayCluster to handle requests.
* `target_capacity` means that the new RayCluster has the capacity to handle `target_capacity`% of the requests.

If requests are dropped, then some assumptions have not been followed.

kevin85421: The gradual traffic switch is still a good idea for me.

kevin85421:
Discuss offline with the community:

* This REP should support rollback.
* If users update the RayService CR during the upgrade process, KubeRay should first roll back and then roll out to the target state.

Here are some thoughts on (2):
* Terminology:
  * Old cluster
  * New cluster
  * Goal state: the state currently defined in the RayService CR

During the upgrade process from the old cluster to the new cluster:

* If the goal state doesn't match either the old cluster or the new cluster, roll back from the new cluster to the old cluster, and then roll out to the goal state.
* If the goal state matches the old cluster, roll back from the new cluster to the old cluster.
* If the goal state matches the new cluster, continue rolling out from the old cluster to the new cluster.

ryanaoleary:
> Discuss offline with the community:
> 
> * This REP should support rollback.
> * If users update the RayService CR during the upgrade process, KubeRay should first roll back and then roll out to the goal state.
> 
> Here are some thoughts on (2):
> 
> * Terminology:
>   
>   * Old cluster
>   * New cluster
>   * Goal state: the state currently defined in the RayService CR
> 
> During the upgrade process from the old cluster to the new cluster:
> 
> * If the goal state doesn't match either the old cluster or the new cluster, roll back from the new cluster to the old cluster, and then roll out to the goal state.
> * If the goal state matches the old cluster, roll back from the new cluster to the old cluster.
> * If the goal state matches the new cluster, continue rolling out from the old cluster to the new cluster.

I've updated the REP based on the community discussion, the current proposal supports the upgrade using the proposed `target_capacity`, `step_size`, and `interval` APIs and would support rollback by comparing the current RayCluster states to the RayService CR 'Goal' state. We can add more features to `UpgradeStrategy` in the future to support more advanced upgrade cases. I think the current proposal is ready for another iteration.

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: Update: based on offline discussion the above process would be changed to first scale `target_capacity` * `total_serve_replicas` replicas on the upgraded RayCluster, wait for the Ray autoscaler to scale the appropriate amount of K8s resources, and then gradually migrate traffic with Gateway at `stepSize` percent every `interval` seconds.

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: Update: I left `target_capacity` in the proposal to support a manual upgrade process, but added `stepSize` and `interval` to control a gradual traffic migration. The Ray autoscaler would be required to scale up the appropriate node resources on the ugpraded RayCluster such that it could handle `target_capacity` * `total_serve_replicas` and thus `target_capacity` * `total_requests`.

edoakes (approved):
LGTM!

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,268 @@
...
+4. KubeRay controller checks the [Serve config files](https://docs.ray.io/en/releases-2.7.0/serve/production-guide/config.html#serve-config-files-serve-build) and creates a Gateway. Specifically, the controller validates the `http_options.port` and `grpc_options.port` to use for the respective listeners. If `grpc_servicer_functions` are missing, the GRPc listener can be ommitted.
+
+Example Gateway:
+```sh
+apiVersion: gateway.networking.k8s.io/v1
+kind: Gateway
+metadata:
+  name: <ray-service-name>-gateway
+spec:
+  gatewayClassName: <ray-service-name>-gateway
```
kevin85421, line 156: `GatewayClass` is a cluster-scoped resource, and it should be created once and reused rather than creating one dynamically for each CR IIUC. Maybe we should expose it in RayService CRD.

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,268 @@
...
+## Stewardship
+
+### Shepherd of the Proposal
+- @kevin85421
+
+## Design and Architecture
+
+### Proposed API
+
+[target_capacity](https://github.com/ray-project/ray/blob/2ae9aa7e3b198ca3dbe5d65f8077e38d537dbe11/python/ray/serve/schema.py#L38) (int32 [0, 100]) - [Implemented in Ray 2.9](https://github.com/ray-project/ray/commit/86e0bc938989e28ada38faf25b75f717f5c81ed3)
```
kevin85421, line 24: I think we can remove `target_capacity` from CR spec for now. At first, this is used to enable users to manually trigger each step, but we currently have `step_size` and `interval`. 

In addition, we plan to support rollback, which means we must support automatic upgrades because the current CR spec cannot express both the target state of the old RayCluster and the target state of the newly upgraded cluster.

kevin85421 (commented):
* RayService Status: I expected some new fields are necessary (e.g. `target_capacity` for both the old RayCluster and the new RayCluster). 

* Scope of alpha:
  * Support rollout and rollback
  * Support only HTTPRoute
  * Support only automatic upgrades
  * API changes for RayService spec: only `step_size_percent`, `interval_s`, `RayServiceUpgradeType`, `max_surge_percent` and gateway class.

Try to make the scope of alpha as simple as possible.

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,268 @@
...
+
+step_size (int32 [0, 100])
+- `step_size` represents the percentage of traffic to transfer to the upgraded cluster every `interval` seconds. `step_size` is therefore the increase in the `HTTPRoute` and `GRPcRoute` weight associated with the upgraded cluster endpoint. The percentage of traffic routed to the upgraded RayCluster will increase by `step_size` until `target_capacity` is reached. At the same time, the percent of traffic routed to the old RayCluster will decrease by `step_size` every `interval` seconds.
+- Required value if `IncrementalCluster` type is set
+
+interval (int32 [0, ...])
+- `interval` represents the number of seconds for the controller to wait between increasing the percentage of traffic routed to the upgraded RayCluster by `step_size` percent.
+- Required value if `IncrementalCluster` type is set
+
+upgrade_start_timeout_s (int32 [0, ...])
```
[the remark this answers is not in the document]
ryanaoleary: Yeah I was thinking that after `UpgradeStartTimeout` the controller would switch traffic even if the applications are not ready, since for a large number of non-declaratively created apps at least 1 might always be updating. What do you mean by update Ray Serve observability?

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,268 @@
...
+  - GRPcRoute - defines how GRPc traffic is routed to the two RayCluster endpoints
+
+
+### Example Upgrade Process
+
+1. K8s cluster admin installs Gateway CRDs and Gateway controller that is compatible with their infrastructure backend.
+
+2. A user creates a RayService with a `RayServiceUpgradeType` of `IncrementalCluster`.
+
+3. The KubeRay controller validates the values of `targetCapacity`, `stepSize`, and `interval` and then checks whether the Gateway CRDs and controller exists using the Kubernetes python client. If the controller doesn't exist, defaults to a blue/green deployment, since incremental upgrade requires that the standard Gateway APIs be available.
```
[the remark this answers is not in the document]
ryanaoleary: > The default strategy should always be blue/green deployment.

I think that's currently how the proposal implements it.

> The KubeRay operator doesn't need to check whether the Gateway CRD or controller exists. We only need to ensure that the KubeRay operator can still run correctly on a Kubernetes cluster without the Gateway CRD or controller installed.

By python client I meant something like [client-go](https://github.com/kubernetes/client-go) which we could use to check if the Gateway CRDs are installed in cluster, since that's managed by the K8s cluster admin and not the controller. I was thinking that if `IncrementalUpgrade` is selected, the controller should check if the required CRDs exist before attempting to create the Gateway and routes. If the CRDs aren't installed and `IncrementalUpgrade` is selected, it'd still default to a blue/green deployment. It might not be necessary to check though (i.e. if we error while attempting to create the Gateway we can still just default to blue/green).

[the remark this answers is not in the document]
ryanaoleary: I'll update the REP to just say something like, "If we fail to create the Gateway objects, the controller defaults to a blue/green deployment and continues".

reps/2024-12-4-ray-service-incr-upgrade.md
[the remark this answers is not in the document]
kevin85421: Ray Serve should expose APIs for users to define when the Ray Serve applications are ready to serve requests.

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,268 @@
...
+ - name: grpc // can be ommitted if grpc_servicer_functions is empty
+    protocol: GRPc
+    port: grpc_options.port // defaults to 9000
+```
+
+5. The KubeRay controller requires the Ray autoscaler (resource level) to be enabled in order to support a gradual traffic migration that scales the upgraded cluster by `stepSize` to serve `target_capacity` traffic routed through Gateway. We could validate that autoscaling is enabled in the previous step and accept/reject the RayService CR accordingly. Alternatively, the controller could flip the `enableInTreeAutoscaling` flag to enable node resource autoscaling and add some default values based on the Serve config.
+
+6. KubeRay creates the upgraded RayCluster with the Ray autoscaler enabled and `target_capacity` * `total_serve_replicas` replicas assigned to it. The Ray autoscaler will scale up the appropriate number of worker group replicas until the Serve replicas on the upgraded cluster are scheduled and 'Ready'. Once the new RayCluster is ready, the KubeRay controller creates `HTTPRoute`s and `GRPcRoute`s with settings according to the serve config. The `weight` associated with the upgraded cluster endpoint should start at 0.
```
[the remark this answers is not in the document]
ryanaoleary: > when will a RayService have more than 1 HTTPRoute or more than 1 GRPCRoute?

I think there would ever only be 1 `HTTPRoute` and/or  `GRPCRoute`, I'll fix it in the REP to say "controller creates a HTTPRoute"

> During the upgrade process, I think the routes should already exist. Why do we need to create new routes?

Yeah the routes should already exist before the upgrade starts (i.e. when the RayCluster is created), I'll make that clearer in the proposal by splitting this into two steps.

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,268 @@
...
+
+### Example Upgrade Process
+
+1. K8s cluster admin installs Gateway CRDs and Gateway controller that is compatible with their infrastructure backend.
+
+2. A user creates a RayService with a `RayServiceUpgradeType` of `IncrementalCluster`.
+
+3. The KubeRay controller validates the values of `targetCapacity`, `stepSize`, and `interval` and then checks whether the Gateway CRDs and controller exists using the Kubernetes python client. If the controller doesn't exist, defaults to a blue/green deployment, since incremental upgrade requires that the standard Gateway APIs be available.
+
+4. KubeRay controller checks the [Serve config files](https://docs.ray.io/en/releases-2.7.0/serve/production-guide/config.html#serve-config-files-serve-build) and creates a Gateway. Specifically, the controller validates the `http_options.port` and `grpc_options.port` to use for the respective listeners. If `grpc_servicer_functions` are missing, the GRPc listener can be ommitted.
```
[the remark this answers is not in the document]
ryanaoleary: Sounds good, I'll add a note in the REP that we'll start with support for HTTP. GRPc can be added later.

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: Yeah the `GatewayClass` is managed by the k8s cluster admin and not the controller, so I think this should actually be set by the user/whoever creates the RayService. I agree that we should expose it in the RayService CRD and then retrieve the `gatewayClassName` from there.

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: So if we remove `target_capacity` (i.e. the upgraded RayCluster will gradually scale to 100% by `stepSize`), does that mean that we scale the upgraded RayCluster by `total_serve_replicas` * `stepSize` each iteration, wait for it to autoscale and become ready, and then migrate `stepSize` traffic?

reps/2024-12-4-ray-service-incr-upgrade.md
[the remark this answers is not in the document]
ryanaoleary: Oh yeah that makes sense, sounds good. So if we add that in the Ray Serve API, then there should be no need for `UpgradeStartTimeout` from the KubeRay side right?

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: The `Gateway` should be created per RayService CR, since the `listeners` can change based on the CR's spec and `containerPorts`, but yeah the `GatewayClass` should be cluster-scoped and created by the k8s cluster admin. We should expose the `gatewayClassName` name in the RayService CRD so that the controller can dynamically create the Gateway.

reps/2024-12-4-ray-service-incr-upgrade.md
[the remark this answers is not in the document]
ryanaoleary: Discussed offline, added a link to a related feature request in: 9255c12

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,268 @@
...
+)
+```
+
+Additionally, the new API fields would be added to the `RayServiceUpgradeStrategy`:
+```sh
+type RayServiceUpgradeStrategy struct {
+  // Type represents the strategy used when upgrading the RayService.
+  // Currently supports `NewCluster`, `IncrementalCluster`, and `None`.
+  Type *RayServiceUpgradeType `json:"type,omitempty"`
+  // The target capacity percentage of the upgraded RayCluster.
```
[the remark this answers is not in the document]
ryanaoleary: [3ea0b5b](https://github.com/ray-project/enhancements/pull/58/commits/3ea0b5b159fccb8c4042a4c26fe6a82ddd9dce05)

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,268 @@
...
+
+```
+type RayServiceUpgradeType string
+
+const (
+	// During upgrade, NewCluster strategy will create new upgraded cluster and switch to it when it becomes ready
+	NewCluster RayServiceUpgradeType = "NewCluster"
+	// During upgrade, IncrementalCluster strategy will create an additional, upgraded cluster to scale to
+	// `target_capacity` and route requests to both the old cluster and new cluster with Gateway.
+	IncrementalCluster RayServiceUpgradeType = "IncrementalCluster"
```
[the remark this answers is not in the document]
ryanaoleary: [3ea0b5b](https://github.com/ray-project/enhancements/pull/58/commits/3ea0b5b159fccb8c4042a4c26fe6a82ddd9dce05)

reps/2024-12-4-ray-service-incr-upgrade.md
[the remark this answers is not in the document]
ryanaoleary: [3ea0b5b](https://github.com/ray-project/enhancements/pull/58/commits/3ea0b5b159fccb8c4042a4c26fe6a82ddd9dce05)

reps/2024-12-4-ray-service-incr-upgrade.md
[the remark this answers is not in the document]
ryanaoleary: [3ea0b5b](https://github.com/ray-project/enhancements/pull/58/commits/3ea0b5b159fccb8c4042a4c26fe6a82ddd9dce05)

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: Discussed offline, added back in `max_surge_percent` to the proposal to define the increase in `target_capacity_new` during scaling operations. [3ea0b5b](https://github.com/ray-project/enhancements/pull/58/commits/3ea0b5b159fccb8c4042a4c26fe6a82ddd9dce05)

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,284 @@
...
+  // The percentage of traffic to switch to the upgraded RayCluster at a set interval after scaling by MaxSurgePercent.
+  StepSizePercent *int32 `json:"stepSizePercent"`
+  // The interval in seconds between transferring StepSize traffic from the old to new RayCluster.
+  IntervalSeconds *int32 `json:"intervalSeconds"`
+  // The name of the Gateway Class installed by the Kubernetes Cluster admin.
+  GatewayClassName string `json:"gatewayClassName"`
+}
+```
+
+We will start only supporting automatic upgrades, but to support manual upgrades we can eventually add a `target_capacity` field to the RayService spec:
```
kevin85421, line 87: Please delete this section (L87 - L93). We are not sure whether we should support manual upgrades, and the API is not sufficient for manual upgrades.

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,284 @@
...
+```sh
+  // The target capacity percentage of the upgraded RayCluster.
+  // Defaults to 100% target capacity.
+  // +kubebuilder:default:=100
+  TargetCapacity *int32 `json:"targetCapacity,omitempty"`
+```
+
+The RayService spec would then look as follows:
```
kevin85421, line 95: Is there any difference from the current CR spec? If not, please delete this section (L95 - L106).

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,284 @@
...
+```sh
+apiVersion: ray.io/v1
+kind: RayService
+metadata:
+  name: example-rayservice
+spec:
+  upgradeStrategy:
+    gatewayClassName: "cluster-gateway" // Created by K8s cluster admin
+    type: "IncrementalUpgrade"
+    maxSurgePercent: 20  // the upgraded cluster will increase target_capacity by 20% each iteration
```
kevin85421, line 118: These fields should be under `IncrementalUpgradeOptions`.

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,284 @@
...
+## Stewardship
+
+### Shepherd of the Proposal
+- @kevin85421
+
+## Design and Architecture
+
+### Proposed API
+
+[target_capacity](https://github.com/ray-project/ray/blob/2ae9aa7e3b198ca3dbe5d65f8077e38d537dbe11/python/ray/serve/schema.py#L38) (int32 [0, 100]) - [Implemented in Ray 2.9](https://github.com/ray-project/ray/commit/86e0bc938989e28ada38faf25b75f717f5c81ed3)
```
kevin85421, line 24: This section is not easy to follow. A writing suggestion: 

```
#### RayService spec

"""
Golang code snippets
"""

* `max_surge_percent`: ...
* `step_size_percent`: ...
* `interval_s`: ...
* `IncrementalUpgrade`: ...

#### RayService status

"""
Golang code snippets
"""

* `target_capacity`: ...
```

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,284 @@
...
+- Definition: The percentage of traffic routed to the RayService the upgraded cluster is expected to handle.
+- Defaults to 100
+- This field will not be added to the RayService CR yet, since we will start with support for automatic upgrades only, but can be added in later to support manually triggering each upgrade step. The RayService CRD should still expose `target_capacity_old` and `target_capacity_new` (not set by the user) fields in the RayService status to enable the controller to scale the respective RayClusters during each upgrade iteration.
```
kevin85421, line 27: I recommend adding `target_capacity` to `RayServiceStatus` instead of `target_capacity_old` and `target_capacity_new`. That is,

```go
type RayServiceStatus struct {
    ...
    TargetCapacity ...
}
```

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,284 @@
...
+  - GRPcRoute - defines how GRPc traffic is routed to the two RayCluster endpoints
+
+
+### Example Upgrade Process
+
+1. K8s cluster admin installs Gateway CRDs, Gateway Class, and Gateway controller that is compatible with their infrastructure backend.
+
+2. A user creates a RayService with a `RayServiceUpgradeType` of `IncrementalUpgrade`.
+
+3. The KubeRay controller validates the values of `maxSurgePercent`, `stepSizePercent`, and `intervalSeconds`. If the controller errors while attempting to begin the incremental upgrade, it will default to a blue/green deployment and continue.
```
kevin85421, line 161:
```suggestion
3. The KubeRay controller validates the values of `maxSurgePercent`, `stepSizePercent`, and `intervalSeconds`.
```

KubeRay should not fall back to behavior which is different from the CR spec.

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: 289b97f

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: Yeah I think that's a lot clearer, added in 289b97f.

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: Done in 289b97f

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: 289b97f

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,284 @@
...
+### Shepherd of the Proposal
+- @kevin85421
+
+## Design and Architecture
+
+### Proposed API
+
+[target_capacity](https://github.com/ray-project/ray/blob/2ae9aa7e3b198ca3dbe5d65f8077e38d537dbe11/python/ray/serve/schema.py#L38) (int32 [0, 100]) - [Implemented in Ray 2.9](https://github.com/ray-project/ray/commit/86e0bc938989e28ada38faf25b75f717f5c81ed3)
+- Definition: The percentage of traffic routed to the RayService the upgraded cluster is expected to handle.
+- Defaults to 100
```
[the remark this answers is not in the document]
ryanaoleary: Done in 289b97f

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,279 @@
...
+- Defaults to 100, which is the same as a blue/green deployment strategy
+
+step_size_percent (int32 [0, 100])
+- `step_size_percent` represents the percentage of traffic to transfer to the upgraded cluster every `interval_s` seconds. `step_size_percent` is therefore the increase in the route weight associated with the upgraded cluster endpoint. The percentage of traffic routed to the upgraded RayCluster will increase by `step_size_percent` until the upgrade is complete. At the same time, the percent of traffic routed to the old RayCluster will decrease by `step_size_percent` every `interval_s` seconds.
+- Required value if `IncrementalUpgrade` type is set
+
+interval_s (int32 [0, ...])
+- `interval_s` represents the number of seconds for the controller to wait between increasing the percentage of traffic routed to the upgraded RayCluster by `step_size_percent` percent.
+- Required value if `IncrementalUpgrade` type is set
+
```
kevin85421, line 38: gateway class name?

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,279 @@
...
+      stepSizePercent: 5  // represents 5% traffic to swap each interval seconds
+      intervalSeconds: 10 // represents 10 second migration interval
+      gatewayClassName: "cluster-gateway" // Created by K8s cluster admin
+  serveConfigV2: |
+    ...
+```
+
+#### RayService status
+
+The Ray Serve schema currently has the field `target_capacity` with the following definition:
```
kevin85421, line 104: Perhaps new fields, such as the percentage of traffic routed to each cluster, are worth adding to RayServiceStatus?

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,279 @@
...
+
+### Example Upgrade Process
+
+1. K8s cluster admin installs Gateway CRDs, Gateway Class, and Gateway controller that is compatible with their infrastructure backend.
+
+2. A user creates a RayService with a `RayServiceUpgradeType` of `IncrementalUpgrade`.
+
+3. The KubeRay controller validates the values of `maxSurgePercent`, `stepSizePercent`, and `intervalSeconds`.
+
+4. KubeRay controller checks the Serve related fields (`containerPort`, etc.) of the RayService spec and creates a Gateway object. If the controller errors when attempting to create the Gateway objects (i.e. missing CRDs, Gateway controller, etc.), it defaults to a blue/green deployment strategy and continues.
```
kevin85421, line 158: https://github.com/ray-project/enhancements/pull/58#discussion_r1969242039 somehow the comment is not deleted.

```suggestion
4. KubeRay controller checks the Serve related fields (`containerPort`, etc.) of the RayService spec and creates a Gateway object.
```

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,279 @@
...
+8. The controller should update the routes to prepare to switch traffic to the upgrading RayCluster using a `backendRef`:
+```sh
+backendRefs:
+...
+  - name: <head-svc-of-upgraded-ray-cluster>
+    weight: 0
+    port: serve_container_port
+```
+
+9. Once the upgraded RayCluster is ready (see related issue to allow users to define when Serve apps are "Ready": https://github.com/ray-project/ray/issues/50883), the KubeRay controller will increment the weight of the upgraded RayCluster `backendRef` by `stepSizePercent`, while decreasing the weight of the old RayCluster `backendRef` by `stepSizePercent` and then wait `intervalSeconds`. The controller should loop incrementing the traffic until the `weight` associated with the upgraded cluster is equal to `target_capacity_new`. Gateway API will route the specified `weight` percentage of traffic to the old and new RayClusters accordingly.
```
kevin85421, line 231: Please update all `target_capacity_new` / `target_capacity_old`

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,279 @@
...
+
+10. Once the upgraded cluster is accepting the additonal `maxSurgePercent` capacity of requests, the controller can scale down the old RayCluster by decreasing `target_capacity_old` by `maxSurgePercent` and allowing the Ray autoscaler to reduce the size of the original cluster.
```
kevin85421, line 233: > Once the upgraded cluster is accepting the additonal `maxSurgePercent` capacity of requests

This is not always true in the last iteration.

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,279 @@
...
+
+11. The controller will loop, increasing the `target_capacity` of the new RayCluster by `maxSurgePercent`, waiting for the Serve apps to become ready, and then performing steps 9 and 10. Once `target_capacity_new` of the new RayCluster is equal to 100%, the upgrade is complete and the KubeRay controller can remove the created routes and delete the old RayCluster object. The Gateway object can be retained for future updates.
```
kevin85421, line 235: > the upgrade is complete and the KubeRay controller can remove the created routes and delete the old RayCluster object

why KubeRay needs to delete routes?

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,279 @@
...
+
+To support a manual upgrade path in the future, we could add the following step:
```
kevin85421, line 237: Delete L237 - L239.

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,279 @@
...
+### Shepherd of the Proposal
+- @kevin85421
+
+## Design and Architecture
+
+### Proposed API
+
+#### RayService spec
+
+max_surge_percent (int32 [0, 100])
```
kevin85421, line 26: Please make sure the wording is consistent with the variable name in the Golang code snippets. For example, change `max_surge_percent` to `MaxSurgePercent`.

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: It could retain the routes, they just wouldn't be used and we'd need to add a new `backendRef` for the upgraded cluster if the RayService upgrades again. I think either way - i.e. KubeRay cleans up the now unused routes or KubeRay retains them and updates them in the case of another upgrade - should both work.

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,279 @@
...
+- Definition: The maximum allowed percentage increase in capacity during a scaling operation. It limits how much the new total capacity can exceed the original target capacity. This field defines the "next state" for the upgraded RayCluster. Every `interval_s` seconds, the upgraded RayCluster should increase its `target_capacity` by 100% + `max_surge_percent` - old RayCluster `target_capacity`.
+- `max_surge_percent` must adhere to `target_capacity_old` + `target_capacity_new` <= (100% + `max_surge_percent`)
+- Defaults to 100, which is the same as a blue/green deployment strategy
+
+step_size_percent (int32 [0, 100])
```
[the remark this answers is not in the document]
ryanaoleary: Yeah that's what I meant by "upgrade is complete", I'll change the wording to state that.

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: I'll change it to retain the Gateway API objects, and just update them as necessary for future upgrades.

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: We always increase `target_capacity` by `maxSurgePercent` -> migrate the traffic by `stepSizePercent` until it's equal to the new `target_capacity` -> decrease the old `target_capacity` by `maxSurgePercent`. So in the last iteration, we should decrease the old `target_capacity` to 0 and then the upgrade is complete. What part is not true for this?

ryanaoleary: I'll add a sentence about the last iteration occuring when new `target_capacity` is set to 100% and then old `target_capacity` is set to 0% after traffic is migrated.

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: I think that could be useful for observability for users, I'll add `TrafficRoutedPercent` as a field to the RayServiceStatus.

reps/2024-12-4-ray-service-incr-upgrade.md
kevin85421: > I think either way - i.e. KubeRay cleans up the now unused routes or KubeRay retains them

How does deletion work? Is it possible for requests to be dropped?

reps/2024-12-4-ray-service-incr-upgrade.md
kevin85421: For example, if `maxSurgePercent` is 30%, then the upgraded cluster receives traffic from 90% to 100%.

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: Done in [6ba9e61](https://github.com/ray-project/enhancements/pull/58/commits/6ba9e6146808e55c1978530eca9bd4bdf36075a3)

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: [6ba9e61](https://github.com/ray-project/enhancements/pull/58/commits/6ba9e6146808e55c1978530eca9bd4bdf36075a3)

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: [3c988dc](https://github.com/ray-project/enhancements/pull/58/commits/3c988dca37a8190b4986045dbcdf1af39addb57a)

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: Sorry didn't see the comment, I don't think it's possible for requests to be dropped once the upgrade is complete since it's routing 100% of traffic to one endpoint. With deletion, we'd just remove the HTTPRoute that routes requests to the containerPort, and instead route requests directly to the containerPort (the same as before the upgrade). However, to avoid re-creating routes with the same setting I changed the REP to retain the routes between upgrades and just update the endpoints accordingly.

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: [341fcac](https://github.com/ray-project/enhancements/pull/58/commits/341fcac1b23b03cdf08da56b972ddc4f4238b4ca)

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: [341fcac](https://github.com/ray-project/enhancements/pull/58/commits/341fcac1b23b03cdf08da56b972ddc4f4238b4ca)

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: [341fcac](https://github.com/ray-project/enhancements/pull/58/commits/341fcac1b23b03cdf08da56b972ddc4f4238b4ca)

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,279 @@
...
+- @kevin85421
+
+## Design and Architecture
+
+### Proposed API
+
+#### RayService spec
+
+max_surge_percent (int32 [0, 100])
+- Definition: The maximum allowed percentage increase in capacity during a scaling operation. It limits how much the new total capacity can exceed the original target capacity. This field defines the "next state" for the upgraded RayCluster. Every `interval_s` seconds, the upgraded RayCluster should increase its `target_capacity` by 100% + `max_surge_percent` - old RayCluster `target_capacity`.
```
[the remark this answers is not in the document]
ryanaoleary: Yeah that's correct, that sentence was worded wrong. I fixed it in [341fcac](https://github.com/ray-project/enhancements/pull/58/commits/341fcac1b23b03cdf08da56b972ddc4f4238b4ca).

reps/2024-12-4-ray-service-incr-upgrade.md
kevin85421: > With deletion, we'd just remove the HTTPRoute that routes requests to the containerPort, and instead route requests directly to the containerPort (the same as before the upgrade).

I am a bit confused. Can you provide more details? Something like:

* HTTPRoute hr1 points to old cluster (weight: x%) and new cluster (weight: y%).
* When ... happens, KubeRay deletes hr1 ...

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,284 @@
...
+
+10. Once the upgraded cluster is accepting the additonal `maxSurgePercent` capacity of requests (up to 100%), the controller can scale down the old RayCluster by decreasing `target_capacity` by `maxSurgePercent` and allowing the Ray autoscaler to reduce the size of the original cluster. In the last iteration, the new `target_capacity` should be increased to a maximum of 100% and the old RayCluster `target_capacity` should be decreased to 0%.
+
+11. The controller will loop, increasing the `target_capacity` of the new RayCluster by `maxSurgePercent`, waiting for the Serve apps to become ready, and then performing steps 9 and 10. Once `target_capacity` of the new RayCluster is equal to 100%, the upgrade is complete and the KubeRay controller can delete the old RayCluster object and remove its `backendRef` from the routes. The Gateway API objects can be retained for future updates.
```
kevin85421, line 244: > Once `target_capacity` of the new RayCluster is equal to 100%, the upgrade is complete and the KubeRay controller can delete the old RayCluster object and remove its `backendRef` from the routes.

The upgrade is complete only when the TargetCapacity of the new cluster is 100% and the traffic is also 100%.

reps/2024-12-4-ray-service-incr-upgrade.md
```diff
@@ -0,0 +1,284 @@
...
+- @kevin85421
+
+## Design and Architecture
+
+### Proposed API
+
+#### RayService spec
+
+MaxSurgePercent (*int32 [0, 100])
+- Definition: The maximum allowed percentage increase in capacity during a scaling operation. It limits how much the new total capacity can exceed the original target capacity. This field defines the "next state" for the upgraded RayCluster. Each iteration that `target_capacity` percent of total traffic is routed to the upgraded RayCluster, the upgraded RayCluster should increase its `target_capacity` by 100% + `MaxSurgePercent` - old RayCluster `target_capacity`.
```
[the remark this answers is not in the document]
ryanaoleary: [7845089](https://github.com/ray-project/enhancements/pull/58/commits/7845089f3cd38d9df911f598c9188d1a6073ec96)

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: [2ba2d7f](https://github.com/ray-project/enhancements/pull/58/commits/2ba2d7f3d7744b236da3e24cbf41384de6c913de)

reps/2024-12-4-ray-service-incr-upgrade.md
ryanaoleary: Done in 30034bc. The controller creates the `HTTPRoute` and configures it to point to the old RayCluster object. At the start of the upgrade, the new RayCluster is created and the controller updates the `HTTPRoute` to point to the new RayCluster too with `weight = 0` (so 0% of traffic is routed to it). The `weight` of the new RayCluster is increased throughout the upgrade until 100% of traffic is routed to it. Once `TargetCapacity` and `TrafficRoutedPercent` of the new RayCluster are both equal to 100%, the old RayCluster is deleted and the `HTTPRoute` is updated to no longer point to it.

kevin85421: I am asking about deleting HTTP route, but it doesn't matter for now because we decide to retain the route.

kevin85421 (approved):
Nice!

pcmoritz (approved):

PR #64, opened 2026-03-24
REP: Embedded Storage Backend for GCS Fault Tolerance

jhasm:
Proposes adding a RocksDB-based StoreClient implementation as an alternative to Redis for GCS fault tolerance. Instead of requiring an external Redis instance, GCS persists state to a local RocksDB
 database on a persistent volume. On head node restart, GCS reads state back from the local database and resumes — same recovery model as Redis-based FT, without the external dependency.

Motivation: Running distributed PyTorch training on Ray via KubeRay, head node restarts are inevitable (node maintenance, spot preemption, transient failures). Today this is catastrophic — all
progress since last checkpoint is lost. Embedded storage FT lets these jobs survive head node restarts without managing a separate Redis cluster.

Context: https://github.com/ray-project/ray/issues/53115#issuecomment-4104409561

Scope:
- Ray core: new RocksDbStoreClient implementing the existing StoreClient interface
- KubeRay: PVC provisioning and configuration for the embedded storage path
- Fully backward compatible — Redis-based FT remains supported, this is a new opt-in alternative

Requested reviewers: @jjyao (GCS maintainer), @kevin85421 or @andrewsykim  (KubeRay / GCS FT)

### DIFF
--- a/reps/2026-02-23-gcs-embedded-storage.md
+++ b/reps/2026-02-23-gcs-embedded-storage.md
@@ -17,7 +17,7 @@ This enhancement proposes an **embedded storage backend** (RocksDB) as an altern
 
 ### Should this change be within `ray` or outside?
 
-This change affects both `ray` (new `StoreClient` implementation in C++) and `kuberay` (PVC provisioning and configuration for the embedded storage path).
+This change affects both `ray` (new `StoreClient` implementation in C++) and `kuberay` (automatic PVC lifecycle management so users don't need to create or clean up PVCs manually).
 
 ## Stewardship
 
@@ -152,22 +152,31 @@ Operators can tune these settings via environment variables for their deployment
 |---|---|---|
 | `RAY_GCS_STORAGE_PATH` | *(required)* | Path to persistent volume for RocksDB data. |
 | `RAY_GCS_ROCKSDB_SYNC_WRITES` | `1` | fsync on every write. Set `0` for dev/test when crash safety is not needed (~5–10× write throughput improvement). |
-| `RAY_GCS_ROCKSDB_BLOCK_CACHE_MB` | `32` | Block cache size in MB. Increase for clusters with very large metadata (thousands of actors/jobs). Range: 8–256. |
-| `RAY_GCS_ROCKSDB_WRITE_BUFFER_SIZE_MB` | `16` | Per-column-family write buffer size in MB. Increase for high metadata churn. Range: 4–64. |
-| `RAY_GCS_ROCKSDB_MAX_BACKGROUND_JOBS` | `2` | Max compaction/flush threads. Increase to 4 on head nodes with spare CPU if compaction falls behind. Range: 1–8. |
-| `RAY_GCS_ROCKSDB_COMPRESSION` | `1` | LZ4 compression on L2+ levels. Set `0` to disable — saves CPU at cost of slightly more disk. |
 | `RAY_GCS_ROCKSDB_WAL_DIR` | *(same as storage path)* | Separate directory for WAL files. Useful if main storage is slow but a faster disk is available for WAL. |
+
+Additional tuning knobs (block cache size, write buffer size, background jobs, compression) are intentionally omitted from the initial release. The hardcoded defaults are sized for the GCS metadata workload (10–100 MB). If operators report specific performance issues, we can expose targeted env vars in a follow-up — starting restrictive is safer than starting permissive.
+
+##### Why Environment Variables, Not a Config File
+
+An alternative is a single `RAY_GCS_ROCKSDB_CONFIG_FILE` pointing to a RocksDB options file. We chose env vars because:
+
+- **Safety guardrails.** A raw RocksDB options file exposes ~200 settings, including ones that silently break correctness: changing `merge_operator` corrupts the job ID sequence, changing `create_if_missing` prevents startup, changing compaction style causes pathological space amplification. The env var approach acts as a curated API surface that only exposes safe-to-tune options.
+- **Ray convention.** Every Ray component is configured via environment variables (`RAY_REDIS_ADDRESS`, `RAY_GCS_SERVER_PORT`, etc.). A config file would be the only exception.
+- **Kubernetes-native.** Env vars map directly to pod spec `.env[]`. A config file requires creating a ConfigMap, mounting it, and keeping it in sync — extra operational burden for no functional benefit.
+
+If advanced tuning demand emerges in production, a config file override can be added as follow-on work with an explicit allowlist of safe options.
 
 ##### What Is NOT Configurable (and Why)
 
 The following are intentionally hardcoded because incorrect values can break correctness or crash safety:
 
-- **`create_if_missing` / `create_missing_column_families`** — Must be `true`. GCS creates the DB on first start, and new Ray versions may add column families. Exposing these invites misconfiguration that prevents startup or breaks upgrades.
-- **Compaction style and level sizing** — Level compaction with dynamic level bytes is the only correct choice for this workload. Universal compaction increases space amplification; FIFO is for TTL workloads. `target_file_size_base` and `max_bytes_for_level_base` are coupled — changing one without the other causes pathological compaction.
-- **Bloom filter bits and block size** — 10 bits/key and 16 KB blocks are tuned for point lookups on metadata. These have subtle interactions with compression ratios and cache efficiency.
-- **`prefix_extractor`** — RocksDB warns this cannot be changed after creation. GCS uses full-key lookups, not prefix scans on key substrings. Incorrect settings are unrecoverable.
+- **`create_if_missing` / `create_missing_column_families`** — Must be `true`. GCS creates the DB on first start, and new Ray versions may add column families.
+- **Compaction style and level sizing** — Level compaction with dynamic level bytes is the only correct choice for this workload. `target_file_size_base` and `max_bytes_for_level_base` are coupled — changing one without the other causes pathological compaction.
+- **Bloom filter bits and block size** — 10 bits/key and 16 KB blocks are tuned for point lookups on metadata. Subtle interactions with compression ratios and cache efficiency.
+- **`prefix_extractor`** — Cannot be changed after DB creation. Incorrect settings are unrecoverable.
 - **`merge_operator`** — Custom operator for `AsyncGetNextJobID` atomic increment. Changing it corrupts the job ID sequence.
 - **Direct I/O** — No durability benefit (doesn't apply to WAL) and can break on Kubernetes PV implementations.
+- **Memory settings** (block cache, write buffers) — Hardcoded at conservative values (32 MB cache, 16 MB × 2 write buffers per CF). These ensure bounded memory on shared head nodes without operator tuning.
 
 #### Recovery Flow
 
@@ -219,30 +228,61 @@ All changes are in `src/ray/gcs/store_client/`:
 
 ### KubeRay Changes
 
-1. **RayCluster CRD extension**
+When `gcsFaultTolerance.backend: rocksdb` is set, the KubeRay operator automatically manages the full PVC lifecycle — users don't create, mount, or clean up PVCs manually.
 
-   ```yaml
-   spec:
-     gcsFaultTolerance:
-       # Existing: Redis-based
-       # redisAddress: "redis:6379"
+**Note:** RocksDB is an embedded C++ library linked into the GCS process — it is not a sidecar or separate service. The PVC simply provides a persistent directory for GCS to write to.
 
-       # New: Embedded storage
-       backend: rocksdb
-       storage:
-         size: 10Gi
-         storageClassName: ssd   # optional, uses default SC if omitted
-   ```
+#### User-Facing Configuration
 
-2. **Operator reconciliation**
-   - When `backend: rocksdb` is set, create a PVC and mount it on the head pod at the configured path
-   - Set `RAY_GCS_STORAGE=rocksdb` and `RAY_GCS_STORAGE_PATH` env vars on the head container
-   - Ensure PVC lifecycle is tied to the RayCluster (deleted when cluster is deleted, persists across head pod restarts)
+```yaml
+apiVersion: ray.io/v1
+kind: RayCluster
+metadata:
+  name: my-ray-cluster
+spec:
+  gcsFaultTolerance:
+    backend: rocksdb
+    storage:
+      size: 1Gi                # GCS metadata is 10-100MB; 1Gi provides ample headroom
+      storageClassName: ssd    # optional, uses default SC if omitted
+```
 
-3. **Head pod restart behavior**
-   - KubeRay already restarts the head pod on failure
-   - The PVC survives pod deletion/restart by design
-   - No special handling needed beyond ensuring the PVC is mounted
+This is all the user specifies. The operator handles everything else.
+
+#### What the Operator Does
+
+When `backend: rocksdb` is set, the KubeRay operator reconciliation:
+
+1. **Creates a PVC** named `{cluster-name}-gcs-pvc` with `ReadWriteOnce` access mode, the specified size and storage class, and `ownerReferences` pointing to the RayCluster (so Kubernetes garbage collection deletes the PVC automatically when the cluster is deleted).
+2. **Mounts the PVC** on the head pod at `/data/gcs-state`.
+3. **Sets environment variables** on the head container: `RAY_GCS_STORAGE=rocksdb` and `RAY_GCS_STORAGE_PATH=/data/gcs-state`.
+
+The resulting head pod spec (generated by the operator, not written by the user):
+
+```yaml
+# Auto-generated by KubeRay operator — users do not write this
+containers:
+- name: ray-head
+  env:
+  - name: RAY_GCS_STORAGE
+    value: "rocksdb"
+  - name: RAY_GCS_STORAGE_PATH
+    value: "/data/gcs-state"
+  volumeMounts:
+  - name: gcs-storage
+    mountPath: /data/gcs-state
+volumes:
+- name: gcs-storage
+  persistentVolumeClaim:
+    claimName: my-ray-cluster-gcs-pvc
+```
+
+#### PVC Lifecycle
+
+- **Created** by the operator when the RayCluster is created with `backend: rocksdb`.
+- **Survives head pod restarts.** Kubernetes does not delete a PVC when its pod is deleted. When the head pod restarts, it re-mounts the same PVC and GCS recovers from the existing RocksDB data.
+- **Deleted automatically** when the RayCluster is deleted, via `ownerReferences` and Kubernetes garbage collection. No manual cleanup required.
+- **Stale data protection.** On startup, `RocksDbStoreClient` should validate that the RocksDB data directory belongs to this cluster (e.g., by storing and checking a cluster ID marker). If a PVC is accidentally reused with a different cluster, GCS should fail-fast rather than load stale state from a previous cluster.
 
 ## Compatibility, Deprecation, and Migration Plan
 
@@ -267,8 +307,8 @@ All changes are in `src/ray/gcs/store_client/`:
 
 ### KubeRay E2E Tests
 
-- RayCluster with `gcsFaultTolerance.backend: rocksdb` creates PVC and mounts it
-- Head pod deletion + restart recovers cluster state
+- RayCluster with `gcsFaultTolerance.backend: rocksdb` creates PVC and mounts it on head pod automatically
+- Head pod deletion + restart recovers cluster state from PVC
 - RayJob completes successfully after head pod restart mid-job
 - PVC is cleaned up when RayCluster is deleted
 
--- a/reps/2026-02-23-gcs-embedded-storage.md
+++ b/reps/2026-02-23-gcs-embedded-storage.md
@@ -1,0 +1,258 @@
+# REP: Embedded Storage Backend for GCS Fault Tolerance
+
+## Summary
+
+### General Motivation
+
+Ray's Global Control Store (GCS) is a single process on the head node that stores all cluster metadata — node membership, actor state, placement groups, job history, and resource availability. Today, GCS fault tolerance requires an **external Redis instance** as a persistence backend. While this works, it introduces operational complexity and a fragile dependency:
+
+- Redis must be provisioned, monitored, and maintained separately
+- The Redis client implementation has multiple P1 bugs: SIGSEGV on connection reset ([#53475](https://github.com/ray-project/ray/issues/53475)), worker pods restarting unexpectedly during recovery ([#52480](https://github.com/ray-project/ray/issues/52480)), silent crashes on timeouts ([#47419](https://github.com/ray-project/ray/issues/47419)), and lost job history after recovery ([#44218](https://github.com/ray-project/ray/issues/44218))
+- Redis Sentinel/Cluster topology changes cause GCS crashes ([#46982](https://github.com/ray-project/ray/issues/46982))
+- No support for Redis IAM authentication ([#49001](https://github.com/ray-project/ray/issues/49001)) or SSL certificate validation ([#41161](https://github.com/ray-project/ray/issues/41161))
+
+This enhancement proposes an **embedded storage backend** (RocksDB) as an alternative to Redis for GCS fault tolerance. Instead of writing state over the network to an external Redis instance, GCS writes to a local embedded database on a persistent volume. On head node restart, GCS reads state back from the local database and resumes — the same recovery model as Redis-based FT, but without the external dependency.
+
+**Production context:** At LinkedIn, we run distributed PyTorch training jobs on Ray via Flyte on Kubernetes. These are long-running jobs (hours to days) where head node restarts — due to node maintenance, spot preemption, or transient failures — are inevitable. Today, head node failure is catastrophic: all training progress since the last application-level checkpoint is lost, and the entire job must be restarted. GCS fault tolerance with embedded storage, combined with training checkpointing, would allow these jobs to survive head node restarts and resume from where they left off — minimizing repeated computation and eliminating manual intervention.
+
+### Should this change be within `ray` or outside?
+
+This change affects both `ray` (new `StoreClient` implementation in C++) and `kuberay` (PVC provisioning and configuration for the embedded storage path).
+
+## Stewardship
+
+- **Required Reviewers**: @jjyao (GCS maintainer), @kevin85421 (KubeRay / GCS FT integration)
+- **Shepherd of the Proposal**: TBD — requesting @jjyao or another GCS-experienced committer
+
+## Design and Architecture
+
+### Current Architecture (Redis based)
+
+```mermaid
+graph LR
+       GCS1[GCS Server<br/>on Head Node] -->|network writes| Redis[(External Redis)]
+       W1[Worker 1] --> GCS1
+       W2[Worker 2] --> GCS1
+       W3[Worker 3] --> GCS1
+```
+
+### Proposed Architecture (RocksDB based)
+
+```mermaid
+graph LR
+       GCS2[GCS Server<br/>on Head Node] -->|local writes| RDB[(RocksDB on PVC)]
+       W4[Worker 1] --> GCS2
+       W5[Worker 2] --> GCS2
+       W6[Worker 3] --> GCS2
+```
+
+GCS fault tolerance is built on a clean `StoreClient` abstraction (`src/ray/gcs/store_client/store_client.h`) with two implementations:
+
+1. **`InMemoryStoreClient`** — hash maps in GCS process memory. Used when FT is disabled. State is lost on crash.
+2. **`RedisStoreClient`** — persists state to an external Redis instance. Used when `RAY_REDIS_ADDRESS` is set.
+
+The `StoreClient` interface exposes ~9 async methods operating on a table-oriented key-value model:
+
+```cpp
+// Simplified interface
+class StoreClient {
+  virtual Status AsyncPut(table, key, data, overwrite, callback);
+  virtual Status AsyncGet(table, key, callback);
+  virtual Status AsyncGetAll(table, callback);
+  virtual Status AsyncMultiGet(table, keys, callback);
+  virtual Status AsyncDelete(table, key, callback);
+  virtual Status AsyncBatchDelete(table, keys, callback);
+  virtual Status AsyncGetNextJobID(callback);
+  virtual Status AsyncGetKeys(table, prefix, callback);
+  virtual Status AsyncExists(table, key, callback);
+};
+```
+
+GCS recovery flow today (with Redis):
+
+```mermaid
+flowchart TD
+    A[Head pod crashes] --> B[KubeRay detects pod<br/>failure]
+    B --> C[KubeRay restarts<br/>head pod]
+    C --> D[GCS process starts,<br/>connects to Redis]
+    D --> E[GCS reads all state<br/>from Redis]
+    E --> F["Workers reconnect<br/>(within gcs_rpc_server_<br/>reconnect_timeout_s)"]
+    F --> G[Cluster resumes operation]
+```
+
+### Proposed Change: RocksDB StoreClient
+
+We propose adding a third `StoreClient` implementation: **`RocksDbStoreClient`**.
+
+#### Data Model Mapping
+
+The `StoreClient` interface is table-oriented (table_name, key, value). RocksDB supports this naturally using **column families** (one per table) or **key prefixes** (`{table_name}/{key}`). We propose column families for cleaner isolation:
+
+```cpp
+// Each GCS table maps to a RocksDB column family
+// Tables: ACTOR, NODE, PLACEMENT_GROUP, JOB, WORKER, etc.
+//
+// Put("ACTOR", "actor_123", data)  →  CF["ACTOR"].Put("actor_123", data)
+// GetAll("ACTOR")                  →  CF["ACTOR"].NewIterator()
+```
+
+#### AsyncGetNextJobID
+
+The `RedisStoreClient` uses Redis `INCR` for atomic job ID generation. RocksDB provides `Merge` operators that can implement atomic increment:
+
+```cpp
+// Option 1: Read-modify-write under a mutex (simple, GCS is single-process)
+// Option 2: RocksDB Merge operator with uint64 addition
+//
+// Since GCS is a single process, a simple mutex-guarded read-modify-write
+// on a reserved key is sufficient. No distributed coordination needed.
+```
+
+#### Storage Path and Configuration
+
+```bash
+# Enable embedded storage FT
+export RAY_GCS_STORAGE=rocksdb
+export RAY_GCS_STORAGE_PATH=/data/gcs-state    # must be on a persistent volume
+
+# Existing Redis-based FT remains supported
+export RAY_GCS_STORAGE=redis                    # or: export RAY_REDIS_ADDRESS=...
+```
+
+#### Recovery Flow
+
+```mermaid
+flowchart TD
+    A[Head pod crashes] --> B[KubeRay detects pod<br/>failure, restarts<br/>head pod]
+    B --> C[PVC with RocksDB<br/>data is re-mounted<br/>on new head pod]
+    C --> D["GCS process starts,<br/>opens RocksDB at<br/>RAY_GCS_STORAGE_PATH"]
+    D --> E["GCS reads all<br/>state from local<br/>RocksDB<br/>(fast — no network)"]
+    E --> F[Workers reconnect]
+    F --> G[Cluster resumes<br/>operation]
+```
+
+#### Performance Characteristics (Expected)
+
+| Operation | Redis (network) | RocksDB (local SSD) |
+|-----------|----------------|---------------------|
+| Write latency | 0.5–2ms (network RTT) | 0.01–0.1ms (local I/O) |
+| Read latency | 0.5–2ms | 0.01–0.05ms |
+| Recovery read (full state) | Bound by network bandwidth | Bound by disk bandwidth (faster) |
+| Failure mode | Network partition, Redis crash, topology change | Disk failure (rare with cloud PVs) |
+
+#### What This Does NOT Provide
+
+This proposal provides **persistence**, not **high availability**. The distinction:
+
+- **Persistence**: GCS state survives head pod restart. There is a recovery window (seconds to minutes) during which the cluster is unavailable.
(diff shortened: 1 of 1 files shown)


reps/2026-02-23-gcs-embedded-storage.md
```diff
@@ -0,0 +1,258 @@
...
+flowchart TD
+    A[Head pod crashes] --> B[KubeRay detects pod<br/>failure]
+    B --> C[KubeRay restarts<br/>head pod]
+    C --> D[GCS process starts,<br/>connects to Redis]
+    D --> E[GCS reads all state<br/>from Redis]
+    E --> F["Workers reconnect<br/>(within gcs_rpc_server_<br/>reconnect_timeout_s)"]
+    F --> G[Cluster resumes operation]
+```
+
+### Proposed Change: RocksDB StoreClient
```
andrewsykim, line 83: Any particular reason to start with RocksDB vs other options?

jhasm: That is a good point. I did some research on comparing embedded DB options for Ray use cases, but did not include it in the REP. Let me add a comparison section.

siyuanfoundation:
I think it is helpful to provide a persisted option shipped with Ray. 

RocksDB has many tuning options. In the REP, can you elaborate on what default configurations you plan to use and what options will be configurable for Ray users?

reps/2026-02-23-gcs-embedded-storage.md
jhasm: Added a "Embedded Storage Backend Selection" section in the proposal. And happy to hear more if anyone has tried any of these candidates (or other embedded db) with ray and has any interesting observations.

RocksDb itself is proven at hyperscale, but using with Ray is a new integration here.

jhasm:
> I think it is helpful to provide a persisted option shipped with Ray.
> 
> RocksDB has many tuning options. In the REP, can you elaborate on what default configurations you plan to use and what options will be configurable for Ray users?


@siyuanfoundation  added a "RocksDB Configuration" section to the proposal. Let me know what you think.

kevin85421 (commented):
Hi @jhasm,

Thank you for the proposal—it’s a good idea. I’ve also heard that some users have complained about the operational costs of external Redis services. This proposal could make their lives easier.

> KubeRay: PVC provisioning and configuration for the embedded storage path

KubeRay exposes the full PodSpec. You can run RocksDB in a sidecar and configure the PVC in the RayCluster CR yourself. No changes are needed in KubeRay because containers in the same Pod share the same network namespace; the only required work is implementing the integration between GCS and RocksDB.

In this case, it will make the system more stable and more extensible if we want to support more databases.

kevin85421:
also cc Ray Core TL at Anyscale @edoakes

andrewsykim:
yeah in general I agree that from KubeRay's perspective, we can defer adding any operator logic that is specific to storage solutions. This is already true for Redis, where KubeRay is not responsible for provisioning or operating Redis in any way. However, we should ensure that any storage solution can be layered nicely on top of RayCluster via user configs.

rueian:
I think a native RocksDB integration is great to have. I am not sure when this will be implemented, but if you are in a hurry:

>  requiring an external Redis instance

The Redis instance for GCS FT doesn't have to be external. It can be a local Redis instance persisted with PV.

> Today this is catastrophic — all progress since last checkpoint is lost. 

This is true for Redis by default, but you can have Redis AOF enabled or even replace it with a Kvrocks instead, which talks the Redis protocol but is backed by RocksDB.

siyuanfoundation:
> Hi @jhasm,
> 
> Thank you for the proposal—it’s a good idea. I’ve also heard that some users have complained about the operational costs of external Redis services. This proposal could make their lives easier.
> 
> > KubeRay: PVC provisioning and configuration for the embedded storage path
> 
> KubeRay exposes the full PodSpec. You can run RocksDB in a sidecar and configure the PVC in the RayCluster CR yourself. No changes are needed in KubeRay because containers in the same Pod share the same network namespace; the only required work is implementing the integration between GCS and RocksDB.
> 
> In this case, it will make the system more stable and more extensible if we want to support more databases.

I think one major benefit of RocksDB is it is an embedded DB, so it can be shipped with the Ray binary. If we just run RocksDB in a sidecar, in theory you can also run Redis with a PVC in a sidecar as well,  that seems to defeat the purpose of making backup with RocksDB easier to use than Redis. 
Effort-wise, integrating RocksDB through sidecar requires similar dev and maintenance work to wire the interface of GCS with RocksDB, so it is not really less work than embedding it. The downside of embedding it is increasing the Ray binary size by ~30M, and this extra dependency on RocksDB.

reps/2026-02-23-gcs-embedded-storage.md
```diff
@@ -0,0 +1,375 @@
...
+
+| Env Var | Default | Description |
+|---|---|---|
+| `RAY_GCS_STORAGE_PATH` | *(required)* | Path to persistent volume for RocksDB data. |
+| `RAY_GCS_ROCKSDB_SYNC_WRITES` | `1` | fsync on every write. Set `0` for dev/test when crash safety is not needed (~5–10× write throughput improvement). |
+| `RAY_GCS_ROCKSDB_BLOCK_CACHE_MB` | `32` | Block cache size in MB. Increase for clusters with very large metadata (thousands of actors/jobs). Range: 8–256. |
+| `RAY_GCS_ROCKSDB_WRITE_BUFFER_SIZE_MB` | `16` | Per-column-family write buffer size in MB. Increase for high metadata churn. Range: 4–64. |
+| `RAY_GCS_ROCKSDB_MAX_BACKGROUND_JOBS` | `2` | Max compaction/flush threads. Increase to 4 on head nodes with spare CPU if compaction falls behind. Range: 1–8. |
+| `RAY_GCS_ROCKSDB_COMPRESSION` | `1` | LZ4 compression on L2+ levels. Set `0` to disable — saves CPU at cost of slightly more disk. |
+| `RAY_GCS_ROCKSDB_WAL_DIR` | *(same as storage path)* | Separate directory for WAL files. Useful if main storage is slow but a faster disk is available for WAL. |
```
siyuanfoundation: How about adding a single ROCKSDB_CONFIG_FILE to avoid all these extra flags?

andrewsykim:
> If we just run RocksDB in a sidecar, in theory you can also run Redis with a PVC in a sidecar as well, that seems to defeat the purpose of making backup with RocksDB easier to use than Redis.

For the record, this is a viable pattern that has been tested before and we should definitely weight it against supporting new storage types. If you can run Redis as a sidecar and just store snapshots in shared storage (like GCSFuse) and restore from snapshot as fast as (or close to) something like RocksDB, then maybe the effort is not worthwhile. Would be good to confirm that new storage types can restore state significantly faster than Redis can restore from snapshots.

andrewsykim:
Btw I created this example as a reference going forward for running Redis as a sidecar https://github.com/ray-project/kuberay/pull/4657

jhasm:
> > If we just run RocksDB in a sidecar, in theory you can also run Redis with a PVC in a sidecar as well, that seems to defeat the purpose of making backup with RocksDB easier to use than Redis.
> 
> For the record, this is a viable pattern that has been tested before and we should definitely weight it against supporting new storage types. If you can run Redis as a sidecar and just store snapshots in shared storage (like GCSFuse) and restore from snapshot as fast as (or close to) something like RocksDB, then maybe the effort is not worthwhile. Would be good to confirm that new storage types can restore state significantly faster than Redis can restore from snapshots.

Hey @andrewsykim,
I think I agree with @siyuanfoundation's argument about embedded db, it is more about the tight integration and UX then the restore speed. The speed with embedded db in this case is a side product, IMO. 

We can run Redis in a sidecar and that is a good viable option right now, but that is technically not same as having the db running in the same process. Containers can restart independently and can technically can have a race condition in being available. Redis runs as a separate service, while Rocksdb will run in the same process as GCS. From speed perspective, this is as fast as it can get, as most of the data will be in memory for GCS use case. But I still think that should not be our primary motivation, since speed is not a bottleneck yet.

Same argument applies to using KVRocks, which support Redis protocol over RocksDb, and runs as a service. But this option has one more downside, we have to keep testing the KVRocks integration with every Ray version upgrade, because that is not tested as a part of Ray release.

I am not saying the alternatives brought up here are not good options, they are good. The embedded db options provides a tighter integration and "batteries included" experience. And this can also potentially be a stepping stone for running GCS in replicated mode, but that discussion is out of scope for this proposal.

For the size of the Ray binary, @siyuanfoundation do you think we can make it optional that can be installed with "ray[rocksdb]" ?

siyuanfoundation:
> For the size of the Ray binary, @siyuanfoundation do you think we can make it optional that can be installed with "ray[rocksdb]" ?

While it is doable to have a separate release of "ray[rocksdb]", maintaining a separate release is probably more costly than 30M binary size increase

reps/2026-02-23-gcs-embedded-storage.md
jhasm: I think using the envvars is a safer option than supporting a file:
 - A raw RocksDB options file exposes ~200 settings including ones that break correctness (merge_operator, create_if_missing, compaction style). The env var approach acts as a curated, safe API surface.
 - Env vars are Ray's standard configuration pattern and Kubernetes-native (no ConfigMap needed).
 - If advanced tuning demand emerges, a config file override can be added in a follow-up — but starting
 restrictive is safer than starting permissive.

But I also see that 7 might be too many options without a strong reason to expose them. So I propose that we reduce the user options to 3:
   - RAY_GCS_STORAGE_PATH (required — where to store data)
   - RAY_GCS_ROCKSDB_SYNC_WRITES (the only setting with a meaningful dev/prod tradeoff)
   - RAY_GCS_ROCKSDB_WAL_DIR (operational — separate WAL to faster disk)
 - Remove BLOCK_CACHE_MB, WRITE_BUFFER_SIZE_MB, MAX_BACKGROUND_JOBS, COMPRESSION, these are premature. We can add them later if users report actual need.

jhasm:
> Hi @jhasm,
> 
> Thank you for the proposal—it’s a good idea. I’ve also heard that some users have complained about the operational costs of external Redis services. This proposal could make their lives easier.
> 
> > KubeRay: PVC provisioning and configuration for the embedded storage path
> 
> KubeRay exposes the full PodSpec. You can run RocksDB in a sidecar and configure the PVC in the RayCluster CR yourself. No changes are needed in KubeRay because containers in the same Pod share the same network namespace; the only required work is implementing the integration between GCS and RocksDB.
> 
> In this case, it will make the system more stable and more extensible if we want to support more databases.

Hey @kevin85421 ,
Good point about the full pod spec support in kuberay, that simplifies things. Let me update the proposal to reflect that, we probably don't need any changes in Kuberay operator.

The sidecar part is not applicable in this case though, because the RocksDb will run inside the GCS process (embedded db) not as a separate service/process like Redis.

jhasm:
> > Hi @jhasm,
> > Thank you for the proposal—it’s a good idea. I’ve also heard that some users have complained about the operational costs of external Redis services. This proposal could make their lives easier.
> > > KubeRay: PVC provisioning and configuration for the embedded storage path
> > 
> > 
> > KubeRay exposes the full PodSpec. You can run RocksDB in a sidecar and configure the PVC in the RayCluster CR yourself. No changes are needed in KubeRay because containers in the same Pod share the same network namespace; the only required work is implementing the integration between GCS and RocksDB.
> > In this case, it will make the system more stable and more extensible if we want to support more databases.
> 
> Hey @kevin85421 , Good point about the full pod spec support in kuberay, that simplifies things. Let me update the proposal to reflect that, we probably don't need any changes in Kuberay operator.
> 
> The sidecar part is not applicable in this case though, because the RocksDb will run inside the GCS process (embedded db) not as a separate service/process like Redis.

One question @kevin85421  and @andrewsykim: 
If users manually configure PVC in the pod spec, who takes care of clean up? IIUC we have following options for PVC management,

1. Kuberay creates PVC with RayCluster and deletes it. Most user friendly, but needs a change in Kuberay.
2. Use Kubernetes `ownerReference` to make the RayCluster the owner of the PVC. But this requires getting the RayCluster UID and setting that on the PVC. So this needs to be done after the RayCluster creation and needs to be orchestrated by something (deployment script or orchestrator). Not user friendly.
3. Users manually manage PVC lifecycle. A super set of option-2 and not user friendly.

I feel like we should go with option-1 if we want to help adoption of this feature.

edoakes:
> Motivation: Running distributed PyTorch training on Ray via KubeRay, head node restarts are inevitable (node maintenance, spot preemption, transient failures). Today this is catastrophic — all progress since last checkpoint is lost. Embedded storage FT lets these jobs survive head node restarts without managing a separate Redis cluster.

I want to point out that supporting GCS fault tolerance alone does not solve this problem. The driver program will still be a single point of failure. In order to address that, we'd need a separate enhancement -- something like a first-class "checkpointable driver" concept that can recover its state while its downstream tasks & actors continue to run.

Reducing the barrier to entry for GCS fault tolerance is still a useful step in the right direction though.

jhasm:
> > Motivation: Running distributed PyTorch training on Ray via KubeRay, head node restarts are inevitable (node maintenance, spot preemption, transient failures). Today this is catastrophic — all progress since last checkpoint is lost. Embedded storage FT lets these jobs survive head node restarts without managing a separate Redis cluster.
> 
> I want to point out that supporting GCS fault tolerance alone does not solve this problem. The driver program will still be a single point of failure. In order to address that, we'd need a separate enhancement -- something like a first-class "checkpointable driver" concept that can recover its state while its downstream tasks & actors continue to run.
> 
> Reducing the barrier to entry for GCS fault tolerance is still a useful step in the right direction though.

I agree, this proposal alone is not sufficient, it is one part of the whole solution.

The driver may not survive this disruption, and there are two types of use cases: (1) RayJob CR: driver runs on a separate submitter pod (2) driver runs on the head node. And both of these cases might need a different solution.

The other piece of the puzzle is checkpointing in the workload (e.g. training job), which will need a whole new discussion, if we were to extend resilience in that direction as well.

I can update the REP to reflect this limitation.

andrewsykim:
> The driver may not survive this disruption, and there are two types of use cases: (1) RayJob CR: driver runs on a separate submitter pod (2) driver runs on the head node. And both of these cases might need a different solution.

We _might_ be able to address most use-cases at the KubeRay RayJob submitter level. If we assume that in the future, GCS fail over can happen in O(seconds) with embedded storage and GCS replication, and the workload itself does checkpointing during training, then KubeRay simply needs to resubmit a job right after GCS fail over happens to effectively resume from last checkpoint. It's not as elegant as the first-class checkpointable driver solution that Edward mentioned, but it might be good enough for a lot of cases.

andrewsykim:
@siyuanfoundation @YoyinZyc I know you did some investigation on GCS storage backends, can you weight in on what storage backend you would recommend here? Does RocksDB seem like a good starting point?

YoyinZyc:
Thanks for the proposal @jhasm 
I just have some high level questions:
1. As kuberay change is in the scope, does it mean it only works or suggests for ray on k8s? If not, what is the behavior for ray on other platforms?
2. I noticed in the above conversation, there are some discussion about the downside of embedded db including the image size. Is there any other limitations? As it is an opt-in feature, it would be nice to have a best practice guidance about the choice between RocksDB and Redis.
3. OOM is the major reason that kills GCS. With this proposal, is it going to make this condition better, worse or it is not related?

> GCS metadata is small (10–100 MB across ~10 column families: ACTOR, NODE, JOB, PLACEMENT_GROUP, WORKER, etc.) with moderate write throughput and read-heavy recovery. The configuration priorities are crash safety and fast recovery, not write throughput.

I wonder where does this data come from? Is it possible the GCS size goes beyond 100MB for a large long-running Ray cluster?

> GCS acknowledges state changes to callers. Without fsync, a crash can lose acknowledged writes — violating the FT durability contract. On SSD, fsync adds ~0.1–0.5ms per write, acceptable for GCS's moderate write rate.

Is there a separated thread handling the write or it is on main thread? If it is async write, we still cannot ensure the strong data consistency between memory and db even with the fsync enablement?

jhasm:
Thanks @YoyinZyc for the thoughtful questions. I've pushed an update to the REP that addresses all of them inline; summaries below.

1. Does this only work on Kubernetes?

No — the Ray core change (new RocksDbStoreClient) is platform-agnostic. Anywhere a persistent local directory is available (bare metal, VMs, cloud block volumes, on-prem NAS, other orchestrators) a user can set RAY_GCS_STORAGE=rocksdb and RAY_GCS_STORAGE_PATH=/some/persistent/path and it works. The KubeRay piece only automates PVC lifecycle so k8s users don't have to wire mounts manually; equivalent integrations for Nomad / Slurm / ECS would be orchestrator-specific follow-on work.

I added a "Deployment on Non-Kubernetes Platforms" subsection with concrete examples.

2. Other limitations + when to choose RocksDB vs Redis?

Beyond image size, the main trade-offs of the embedded backend are:

- Node-local coupling — state is tied to the PV(C); with ReadWriteOnce volumes, the head pod must be re-schedulable to a node that can attach the same volume. Redis is external and decouples state from head-node locality.
- External inspectability — Redis can be inspected with redis-cli from anywhere; RocksDB state lives on the head pod and needs tooling to inspect offline.
- Memory overhead on the head pod — ~20–50 MB baseline for RocksDB (block cache + write buffers) lives inside the GCS process; Redis moves that memory out-of-process.
- Shared-nothing — single writer only, no natural story for multiple Ray clusters sharing backing state.

I added a "When to Use Embedded Storage vs Redis" table in the REP with rough guidance. TL;DR:
- Embedded (RocksDB) for new deployments (especially KubeRay) that want fewer moving parts and lower write latency.
- Redis when you already operate managed Redis at high quality, need external inspection / cross-cluster sharing, or head-pod memory is very tight.

3. Will this help, hurt, or not affect GCS OOMs?

Largely orthogonal, and slightly worse in absolute memory. OOMs today are driven by GCS's in-memory working set (actor / task / node tables, task lineage) — the storage backend doesn't change how much state GCS keeps resident. RocksDB adds ~20–50 MB of constant overhead, so strictly it makes head-pod memory pressure modestly worse, not better.

That said, embedded durable storage does unlock a future optimization: GCS could page cold state out of RAM and fault it back from RocksDB on demand (analogous to a database buffer pool). That's a genuine path to making GCS OOM-resistant, but it's out of scope for this REP — called out under follow-on work. Added an "Impact on GCS OOM" subsection to make this explicit.

4. Where does the 10–100 MB figure come from? Could it grow?

It's a typical operating-size estimate for a steady-state cluster (hundreds of actors, dozens of nodes, moderate job history), not a hard ceiling. Long-lived clusters with accumulating job history, placement-group records, and dead-actor entries can absolutely exceed 100 MB — and this is not unique to RocksDB; the same unbounded growth exists with Redis-backed FT today. "State compaction and TTL" is listed as follow-on work because it's a prerequisite for very long-running clusters regardless of backend. I softened the REP language accordingly.

5. Async writes + fsync — is consistency actually guaranteed?

Yes. Writes are executed synchronously by the storage layer even though the StoreClient API is async. The plan is: AsyncPut dispatches to an I/O worker thread, calls rocksdb::Put with WriteOptions::sync = true (which fsyncs the WAL), and invokes the callback only after fsync completes. GCS then sends the RPC reply inside that callback — so the caller sees an ack only after the write is durable on disk. There's no separate in-memory GCS cache that can diverge from RocksDB; the store client is authoritative.

The "async" here refers to thread decoupling (don't block the GCS event loop on I/O), not fire-and-forget semantics. Added a "Durability and Consistency Semantics" subsection in the REP making this explicit — it was a fair question to ask from the API signature alone.

reps/2026-02-23-gcs-embedded-storage.md
```diff
@@ -0,0 +1,477 @@
...
+
+#### User-Facing Configuration
+
+```yaml
+apiVersion: ray.io/v1
+kind: RayCluster
+metadata:
+  name: my-ray-cluster
+spec:
+  gcsFaultTolerance:
```
rueian:
```suggestion
  gcsFaultToleranceOptions:
```

reps/2026-02-23-gcs-embedded-storage.md
```diff
@@ -0,0 +1,477 @@
...
+    backend: rocksdb
+    storage:
```
rueian, line 354: should also allow a predefined PVC.

reps/2026-02-23-gcs-embedded-storage.md
```diff
@@ -0,0 +1,477 @@
...
+      size: 1Gi                # GCS metadata is 10-100MB; 1Gi provides ample headroom
+      storageClassName: ssd    # optional, uses default SC if omitted
```
rueian: should we also allow accessModes?

rueian: should we also allow changing the RAY_GCS_STORAGE_PATH?

reps/2026-02-23-gcs-embedded-storage.md
rueian: Why would users want to disable RAY_GCS_ROCKSDB_SYNC_WRITES? Probably just remove it as well.

reps/2026-02-23-gcs-embedded-storage.md
jhasm: Good catch — `gcsFaultToleranceOptions` is the existing field on KubeRay's RayCluster CRD, so the REP should extend that struct rather than introduce a new one. Updated the YAML, the "What the Operator Does" section, the head pod spec example, and the E2E test list. Also added a comment in the YAML noting the field extends the existing struct.

reps/2026-02-23-gcs-embedded-storage.md
jhasm: Agreed — added existingClaim under storage as a mutually-exclusive alternative to size / storageClassName / accessModes. When set, the operator consumes the PVC as-is and never creates or deletes it (no ownerReferences); the user owns its lifecycle. The operator-managed path remains the default for the simple case. Also called out in the PVC Lifecycle section that stale-data protection matters most for BYO-PVC and shared-subdir scenarios.

reps/2026-02-23-gcs-embedded-storage.md
jhasm: Done — added accessModes (default [ReadWriteOnce]) under storage. The "Volume Topology and Failover Scope" section already discussed RWX/RWOP tradeoffs, so the knob was needed to actually realize those topologies via the operator-managed PVC. Wired through to "What the Operator Does" so the operator passes it through on PVC creation.

---

On VM / bare-metal deployments RAY_GCS_STORAGE_PATH is already the primary user-facing env var (REP §"Deployment on Non-Kubernetes Platforms") — users pick whatever path they want, no operator wrapping it. Inside KubeRay, the operator sets both the volumeMount path and that env var to the same value, so the operationally meaningful flexibility for shared-storage scenarios (NFS / Vast preallocated dirs, multi-tenant exports) is subPath, not the absolute container path. Added storage.subPath to the schema; the operator applies it as volumeMounts[].subPath so each cluster lives under its preallocated subdirectory of the shared export. Left the absolute mount path internal.

reps/2026-02-23-gcs-embedded-storage.md
jhasm: Fair point — agreed. Anyone disabling fsync in production silently breaks the FT durability contract this REP exists to provide; that's a footgun, not a feature. Removed RAY_GCS_ROCKSDB_SYNC_WRITES from the user-configurable env vars table and added WriteOptions::sync to the "What Is NOT Configurable" list with the rationale. Tests that need non-sync mode use an internal-only fixture flag rather than a public env var. Public env-var surface is now just RAY_GCS_STORAGE_PATH (required) and RAY_GCS_ROCKSDB_WAL_DIR (optional).

rueian (approved):

andrewsykim:
We met today to discuss next steps for this REP. As there are some unknowns around GCS stability and performance using disk-based storage, we're going to build a small prototype using RocksDB as a sanity check before merging this PR

reps/2026-02-23-gcs-embedded-storage.md
```diff
@@ -0,0 +1,488 @@
...
+|---|---|---|
+| **Cluster metadata** | Preserved in RocksDB, restored on restart | **This REP** |
+| **Worker reconnection** | Workers reconnect within `gcs_rpc_server_reconnect_timeout_s` | Existing GCS FT mechanism |
+| **Driver/submitter pod** | Survives if running in a separate pod (RayJob), but pending RPCs may timeout and crash the driver process | **Needs separate work** — RayJob controller and driver reconnection logic |
+| **Distributed training (NCCL)** | NCCL training group breaks — NCCL has no reconnection. Training must restart from last checkpoint | **Application responsibility** — Ray Train + periodic checkpointing (`TorchTrainer.restore()`) |
+| **High availability** | GCS is unavailable during restart (seconds to minutes) | **Separate effort** — active/standby GCS with Raft or WAL replication |
+
+For the distributed training use case that motivates this REP: GCS persistence eliminates the need to reprovision the entire cluster after a head node failure. Combined with training checkpointing, this reduces the blast radius from "restart a 24-hour job from hour 0" to "restart training from the last checkpoint while the cluster recovers in place."
```
andrewsykim: We should make it clearer that this doesn't improve resilence for training use-cases at all, but would open the door for us to consider "detached jobs" or similar in the future, especially if using RocksDB reduces restart latency significantly compared to Redis.

reps/2026-02-23-gcs-embedded-storage.md
```diff
@@ -0,0 +1,488 @@
...
+- **Bare metal / VM** — attach a local SSD or block volume, mount at a stable path, set `RAY_GCS_STORAGE=rocksdb` and `RAY_GCS_STORAGE_PATH=/path/to/mount`.
+- **Cloud VMs (EC2, GCE, Azure)** — mount an attached volume (EBS, Persistent Disk, Managed Disk) at the configured path.
+- **On-prem NAS / SAN** — any stable POSIX filesystem mount works, with the usual caveats about NFS and fsync semantics.
+- **Other orchestrators (Nomad, Slurm, ECS, etc.)** — the env-var configuration surface is the same; automating volume lifecycle (equivalent to the KubeRay PVC work below) would be orchestrator-specific follow-on work.
+
+The KubeRay changes described below automate volume lifecycle for Kubernetes. Equivalent integrations for other orchestrators are out of scope for this REP.
+
+### KubeRay Changes
+
+When `gcsFaultToleranceOptions.backend: rocksdb` is set, the KubeRay operator automatically manages the full PVC lifecycle — users don't create, mount, or clean up PVCs manually.
```
andrewsykim, line 340: I think for the first iteration, we want to put the onus on the user to configure the PV / PVC mounts. This is mainly because PVs end up looking slightly different across cloud providers and I don't think we need to automate this yet. However, we still want `gcsFaultToleranceOptions.backend: rocksdb` to automatically set the correct env vars on the head pod

andrewsykim: On second thought, maybe this is fine as it will be mainly the storage class that is different and we are providing a way to override that.

reps/2026-02-23-gcs-embedded-storage.md
```diff
@@ -0,0 +1,488 @@
...
+
+This is all the user specifies. The operator handles everything else.
+
+#### What the Operator Does
+
+When `backend: rocksdb` is set, the KubeRay operator reconciliation:
+
+1. **Provisions storage:**
+   - If `existingClaim` is set, the operator uses that PVC as-is. It does not create or delete the PVC, and does not set `ownerReferences` — the user owns the PVC's lifecycle.
+   - Otherwise, the operator **creates a PVC** named `{cluster-name}-gcs-pvc` with the specified `accessModes` (default `[ReadWriteOnce]`), `size`, and `storageClassName`, and `ownerReferences` pointing to the RayCluster (so Kubernetes garbage collection deletes the PVC automatically when the cluster is deleted).
```
andrewsykim, line 373: future note for KubeRay implementation: the owner ref may need to be RayService so we don't delete the PVC on upgrade

reps/2026-02-23-gcs-embedded-storage.md
```diff
@@ -0,0 +1,488 @@
...
+- **Volume provider topology:**
+  - **Cloud block storage** (EBS, GCE Persistent Disk, Azure Managed Disk) is *zone-scoped*: the volume can re-attach to any node in the same zone but not across zones. This covers the common failure modes (node reboot, pod eviction, spot preemption within the zone).
+  - **Regional disks** (GCE regional PD, Azure ZRS) span two zones, enabling failover across one zone boundary.
+  - **RWX filesystems** (EFS, CephFS, NFS, Azure Files) have no locality restriction but typically trade off latency and per-op cost.
+  - **Local PV / hostPath** is truly node-pinned — unsuitable for this feature because it defeats the recovery model.
+
+**Practical takeaway:** on default KubeRay with cloud block storage, the head pod survives node-level failures within a zone, which covers the typical disruption modes driving this REP. Cross-zone head-node failure requires a regional disk, an RWX volume, or out-of-band backup. Redis, being external, is insulated from head-pod volume topology entirely.
+
+### Ray Core Changes
```
andrewsykim, line 309: @jhasm can you add a note on binary size change for adding rocksdb dependencies?

reps/2026-02-23-gcs-embedded-storage.md
jhasm: Good point — rewrote the paragraph (commit 2810429) to be explicit that this REP alone does **not** improve in-flight training resilience: NCCL still breaks, the driver still restarts, training still resumes from the last app-level checkpoint regardless of backend. Reframed the win as (a) faster cluster recovery than Redis (POC measures sub-100 ms storage-layer cold open at 10k entries) and (b) a foundation for first-class "detached jobs" / checkpointable drivers — cross-referencing @edoakes's earlier comment and Follow-on Work item 1. End-to-end training resilience requires all three pieces: GCS persistence (this REP), driver resilience (follow-on), and application checkpointing (workload responsibility).

reps/2026-02-23-gcs-embedded-storage.md
jhasm: Agreed — the current design already aligns with this. Two paths are supported in the REP:

- **Operator-managed PVC** (default, simple case): user sets `size` plus optional `storageClassName` / `accessModes`; operator handles creation and lifecycle. Cloud-provider differences are absorbed by `storageClassName`.
- **`existingClaim`** (BYO PVC): user owns the PV/PVC entirely (provisioning, sizing, cleanup, ownerRefs); operator only consumes the claim. This is the escape hatch for users who need full control over PV config across cloud providers.

In both paths, `backend: rocksdb` always auto-sets `RAY_GCS_STORAGE` and `RAY_GCS_STORAGE_PATH` on the head pod — only step 1 (PVC creation) is skipped on the `existingClaim` path. Added a one-line clarification under "What the Operator Does" in commit 2810429 to make the env-var-injection-always-runs invariant explicit.

reps/2026-02-23-gcs-embedded-storage.md
jhasm: Good catch — added an explicit "RayService caveat" bullet under PVC Lifecycle in commit 2810429: when the RayCluster is owned by a RayService, the operator should set `ownerReferences` on the PVC to the **RayService**, not the RayCluster, so that a zero-downtime upgrade (RayService creates a new RayCluster and deletes the old one) doesn't garbage-collect the GCS PVC and discard state mid-upgrade. Flagged as an implementation note for the KubeRay change rather than a user-configurable option.

reps/2026-02-23-gcs-embedded-storage.md
jhasm: Added a "Binary Size Impact" section in commit 2810429 with POC-measured numbers (from ray-project/ray#63032):

| Artifact | Master | With RocksDB | Δ |
|---|---|---|---|
| `gcs_server` (stripped) | 16.5 MB | 23.1 MB | +6.5 MB (+39%) |
| `gcs_server` (unstripped) | 23.8 MB | 32.0 MB | +8.2 MB (+34%) |
| ray wheel (py3.10 manylinux x86_64) | 70.2 MiB | 73.4 MiB | +3.2 MiB (+4.6%) |
| Docker image (`ray-py3.10-cpu`) | 2.15 GB | 2.17 GB | +20 MB (+0.9%) |

RocksDB is built via `rules_foreign_cc` with all optional features OFF (compression backends, RocksDB tools, Java bindings). Worth noting: the measured wheel delta (+3.2 MiB) is meaningfully smaller than the ~30 MB figure that came up in @siyuanfoundation's earlier thread, so the binary-size concern turned out weaker than initially argued. Also noted that an `extras_require` split (`ray[rocksdb]`) was considered and rejected — not worth a separate release artifact for a < 5% wheel-size increase.

reps/2026-02-23-gcs-embedded-storage.md
```diff
@@ -0,0 +1,529 @@
+# REP: Embedded Storage Backend for GCS Fault Tolerance
+
+## Summary
```
andrewsykim, line 3: Per offline discussion, let's outline the acceptance criterias somewhere in the REP
* Passing Ray release tests
* Passing Ray Core Redis test equivalent
* Passing Ray scale tests (e.g. test_many_actors.py)
* Passing Ray Serve chaos testing

edoakes: To clarify, the 2nd one should be a postmerge CI job alongside the existing redis one: https://github.com/ray-project/ray/blob/master/.buildkite/core.rayci.yml#L114

This can use rocksdb against a local file. We can label it as `skip-premerge` but should also have it run when GCS code is changed.

reps/2026-02-23-gcs-embedded-storage.md
```diff
@@ -0,0 +1,529 @@
...
+
+**Production context:** At LinkedIn, we run distributed PyTorch training jobs on Ray via Flyte on Kubernetes. These are long-running jobs (hours to days) where head node restarts — due to node maintenance, spot preemption, or transient failures — are inevitable. Today, head node failure is catastrophic: all training progress since the last application-level checkpoint is lost, and the entire job must be restarted. GCS fault tolerance with embedded storage, combined with training checkpointing, would allow these jobs to survive head node restarts and resume from where they left off — minimizing repeated computation and eliminating manual intervention.
+
+### Should this change be within `ray` or outside?
+
+This change affects both `ray` (new `StoreClient` implementation in C++) and `kuberay` (automatic PVC lifecycle management so users don't need to create or clean up PVCs manually).
+
+## Stewardship
+
+- **Required Reviewers**: @jjyao (GCS maintainer), @kevin85421 (KubeRay / GCS FT integration)
```
edoakes:
```suggestion
- **Required Reviewers**: @rueian, @andrewsykim, @MengjinYan 
```

reps/2026-02-23-gcs-embedded-storage.md
```diff
@@ -0,0 +1,529 @@
...
+- **Shepherd of the Proposal**: TBD — requesting @jjyao or another GCS-experienced committer
```
edoakes:
```suggestion
- **Shepherd of the Proposal**: @edoakes 
```

reps/2026-02-23-gcs-embedded-storage.md
```diff
@@ -0,0 +1,529 @@
...
+
+The recovery model is identical to today's Redis-based FT: GCS goes down, restarts, reads state, workers reconnect. The difference is purely in where state is persisted — local disk instead of a remote Redis instance.
+
+#### Impact on GCS OOM
+
+GCS is OOM-killed today primarily due to its in-memory working set (actor / task / node tables, task lineage), not the storage backend. This proposal does not change how much state GCS keeps resident and adds a small constant memory overhead (~20–50 MB for the RocksDB block cache and write buffers). In strict terms this makes head-pod memory pressure modestly worse, not better.
+
+Durable local storage does, however, *unlock* a future optimization: paging cold state out of RAM and faulting it back from RocksDB on demand (analogous to a database buffer pool). That path to making GCS OOM-resistant is out of scope for this REP and is listed under "State compaction and TTL" in follow-on work.
```
edoakes: This would be very cool! Or for example we could use rocksdb-on-disk for storing non-critical information like observability data

reps/2026-02-23-gcs-embedded-storage.md
```diff
@@ -0,0 +1,529 @@
...
+1. Best API fit — every `StoreClient` method has a natural, efficient RocksDB counterpart
+2. Bazel-native integration via BCR
+3. Largest production ecosystem and active maintenance
+4. Apache 2.0 license compatibility
+
+## Follow-on Work
+
+1. **[needed] Driver resilience during head pod restart** — The RayJob submitter pod survives head pod restarts physically, but the driver process may crash due to gRPC timeouts on pending RPCs. The RayJob controller and driver reconnection logic need hardening so the driver can survive GCS downtime gracefully and resume orchestration after recovery.
+
+2. **[potential] Active/Standby GCS** — A hot standby GCS process that takes over on failure, using RocksDB on a shared ReadWriteMany volume or replicated via WAL shipping. This would provide true HA (zero-downtime failover). The embedded storage backend proposed here is a prerequisite.
```
edoakes: There is another REP that was merged recently and the work is in flight to support this: https://github.com/ray-project/enhancements/pull/65

Please add a section to the REP about how this would be supported with rocksdb (`ReadWriteMany` volumes and any split brain issues).

edoakes (commented):
Looks good to me pending a few comments (including Andrew's). Let's give @rueian and @MengjinYan a chance to review before merging the REP (they are both traveling currently).

reps/2026-02-23-gcs-embedded-storage.md
jhasm: Added all four to the existing Acceptance Criteria subsection under "Test Plan and Acceptance Criteria".

Criterion #2 now explicitly calls out a postmerge CI job mirroring the Redis GCS FT job in .buildkite/core.rayci.yml, running against a local-file RocksDB, labeled skip-premerge, with path-based triggers on src/ray/gcs/** so it runs whenever GCS code changes.

reps/2026-02-23-gcs-embedded-storage.md
jhasm: Yeah that is good use case. Thanks for pointing out!

reps/2026-02-23-gcs-embedded-storage.md
jhasm: Thanks for pointing out. I updated the section "Interaction with Active-Passive Head" with the details. 

Fencing and lease mechanisms are same for Redis of Rocksdb and they both support atomic transactions. Rocksdb has additional LOCK that protects against multiple writers (corruption) and we'll have to rely on cooperation (clean demotion and exit) or kuberay kill (degraded case for frozen leader) for releasing the lock.

PR #65, opened 2026-04-23
REP: Ray Active-Passive Head Architecture

YoyinZyc:
This REP proposes the transition from a single-point-of-failure GCS model to a highly available Active-Passive setup. The design aims to reduce recovery latency from minutes to seconds.

Scope:
- Ray core: GCS native lease-based leader election client
- KubeRay: Ability to provision multiple head nodes
- Fault-tolerance: Align with existing GCS FT scope, targeting KubeRay based cluster.

Requested reviewers: @andrewsykim @edoakes @MengjinYan

### DIFF
--- a/reps/2026-04-23-active-passive-head.md
+++ b/reps/2026-04-23-active-passive-head.md
@@ -43,7 +43,7 @@ The initial design and implementation targets **KubeRay** exclusively for the fo
 2. **Election**: Both pods attempt to acquire the Kubernetes Lease (e.g., `ray-gcs-leader-lock`).
 3. **Active Node**: The winner acquires the lease, initializes its GCS managers, loads data from Redis, and starts the full GCS gRPC server.
 4. **Passive Node**: The loser enters **Observer Mode**:
-   - Starts the GCS gRPC server (`rpc_server_`) in **read-only mode** (allowing health check endpoints like `gcs_healthz` to succeed).
+   - Starts the GCS gRPC server (`rpc_server_`) in **standby mode** (allowing only the health check endpoint to succeed).
    - Defers loading Redis data.
    - Enters a continuous background loop polling the lease status.
 
@@ -71,7 +71,7 @@ The initial design and implementation targets **KubeRay** exclusively for the fo
 
 ### GCS Native Leader Election
 
-We propose integrating the Lease election protocol directly into the GCS C++ runtime. 
+We propose integrating the lease election protocol directly into the GCS C++ runtime. 
 
 #### 1. Protocol
 
@@ -104,13 +104,15 @@ The **Promotion Protocol** is designed to be non-blocking and safe:
 - `RAY_LEADER_ELECT_RESOURCE_NAMESPACE`: The operational namespace mapping the lease scope.
 
 #### 3. Request and Traffic Routing
-- **Kubernetes Service**: KubeRay creates a Service routing traffic to head pods.
+- **[Existing]Kubernetes Service for Worker Discovery**: 
+  - **Stable DNS Endpoint**: KubeRay provisions a static Head Service (e.g., `<cluster-name>-head-svc`) acting as the central, unchangeable discovery address for all worker pods. Workers configure their startup parameters to target this Service DNS name (`--address=<cluster-name>-head-svc:6379`) instead of ephemeral pod IPs.
 - **Readiness Probe Control**: 
   - The default KubeRay `readinessProbe` for the head node is usually configured to use an `exec` probe checking the local Dashboard API endpoints:
     `wget -T 2 -q -O- http://localhost:52365/api/local_raylet_healthz | grep success && wget -T 10 -q -O- http://localhost:8265/api/gcs_healthz | grep success`. We should extend the `gcs_healthz` endpoint to check the leadership status. It should return 200 OK if GCS is the active leader AND is healthy.
   - In **Passive Mode**, the readiness probe explicitly fails, removing the pod from Service endpoints.
   - Upon **Promotion**, the readiness probe succeeds, routing traffic to the new leader.
   - This mechanism prevents clients and workers from attempting to communicate with a standby GCS that has deferred loading its state.
+- **Seamless Failover Discovery**: During a transition, the unready standby pod is physically excluded from the Service endpoints. Once promoted, the new leader passes readiness checks, instantly updating the Service endpoints. Consequently, reconnecting workers automatically discover the new active GCS instance via the identical DNS address without requiring configuration updates or container restarts.
 - **Head Node Internal Traffic**: Processes on the head pod (e.g., Autoscaler, Dashboard) must be configured to connect **directly to the local GCS instance** (e.g., via `127.0.0.1`) rather than the Kubernetes Service address to avoid cross-pod state confusion.
 
 #### 4. Split-Brain Safeguards
@@ -161,7 +163,7 @@ Besides those generic safeguards, we can also apply the following Ray-specific o
 
 1. **Cross-head communication**: Establish the connection between heads. The passive cannot be promoted until it notices that the previous head is stepped down.
 
-2. **Worker Awareness**: The worker nodes track the GCS leader’s state(e.g. Pod name, generation ID…). If the local worker receives requests issued by a stale ID or unrecognized head, it drops the packet and ignores the request. 
+2. **Worker Awareness**: The worker nodes track the GCS leader’s state(e.g. Pod name, generation ID…). If the local worker receives requests issued by a stale ID or unrecognized head, it drops the RPC and ignores the request. 
 ##### Metrics and Monitoring
 - `gcs_is_leader`: Gauge metrics. Emits a 1 if the current instance is the leader, and a 0 if it is a passive head.
 - `gcs_leader_transitions_total`: Counter metrics. Incremented every time a leadership change is detected. For a single GCS process, the metrics will only ever transition from 0 to 1. Next time a failover occurs, a brand-new GCS pod will spin up, start at 0, and then transition to 1 when it becomes the active leader. 
@@ -290,9 +292,9 @@ When a `RayCluster` is deployed with `headGroupSpec.replicas: 2`, the KubeRay co
 
   **2. Sidecar components**
     - **Autoscaler**: In standard KubeRay setups where `enableInTreeAutoscaling: true` is configured, the Autoscaler is typically injected as a dedicated sidecar container running alongside the ray-head container inside the same pod. It is a critical, high-risk background component that must be suppressed on passive head.
-      - **Startup**: Upon boot, the local autoscaler targets the local GCS and periodically sends a GetClusterResourceState gRPC request to check if the cluster resource state in GCS is ready. The cluster resource state should only be ready when the GCS is in active mode.
+      - **Startup**: Upon boot, the local autoscaler targets the local GCS and periodically sends a `GetClusterResourceState` gRPC request to check if the cluster resource state in GCS is ready. The cluster resource state should only be ready when the GCS is in active mode.
       - **Active State**: Autoscaler gets the cluster resource state from GCS and continues with the normal autoscaling logic including updating KubeRay CRDs, reporting the autoscaling state back to GCS.
-      - **Passive State**: Autoscaler gets the cluster resource state from GCS, which indicates the cluster resource state is not ready. The autoscaler skips the rest of the logic.
+      - **Passive State**: Autoscaler continuously gets the cluster resource state from GCS, which indicates the cluster resource state is not ready. The autoscaler skips the rest of the logic.
 
 ## Baseline latency
 By default, we adhere to the standard `client-go` leader election settings:
--- a/reps/2026-04-23-active-passive-head.md
+++ b/reps/2026-04-23-active-passive-head.md
@@ -163,13 +163,14 @@ Besides those generic safeguards, we can also apply the following Ray-specific o
 
 2. **Worker Awareness**: The worker nodes track the GCS leader’s state(e.g. Pod name, generation ID…). If the local worker receives requests issued by a stale ID or unrecognized head, it drops the packet and ignores the request. 
 ##### Metrics and Monitoring
-- `gcs_is_leader`: Emits a 1 if the current instance is the leader, and a 0 if it is a passive head.
-- `gcs_leader_transitions_total`: Incremented every time a leadership change is detected
-- `gcs_recovery_latency_ms`: The time it takes for a passive head to recover the state and become the leader.
+- `gcs_is_leader`: Gauge metrics. Emits a 1 if the current instance is the leader, and a 0 if it is a passive head.
+- `gcs_leader_transitions_total`: Counter metrics. Incremented every time a leadership change is detected. For a single GCS process, the metrics will only ever transition from 0 to 1. Next time a failover occurs, a brand-new GCS pod will spin up, start at 0, and then transition to 1 when it becomes the active leader. 
+- `gcs_recovery_latency_ms`: Histogram metrics. The time it takes for a passive head to recover the state from the moment that it takes over the leadership.
 
-Even though the passive head is in "read-only" mode and failing readiness probes (so it doesn't get traffic), its metrics endpoint should still be active. The dashboard should plot these metrics for both pods, allowing users to see the health of both the active and passive heads.  
+Even though the passive head is in "read-only" mode and failing readiness probes (so it doesn't get traffic), its metrics endpoint should still be active. An external dashboard should aggregate and plot these metrics for both pods, allowing users to see the health of both the active and passive heads.  
 
-##### Detection and Alerting
+##### Detection and Alerting 
+The cluster-wide metrics aggregation and alert should be built outside of gcs/ray process. Here are some examples:
   - Alert if `gcs_is_leader` is 1 for both heads(split brain).
   - Alert if `gcs_recovery_latency_ms` is continously too high.
   - Alert if `gcs_leader_transitions_total` is too frequent(leader flapping).
@@ -186,10 +187,15 @@ Even though the passive head is in "read-only" mode and failing readiness probes
   To ensure a highly available Active-Passive architecture, all background processes running on the Ray Head Node must operate safely in a multi-head environment.
   - **GCS Server**
     Safe on Standby. Global tracking managers are instantiated empty; state hydration is deferred until promotion. The rpc server can only answer read requests in passive mode. All mutation requests should not be accepted and the rejection should be handled gracefully on the client side. (With the proper traffic routing, the requests will be sent to the active head only. But it is another layer of protection to avoid any potential split brain issue.)
-  - **Other components**
-    Requires Process Gating. The component can only run in passive mode.
-    - Raylet: 
-    (Raylet, Autoscaler, Dashboard, Job API, Serve Controller, Client Server, API Agent)
+  - **Raylet**: it is required to be alive to answer the readiness and liveness probes.
+      - **Startup**: Upon boot, the local standby Raylet targets the local GCS and sends a RegisterNode gRPC request. On passive head, the PUT request should not reach to Redis. 
+      - **Passive State**: The Raylet continuously sends heartbeats to the local GCS to keep its node liveness. 
+      - **Promotion**: During promotion, the Raylet should register itself and  physically writes it to Redis.
+  - **Autoscaler**: it is a critical, high-risk background component that must be suppressed on passive head.
+  - **Dashboard**: the local Dashboard process is started by KubeRay pointing to the local GCS. Because it queries the passive local GCS, the dashboard UI initially displays a single-node cluster (only itself) with no active jobs or actors. Since the standby pod is Not Ready, its dashboard API is completely unreachable by external clients with the current k8s service setup. 
+  - **Job API**: with right k8s service configuration, the `ray job submit` request should not be able to reach the passive head.
+  - **Serve Controller**: The Ray Serve Controller is a standard Ray actor. Since GCS in standby mode has no active workers and no running jobs, no Serve Controller is ever spawned on the standby head.
+  - **Dashboard Agent**: Runs locally and reports metrics to the co-located passive GCS. Upon promotion, it continues reporting to the now-promoted local GCS. Ready to initialize runtime environments if new jobs are submitted post-failover.
 
 ## Baseline latency
 By default, we adhere to the standard `client-go` leader election settings:
--- a/reps/2026-04-23-active-passive-head.md
+++ b/reps/2026-04-23-active-passive-head.md
@@ -269,7 +269,10 @@ When a `RayCluster` is deployed with `gcsFaultToleranceOptions.enableActivePassi
 ## Compatibility, Deprecation, and Migration Plan
 
 - **Head Node Components Compatibility**: 
-  To ensure a highly available Active-Passive architecture, all background processes running on the Ray Head Node must operate safely in a multi-head environment.
+  To ensure a highly available Active-Passive architecture, all background processes running on the Ray Head Node must operate safely in a multi-head environment. Specifically, all components must adhere to the following design principles:
+  1. **Suppress Unnecessary Work in Passive Mode**: Standby components must disable non-essential processes (e.g., registering nodes with the GCS, accepting job requests, publishing/aggregating cluster events).
+  2. **Avoid Interfering with the Active Leader**: Standby components must not perform tasks reserved for the active head node. For instance, only the active leader's autoscaler should update KubeRay CRDs, and only the active GCS server should persist states to Redis.
+  3. **Recover Functionality Automatically Post-Promotion**: Standby components must continuously monitor the leadership status (e.g., via polling periodically) and automatically resume full operations immediately upon promotion. While some components like Raylet and Autoscaler already have the channel to fetch status from local GCS periodically, we need to ensure that the leadership status is propagated to all components. The GCS needs to expose an rpc for other components to fetch its leadership status.
 
   **1. Single-container co-located components**:
   These components are running in the same container as different subprocesses. The gcs termination caused by lease lost will terminate other processes.
@@ -281,9 +284,9 @@ When a `RayCluster` is deployed with `gcsFaultToleranceOptions.enableActivePassi
       - **Passive State**: The Raylet continuously sends heartbeats to the local GCS to keep its node liveness. 
       - **Promotion**: During promotion, the Raylet should register itself and  physically writes it to Redis.
   - **Dashboard**: the local Dashboard process is started by KubeRay pointing to the local GCS. Because it queries the passive local GCS, the dashboard UI initially displays a single-node cluster (only itself) with no active jobs or actors. Since the standby pod is Not Ready, its dashboard API is completely unreachable by external clients with the current k8s service setup. 
-  - **Job API**: with right k8s service configuration, the `ray job submit` request should not be able to reach the passive head.
+  - **Job API**: with right k8s service configuration, the `ray job submit` request should not be able to reach the passive head. If it does, the server should reject the request with a well-defined error message and code under passive mode.
   - **Serve Controller**: The Ray Serve Controller is a standard Ray actor. Since GCS in standby mode has no active workers and no running jobs, no Serve Controller is ever spawned on the standby head.
-  - **Dashboard Agent**: Runs locally and reports metrics to the co-located passive GCS. Upon promotion, it continues reporting to the now-promoted local GCS. Ready to initialize runtime environments if new jobs are submitted post-failover.
+  - **Dashboard Agent**: Runs locally and reports events to the co-located passive GCS. Upon promotion, it continues reporting to the now-promoted local GCS. Ready to initialize runtime environments if new jobs are submitted post-failover. Block gcs event publishing in passive mode.
 
   **2. Sidecar components**
     - **Autoscaler**: In standard KubeRay setups where `enableInTreeAutoscaling: true` is configured, the Autoscaler is typically injected as a dedicated sidecar container running alongside the ray-head container inside the same pod. It is a critical, high-risk background component that must be suppressed on passive head.
--- a/reps/2026-04-23-active-passive-head.md
+++ b/reps/2026-04-23-active-passive-head.md
@@ -176,29 +176,123 @@ The cluster-wide metrics aggregation and alert should be built outside of gcs/ra
   - Alert if `gcs_leader_transitions_total` is too frequent(leader flapping).
 
 #### 5. KubeRay Changes
-1. **API change**: specifying the number of desired head replicas.
-2. ** Configuration changes**: 
-    - Disable Ray Head Service's `publishNotReadyAddresses`.
-    - Configure head internal communication to local GCS.
-    - Pod anti-affinity configuration to schedule active and passive heads on different nodes.
+
+To support the Active-Passive Head architecture, the KubeRay operator requires updates to its CRDs and controller reconciliation logic.
+
+##### 1. Customer-Facing API Changes
+
+Currently, the `HeadGroupSpec` in the `RayCluster` CRD assumes a single head node instance. We propose adding a `Replicas` field to the `HeadGroupSpec` API schema to allow users to explicitly configure the desired number of head pods. To streamline leader election tuning, we also extend `GcsFaultToleranceOptions` to natively support lease parameters.
+
+**Go API Schema Definitions (`apis/ray/v1/raycluster_types.go`):**
+
+```go
+type HeadGroupSpec struct {
+	// Replicas specifies the desired number of head node replicas.
+	// Defaults to 1. If set to 2, Active-Passive High Availability is enabled.
+	// +kubebuilder:default:=1
+	// +kubebuilder:validation:Minimum:=1
+	// +kubebuilder:validation:Maximum:=2
+	Replicas *int32 `json:"replicas,omitempty"`
+
+	// RayStartParams are the params of the start command: node-manager-port, object-store-memory, etc.
+	RayStartParams map[string]string `json:"rayStartParams"`
+...
+}
+
+type GcsFaultToleranceOptions struct {
+	// RedisConfiguration struct defining external storage endpoints
+	...
+
+	// LeaderElectionLeaseDurationSeconds is the duration that non-leader candidates wait before forcing leadership acquisition.
+	// +kubebuilder:default:=15
+	LeaderElectionLeaseDurationSeconds *int32 `json:"leaderElectionLeaseDurationSeconds,omitempty"`
+
+	// LeaderElectionRenewDeadlineSeconds is the acting leader's bounded deadline for executing consecutive renewal sequences.
+	// +kubebuilder:default:=10
+	LeaderElectionRenewDeadlineSeconds *int32 `json:"leaderElectionRenewDeadlineSeconds,omitempty"`
+
+	// LeaderElectionRetryPeriodSeconds is the duration clients wait between sequential resource acquisition attempts.
+	// +kubebuilder:default:=2
+	LeaderElectionRetryPeriodSeconds *int32 `json:"leaderElectionRetryPeriodSeconds,omitempty"`
+}
+```
+
+**Validation Webhook:**
+- A validating webhook will enforce that `Replicas` can only be set to `1` or `2`. Initially, we limit the replicas to 2.
+- If `Replicas` is set to `2`, the webhook validates that external Redis persistence is configured under `spec.gcsFaultToleranceOptions` (since native GCS recovery relies on Redis storage).
+- Enforces timing invariants if customized: `LeaderElectionLeaseDurationSeconds > LeaderElectionRenewDeadlineSeconds > LeaderElectionRetryPeriodSeconds`.
+
+**Example Customer Manifest (`RayCluster` CR):**
+
+```yaml
+apiVersion: ray.io/v1
+kind: RayCluster
+metadata:
+  name: raycluster-ha-sample
+  namespace: default
+spec:
+  rayVersion: '2.9.0'
+  gcsFaultToleranceOptions:
+    redisAddress: "redis:6379"
+    ...
+    leaderElectionLeaseDurationSeconds: 20
+    leaderElectionRenewDeadlineSeconds: 12
+    leaderElectionRetryPeriodSeconds: 3
+  headGroupSpec:
+    replicas: 2 # Enable Active-Passive HA
+    rayStartParams:
+      ...
+    template:
+      spec:
+        containers:
+        - name: ray-head
+          image: rayproject/ray:2.9.0
+```
+
+##### 2. Controller Configuration & Reconciliation Logic
+
+When a `RayCluster` is deployed with `headGroupSpec.replicas: 2`, the KubeRay controller executes the following advanced reconciliation logic:
+
+- **Head Pod Provisioning & Lifecycle**:
+  - Provisions two independent head pods labeled as candidates for leadership.
+  - Automatically maps the fields configured under `gcsFaultToleranceOptions` into the required environment variables inside the head containers to trigger native C++ GCS leader election:
+    - `RAY_LEADER_ELECT=true`
+    - `RAY_LEADER_ELECT_RESOURCE_NAME=<cluster-name>-gcs-leader-lock`
+    - `RAY_LEADER_ELECT_RESOURCE_NAMESPACE=<pod-namespace>`
+    - `RAY_LEADER_ELECT_LEASE_DURATION=<value>`
+    - `RAY_LEADER_ELECT_RENEW_DEADLINE=<value>`
+    - `RAY_LEADER_ELECT_RETRY_PERIOD=<value>`
+- **Service Routing & Readiness Control**:
+  - Modifies the Ray Head Service reconciliation to explicitly set `publishNotReadyAddresses: false` (overriding default behaviors where head services might publish unready addresses).
+  - Configures the Kubernetes Head Service to map its endpoints strictly based on pod readiness, external and worker traffic is dynamically routed exclusively to the active leader pod.
+- **Internal Component Routing**:
+  - Co-located system components running inside the head pod (such as the Autoscaler, Dashboard, and Dashboard Agent) must be configured to communicate with the local loopback GCS interface (`127.0.0.1:6379`) rather than the shared Kubernetes Head Service DNS name. This prevents cross-pod pollution where a background process on the passive head communicates with the active head's GCS.
+- **Automatic Pod Anti-Affinity Injection**:
+  - To prevent a single-node outage from terminating both head pods simultaneously, the controller inspects the head pod template. If no explicit affinity rules are defined, it automatically injects a `podAntiAffinity` policy targeting the `ray.io/node-type: head` label to schedule active and passive heads on different nodes.
 ## Compatibility, Deprecation, and Migration Plan
 
 - **Head Node Components Compatibility**: 
   To ensure a highly available Active-Passive architecture, all background processes running on the Ray Head Node must operate safely in a multi-head environment.
+
+  **1. Single-container co-located components**:
+  These components are running in the same container as different subprocesses. The gcs termination caused by lease lost will terminate other processes.
   - **GCS Server**
     Safe on Standby. Global tracking managers are instantiated empty; state hydration is deferred until promotion. The rpc server can only answer read requests in passive mode. All mutation requests should not be accepted and the rejection should be handled gracefully on the client side. (With the proper traffic routing, the requests will be sent to the active head only. But it is another layer of protection to avoid any potential split brain issue.)
+    - **Leadership Loss**: If the active GCS fails to renew its lease or voluntarily steps down, it immediately self-terminates (`RAY_LOG(FATAL)`) to eliminate split-brain risks, instantly closing all active gRPC client streams.
   - **Raylet**: it is required to be alive to answer the readiness and liveness probes.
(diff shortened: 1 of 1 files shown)


reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,187 @@
...
+
+There are two scenarios that could suffer from this issue:
+
+##### 1. Network Partition
+A network partition occurs while the Leadership Renew Thread is trying to contact the K8s API server. The HTTP client hangs indefinitely waiting for a response. Because the thread is stuck, it cannot check its internal clock, misses the `RenewDeadline`, and fails to trigger a process suicide before K8s elects a new leader.
+
+##### 2. Process Frozen
+The host machine experiences severe CPU starvation, or the OS forcefully pauses your entire GCS process. The K8s lease expires, and a new leader takes over. When the OS finally unfreezes the process, the GCS main thread instantly executes some ongoing resource update and database writes before the renew thread has the microsecond needed to call `fatal()` and kill itself.
+
+##### Safeguards & Mitigations
```
andrewsykim, line 130: @YoyinZyc can we make it clearer which of these are in current scope and which are future considerations?

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,187 @@
...
+
+1. **Time-bounded parameters configuration**: To ensure safe failovers and prevent 99% of the split brain problem, the timing variables(LeaseDuration, RenewDeadline, HTTP Timeout, RetryPeriod) should align with each other. The following shows the formula for best practice:
+	- `LeaseDuration > RenewDeadline > RetryPeriod`;
+	- `LeaseDuration > HTTP Timeout + RenewDeadline`;
+	- `RenewDeadline > RetryPeriod * factor(3 or 5)`;
+
+	Example: LeaseDuration - 15s; RenewDeadline - 10s; HTTP Timeout - 2s; RetryPeriod - 2s;
```
andrewsykim, line 140: Are these the defaults we're planning to use? If so what is the baseline latency for lease transition?

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,187 @@
...
+
+2. **Atomic variable for leadership status**: For the process frozen scenario, we want to have a faster way to send the leadership loss signal to the main thread. We could achieve this by sharing a lock-free `std::atomic<bool>` `is_leader_{true}` between your threads.
+	- The leadership status must be checked before every write operation.
+	- When the Renew Thread wakes up and realizes it missed the deadline, it must set the value to false This memory write takes only a few nanoseconds. After doing this, it can proceed with the slower fatal() system call.
+
+3. **Fencing Token**: The last barrier to prevent data inconsistency. It is also one of the best practices for implementing lease-based leader election in production. At the end of the day, we must make sure the data stored in redis is pure.
+	- **The Token**: Lease transition, it is updated when a new leader takes over the lease
+	- **Write to Redis**: the leader GCS attaches the token with the query
+	- **Fencing**: The Redis rejects the request if the request_epoch is smaller than its current_epoch
```
andrewsykim: @YoyinZyc how exactly does redis reject requests based on the token? Is it a feature in Redis or something else?

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,187 @@
...
+
+The first three mechanisms are production golden standards for leader election implementations. They are generic and must be included in the first phase of this REP. 
+
+The latter two are Ray-specific optimizations for the GCS HA scenario. Techenically, they could provide additional data consistency guarantees. However, it may involve other potential risk and additional implementation work.(e.g. communication between heads/workers could be unreliable) We can leave it to the future iterations. 
+
+#### 5. KubeRay Changes
+1. **API change**: specifying the number of desired head replicas.
+2. **Configuration changes**: 
+    - Readiness Probe for passive heads to fail.
+    - Disable PublishNotReadyAddresses.
```
andrewsykim: specify that this is for the Ray Head Service

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
...
+
+##### Detection and Alerting
+  - Alert if `gcs_is_leader` is 1 for both heads(split brain).
+  - Alert if `gcs_recovery_latency_ms` is continously too high.
+  - Alert if `gcs_leader_transitions_total` is too frequent(leader flapping).
+
+#### 5. KubeRay Changes
+1. **API change**: specifying the number of desired head replicas.
+2. **Configuration changes**: 
+    - Readiness Probe for passive heads to fail.
```
rueian: IIRC, probes are immutable. So, probably the change should be in GCS?

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
+## Summary
+
+### General Motivation
+
+Ray clusters rely on the Global Control Store (GCS) as the centralized metadata store and cluster manager. In High Availability (HA) deployments, the GCS state is backed by an external persistent store (e.g., Redis). When the primary head node fails, standard recovery mechanisms involve provisioning a new head pod from scratch. This introduces significant latency (image pulling, container startup, cluster re-registration) and risks disconnecting active worker nodes due to prolonged GCS downtime.
+
+To minimize downtime, the **Shadow Head** architecture introduces a hot standby (passive) head node that runs alongside the primary. This enhancement proposal details the implementation of a **Native Leader Election** mechanism embedded directly within the C++ GCS process. By using [Kubernetes Leases](https://kubernetes.io/docs/concepts/architecture/leases/), the shadow head can rapidly detect failures, acquire the lease, and promote itself to active leader in seconds.
```
edoakes: nit: let's stick to the terminology of "passive" head instead of calling it "shadow"

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
...
+
+
+Historically, Ray components have assumed a strict single-head node presence. This design explicitly breaks this assumption, opening the window for true Active-Active HA capabilities.
```
edoakes:
```suggestion
Historically, Ray components have assumed a strict single-head node presence. This design explicitly breaks this assumption, laying the ground work for true HA capabilities.
```

edoakes: I don't want to give the idea that we are moving towards active<>active (very unclear if/how we'd want to do it)

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
...
+
+
+### Should this change be within `ray` or outside?
+
+This lease-based leader election change should be within Ray core (specifically the C++ GCS server) to ensure fate-sharing, minimize external dependencies, and optimize state transition performance.
+
+We also need some kuberay implementation to support this change.
```
edoakes:
```suggestion
We also need some KubeRay implementation to support this change.
```

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
...
+
+To minimize downtime, the **Shadow Head** architecture introduces a hot standby (passive) head node that runs alongside the primary. This enhancement proposal details the implementation of a **Native Leader Election** mechanism embedded directly within the C++ GCS process. By using [Kubernetes Leases](https://kubernetes.io/docs/concepts/architecture/leases/), the shadow head can rapidly detect failures, acquire the lease, and promote itself to active leader in seconds.
+
+
+Historically, Ray components have assumed a strict single-head node presence. This design explicitly breaks this assumption, opening the window for true Active-Active HA capabilities.
+
+
+### Should this change be within `ray` or outside?
```
edoakes, line 14: Please also mention that we are currently targeting KubeRay only because that is where most of the production serving workloads run, but it can be extended in the future if needed.

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
...
+
+**Time-Bounded Leases**: A distributed lock pattern where a candidate acquires a time-limited lock (a "lease"). The leader must continuously renew the lease before it expires. If renewal fails, the lease drops, allowing other candidates to contend for ownership.
+
+
+**Why [Kubernetes Leases](https://kubernetes.io/docs/concepts/architecture/leases/)?**
+
+- **Infrastructure Reuse**: Reuses the existing, highly available Kubernetes control plane without introducing new external dependencies.
+- **Simplicity**: Avoids the complexity of embedding a heavy C++ consensus framework (like Raft) directly into Ray.
+- **Fate Sharing**: Leverages Kubernetes' built-in awareness of Pod availability and node health.
+- **Target Environment**: Currently, GCS fault tolerance and High Availability are only supported for Ray deployments running on Kubernetes (via KubeRay).
```
edoakes: move this part up into the motivation/"inside ray" section

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
...
+
+
+### Cluster Lifecycle & Scenarios
+
+#### Scenario A: Cluster Initialization
+1. **Startup**: KubeRay provisions two head pods simultaneously (Active and Passive).
+2. **Election**: Both pods attempt to acquire the Kubernetes Lease (e.g., `ray-gcs-leader-lock`).
+3. **Active Node**: The winner acquires the lease, initializes its GCS managers, loads data from Redis, and starts the full GCS gRPC server.
+4. **Passive Node**: The loser enters **Observer Mode**:
+   - Starts the GCS gRPC server (`rpc_server_`) in **read-only mode** (allowing health check endpoints like `gcs_healthz` to succeed).
```
edoakes:
```suggestion
   - Starts the GCS gRPC server (`rpc_server_`) in **standby mode** (allowing only the health check endpoint to succeed).
```

Don't want to give the impression that it is a "read replica"

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
...
+   - Defers loading Redis data.
+   - Enters a continuous background loop polling the lease status.
+
+#### Scenario B: Active Head Node Failure
+1. **Detection**: The active head pod crashes or becomes unresponsive.
+2. **Lease Expiration**: The active head pod fails to renew the lease, and the lease expires after the configured TTL.
+3. **Promotion**:
+   - The passive head successfully acquires the lease.
+   - Triggers `DoStartLoading()` to load the latest cluster state from Redis.
+   - Promotes itself to the primary leader and accepts write/read traffic.
```
edoakes, line 56: Note here that this is the same process that happens on GCS restart today

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
...
+
+#### Scenario E: Kubernetes API Server Disconnection
+1. **Transient Network Disconnection**: If the active leader cannot renew the lease due to network flaps, it relies on the remaining `LeaseDuration`. If the timeout is reached without renewal, the leader **must aggressively terminate (`RAY_LOG(FATAL)`)** to prevent split-brain.
+2. **Complete Outage**: If the central K8s control plane goes offline, neither pod can modify leases. The standby remains passive. The active leader may optionally continue running to maintain availability, acknowledging that failover is impossible.
+3. **Tradeoff**: It is the tradeoff between availability and safety. From the active head perspective, it has no idea about if the current disconnection is an APIServer outage or just its own network partition. Longer lease duration can avoid frequent leader transition.
+
+### GCS Native Leader Election
+
+We propose integrating the Lease election protocol directly into the GCS C++ runtime.
```
edoakes:
```suggestion
We propose integrating the lease election protocol directly into the GCS C++ runtime. 
```

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
...
+
+We propose integrating the Lease election protocol directly into the GCS C++ runtime. 
+
+#### 1. Protocol
+
+The architecture relies on an **Active-Passive** model with distinct responsibilities:
+- **Active Node (Leader)**: Holds the Kubernetes Lease, initializes all GCS managers, hydrates the cluster state from Redis, and fully serves read/write traffic. It is responsible for continuously renewing the lease.
+- **Passive Node (Standby)**: Operates in a restricted shadow mode. It runs the GCS gRPC server in a read-only state (to satisfy health checks) but defers loading cluster state and isolates itself from external traffic routing. It continuously polls the lease status.
+
+For Ray on Kubernetes, these two pods will be scheduled on different nodes by leveraging `podAntiAffinity` configuration to prevent a single node failure from taking down both the active and passive heads.
```
edoakes, line 82: will this be injected by KubeRay automatically or will users need to set it?

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Required Reviewers
+- @MengjinYan
+- @andrewsykim
+- @edoakes
+
+### Shepherd of the Proposal (should be a senior committer)
+- @edoakes
+
+## Design and Architecture
```
edoakes, line 37: Can you please add the proposed API/CR changes to KubeRay here as well?

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
...
+**New Environment Variables:**
+- `RAY_LEADER_ELECT`: Enables native leader election for the GCS process.
+- `RAY_LEADER_ELECT_LEASE_DURATION`: The duration that non-leader candidates will wait before forcing an acquisition of leadership (default: 15s).
+- `RAY_LEADER_ELECT_RENEW_DEADLINE`: The acting leader's bounded deadline for executing consecutive renewal sequences before voluntarily stepping down (default: 10s).
+- `RAY_LEADER_ELECT_RETRY_PERIOD`: The duration clients wait between sequential resource attempts (default: 2s).
+- `RAY_LEADER_ELECT_RESOURCE_NAME`: The name identifier of the target Kubernetes Lease object.
+- `RAY_LEADER_ELECT_RESOURCE_NAMESPACE`: The operational namespace mapping the lease scope.
+
+#### 3. Request and Traffic Routing
+- **Kubernetes Service**: KubeRay creates a Service routing traffic to head pods.
```
edoakes: What type of k8s service will this be?

Currently the raylets & workers take a raw IP for the GCS. This will mean that they take a hostname instead.

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
...
+- **Readiness Probe Control**: 
+  - In **Passive Mode**, the readiness probe explicitly fails, removing the pod from Service endpoints.
+  - Upon **Promotion**, the readiness probe succeeds, routing traffic to the new leader.
+  - This mechanism prevents clients and workers from attempting to communicate with a standby GCS that has deferred loading its state.
+- **Head Node Internal Traffic**: Processes on the head pod (e.g., Autoscaler, Dashboard) must be configured to connect **directly to the local GCS instance** (e.g., via `127.0.0.1`) rather than the Kubernetes Service address to avoid cross-pod state confusion.
```
edoakes, line 116: will the autoscaler & dashboard also be in a "passive standby" mode? and how do they know when to begin operating?

autoscaler split brain is as bad as GCS split brain

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
...
+	**Note**: The longer the LeaseDuration, the more resilient the cluster is to temporary network glitches, but the longer it takes for a failed node to be considered offline.
+
+2. **Fencing Token**: The last barrier to prevent data inconsistency. It is also one of the best practices for implementing lease-based leader election in production. At the end of the day, we must make sure the data stored in redis is pure.
+	- **The Token**: Lease transition, it is updated when a new leader takes over the lease
+	- **Write to Redis**: the leader GCS attaches the token with the query
+	- **Fencing**: The Redis rejects the request if the request_epoch is smaller than its current_epoch
+	
+3. **Cross-head communication**: Establish the connection between heads. The passive cannot be promoted until it notices that the previous head is stepped down.
+
+4. **Worker Awareness**: The worker nodes track the GCS leader’s state(e.g. Pod name, generation ID…). If the local worker receives requests issued by a stale ID or unrecognized head, it drops the packet and ignores the request.
```
edoakes:
```suggestion
4. **Worker Awareness**: The worker nodes track the GCS leader’s state(e.g. Pod name, generation ID…). If the local worker receives requests issued by a stale ID or unrecognized head, it drops the RPC and ignores the request. 
```

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
...
+
+There are two scenarios that could suffer from this issue:
+
+##### 1. Network Partition
+A network partition occurs while the Leadership Renew Thread is trying to contact the K8s API server. The HTTP client hangs indefinitely waiting for a response. Because the thread is stuck, it cannot check its internal clock, misses the `RenewDeadline`, and fails to trigger a process suicide before K8s elects a new leader.
+
+##### 2. Process Frozen
+The host machine experiences severe CPU starvation, or the OS forcefully pauses your entire GCS process. The K8s lease expires, and a new leader takes over. When the OS finally unfreezes the process, the GCS main thread instantly executes some ongoing resource update and database writes before the renew thread has the microsecond needed to call `fatal()` and kill itself.
+
+##### Safeguards & Mitigations
```
edoakes, line 130: also note that cgroup resource isolation helps address this

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
...
+	- **Write to Redis**: the leader GCS attaches the token with the query
+	- **Fencing**: The Redis rejects the request if the request_epoch is smaller than its current_epoch
+	
+3. **Cross-head communication**: Establish the connection between heads. The passive cannot be promoted until it notices that the previous head is stepped down.
+
+4. **Worker Awareness**: The worker nodes track the GCS leader’s state(e.g. Pod name, generation ID…). If the local worker receives requests issued by a stale ID or unrecognized head, it drops the packet and ignores the request. 
+
+The first three mechanisms are production golden standards for leader election implementations. They are generic and must be included in the first phase of this REP. 
+
+The latter two are Ray-specific optimizations for the GCS HA scenario. Techenically, they could provide additional data consistency guarantees. However, it may involve other potential risk and additional implementation work.(e.g. communication between heads/workers could be unreliable) We can leave it to the future iterations.
```
edoakes:
```suggestion
The latter two are Ray-specific optimizations for the GCS HA scenario. Techenically, they could provide additional data consistency guarantees. However, it may involve other potential risk and additional implementation work.(e.g. communication between heads/workers could be unreliable). We can leave them to future iterations. 

```

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
...
+3. **Cross-head communication**: Establish the connection between heads. The passive cannot be promoted until it notices that the previous head is stepped down.
+
+4. **Worker Awareness**: The worker nodes track the GCS leader’s state(e.g. Pod name, generation ID…). If the local worker receives requests issued by a stale ID or unrecognized head, it drops the packet and ignores the request. 
+
+The first three mechanisms are production golden standards for leader election implementations. They are generic and must be included in the first phase of this REP. 
+
+The latter two are Ray-specific optimizations for the GCS HA scenario. Techenically, they could provide additional data consistency guarantees. However, it may involve other potential risk and additional implementation work.(e.g. communication between heads/workers could be unreliable) We can leave it to the future iterations. 
+##### Metrics and Monitoring
+- `gcs_is_leader`: Emits a 1 if the current instance is the leader, and a 0 if it is a passive head.
+- `gcs_leader_transitions_total`: Incremented every time a leadership change is detected
```
edoakes: where would this be exported?

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
...
+- `gcs_recovery_latency_ms`: The time it takes for a passive head to recover the state and become the leader.
```
edoakes: same for this, where/how is it recorded?

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
...
+	- **Fencing**: The Redis rejects the request if the request_epoch is smaller than its current_epoch
+	
+3. **Cross-head communication**: Establish the connection between heads. The passive cannot be promoted until it notices that the previous head is stepped down.
+
+4. **Worker Awareness**: The worker nodes track the GCS leader’s state(e.g. Pod name, generation ID…). If the local worker receives requests issued by a stale ID or unrecognized head, it drops the packet and ignores the request. 
+
+The first three mechanisms are production golden standards for leader election implementations. They are generic and must be included in the first phase of this REP. 
+
+The latter two are Ray-specific optimizations for the GCS HA scenario. Techenically, they could provide additional data consistency guarantees. However, it may involve other potential risk and additional implementation work.(e.g. communication between heads/workers could be unreliable) We can leave it to the future iterations. 
+##### Metrics and Monitoring
```
edoakes, line 167: are there any other metrics around API server latency / health that we should observe?

reps/2026-04-23-active-passive-head.md
rueian: I think this can be implemented with redis lua script, but I am wondering if doing fencing in redis is enough. Shouldn't we also have this fencing in all GCS -> raylet communication?

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
...
+    - Configure head internal communication to local GCS.
+    - Pod anti-affinity configuration to schedule active and passive heads on different nodes.
+## Compatibility, Deprecation, and Migration Plan
+
+- **Head Node Components Compatibility**: 
+  To ensure a highly available Active-Passive architecture, all background processes running on the Ray Head Node must operate safely in a multi-head environment.
+  - **GCS Server (`gcs_server`)**
+    Safe on Standby. Global tracking managers are instantiated empty; state hydration is deferred until promotion.
+  - **Raylet**
+    Safe to run only if launched with `--num-cpus=0` and `--num-gpus=0`, ensuring the active primary never schedules user tasks onto the standby head node.
```
edoakes: Setting these is not strong enough -- actors/tasks with no resource requests can still be scheduled there.

Can we gate the Raylet from joining the cluster until after the GCS is elected the leader? This is similar to the problem w/ the autoscaler that I pointed out above

reps/2026-04-23-active-passive-head.md
andrewsykim: I think KubeRay can add this

reps/2026-04-23-active-passive-head.md
andrewsykim: We already create a headless Service from KubeRay, so this is referring to an existing Service

edoakes: Ah that's right. I don't think the raylets & workers resolve the address through that service currently though, they still take the raw IP (correct me if I'm wrong)

andrewsykim: no they use DNS hostname managed by K8s that resolves to the Pod IPs directly

andrewsykim: @YoyinZyc we should clarify these details in the doc as it's important for head service discovery during fail over

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
...
+- `RAY_LEADER_ELECT_LEASE_DURATION`: The duration that non-leader candidates will wait before forcing an acquisition of leadership (default: 15s).
+- `RAY_LEADER_ELECT_RENEW_DEADLINE`: The acting leader's bounded deadline for executing consecutive renewal sequences before voluntarily stepping down (default: 10s).
+- `RAY_LEADER_ELECT_RETRY_PERIOD`: The duration clients wait between sequential resource attempts (default: 2s).
+- `RAY_LEADER_ELECT_RESOURCE_NAME`: The name identifier of the target Kubernetes Lease object.
+- `RAY_LEADER_ELECT_RESOURCE_NAMESPACE`: The operational namespace mapping the lease scope.
+
+#### 3. Request and Traffic Routing
+- **Kubernetes Service**: KubeRay creates a Service routing traffic to head pods.
+- **Readiness Probe Control**: 
+  - In **Passive Mode**, the readiness probe explicitly fails, removing the pod from Service endpoints.
```
abrarsheikh, line 112: would the readiness probe fails comes from the gRPC server in GCS?

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,196 @@
+## Summary
+
+### General Motivation
+
+Ray clusters rely on the Global Control Store (GCS) as the centralized metadata store and cluster manager. In High Availability (HA) deployments, the GCS state is backed by an external persistent store (e.g., Redis). When the primary head node fails, standard recovery mechanisms involve provisioning a new head pod from scratch. This introduces significant latency (image pulling, container startup, cluster re-registration) and risks disconnecting active worker nodes due to prolonged GCS downtime.
```
abrarsheikh, line 5: could you explain a little, how/when this can happen with prolonged GCS downtime
```
risks disconnecting active worker nodes due to prolonged GCS downtime.
```

reps/2026-04-23-active-passive-head.md
YoyinZyc: Yeah, I was originally reaching to k8s APIServer directly to fetch the lease status for readiness check but it could be part of the gcs_health endpoint. In this way, no need to update the readiness probe. Updated.

reps/2026-04-23-active-passive-head.md
YoyinZyc: Done

reps/2026-04-23-active-passive-head.md
YoyinZyc: Yes, it is with lua script, added more details and the example lua script. GCS -> raylet communication involves more design and consideration, I added it as the a phase 2 item which should fall into later iteration.

reps/2026-04-23-active-passive-head.md
YoyinZyc: Done

reps/2026-04-23-active-passive-head.md
YoyinZyc: Done

reps/2026-04-23-active-passive-head.md
YoyinZyc: done

reps/2026-04-23-active-passive-head.md
YoyinZyc: Done

reps/2026-04-23-active-passive-head.md
YoyinZyc: Done

reps/2026-04-23-active-passive-head.md
YoyinZyc: Done

reps/2026-04-23-active-passive-head.md
YoyinZyc: Done.

reps/2026-04-23-active-passive-head.md
YoyinZyc: yes, we need to extend the gcs_health endpoint to also check local leadership status.

reps/2026-04-23-active-passive-head.md
YoyinZyc: I think the worker will kill itself if the gcs is disconnected over the specified timeout.

reps/2026-04-23-active-passive-head.md
YoyinZyc: It is the default value of client-go leader election https://pkg.go.dev/k8s.io/client-go/tools/leaderelection. I think we can extend it as default here as well.

I ran failover for 30 iterations, and the average latency is ~20s. It is 5s more than the configured ttl 15s since the:
1. GCS head only checks the Lease status and attempts to acquire leadership once every 2 seconds (retry_period).
2. Potential APIServer delay
3. GCS data loading after leadership transition can take 1~2s

Added a section about the baseline latency.

reps/2026-04-23-active-passive-head.md
YoyinZyc: From the gcs process.

reps/2026-04-23-active-passive-head.md
YoyinZyc: It is recorded from the gcs server. gcs_is_leader is a gauge metrics tracking local leadership status; `gcs_leader_transitions_total` increaments when the passive becomes the leader;
`gcs_recovery_latency_ms` tracks the passive head recovery latency from the moment it acquires the leadership to the moment the head is functional.

reps/2026-04-23-active-passive-head.md
YoyinZyc: APIServer has a bunch of metrics but those are pretty generic like `apiserver_request_total`, `apiserver_request_duration_seconds` and `apiserver_current_inflight_requests`

reps/2026-04-23-active-passive-head.md
YoyinZyc: yes, autoscaler must be suppressed. Raylet may be fine since it anyway only talks to the local gcs but we should restrict its registration which incurs the redis write. Added more detail about components behavior.

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,227 @@
...
+      - **Startup**: Upon boot, the local standby Raylet targets the local GCS and sends a RegisterNode gRPC request. On passive head, the PUT request should not reach to Redis. 
+      - **Passive State**: The Raylet continuously sends heartbeats to the local GCS to keep its node liveness. 
+      - **Promotion**: During promotion, the Raylet should register itself and  physically writes it to Redis.
+  - **Autoscaler**: it is a critical, high-risk background component that must be suppressed on passive head.
+      - **Startup**: Upon boot, the local autoscaler targets the local GCS and periodically sends a GetClusterResourceState gRPC request to check if the cluster resource state in GCS is ready. The cluster resource state should only be ready when the GCS is in active mode.
+      - **Active State**: Autoscaler gets the cluster resource state from GCS and continues with the normal autoscaling logic including updating KubeRay CRDs, reporting the autoscaling state back to GCS.
+      - **Passive State**: Autoscaler gets the cluster resource state from GCS, which indicates the cluster resource state is not ready. The autoscaler skips the rest of the logic.
```
edoakes: Will each of these also be monitoring that its GCS is the leader and fast-failing when it loses the lease?

reps/2026-04-23-active-passive-head.md
edoakes: It is probably worth also exporting similar metrics from our lease client

reps/2026-04-23-active-passive-head.md
edoakes: Ping on this -- let's include the full scope of user-facing KubeRay changes

reps/2026-04-23-active-passive-head.md
edoakes: Sounds good, that will be better. @YoyinZyc let's add it to a kuberay section please

edoakes (commented):
Largely looks good. Primary open feedback:

- Please document the full scope of KubeRay changes (especially any user-facing changes).
- We will need the same safety guarantees for the autoscaler process. It will either need to fast fail when the GCS fails or also directly check w/ a lease client.

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,296 @@
...
+**Go API Schema Definition (`apis/ray/v1/raycluster_types.go`):**
+
+```go
+type HeadGroupSpec struct {
+	// Replicas specifies the desired number of head node replicas.
+	// Defaults to 1. If set to 2, Active-Passive High Availability is enabled.
+	// +kubebuilder:default:=1
+	// +kubebuilder:validation:Minimum:=1
+	// +kubebuilder:validation:Maximum:=2
+	Replicas *int32 `json:"replicas,omitempty"`
```
andrewsykim: Are we introducing `replicas` now because we anticipate it can be more than 2 head nodes in the future?

andrewsykim: (As opposed to something like `mode: SingleReplica|ActivePassive`)

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,296 @@
...
+
+```yaml
+apiVersion: ray.io/v1
+kind: RayCluster
+metadata:
+  name: raycluster-ha-sample
+  namespace: default
+spec:
+  rayVersion: '2.9.0'
+  gcsFaultToleranceOptions:
```
andrewsykim, line 233: should we allow configuring the leader election leasing parameters through gcsFaultToleranceOptions?

andrewsykim: specifically thinking about lease duration, retries, etc

reps/2026-04-23-active-passive-head.md
andrewsykim: I think `replicas` is confusing because it implies a user can set it to any arbitrary value.

A field in `gcsFaultToleranceOption` is probably better?
```
spec:
  gcsFaultToleranceOptions:
    mode: ActivePassive
    ..
    ...
``` 

This is slightly more intuitive because active-passive mode also requires enabling Redis persistence anyways right?

andrewsykim: WDYT @rueian ?

reps/2026-04-23-active-passive-head.md
YoyinZyc: yes, added.

reps/2026-04-23-active-passive-head.md
YoyinZyc: I was also thinking about just using a boolean or mode to control the enablement. But using `replicas` is more kubernetes native for controller to reconcile and can be extended if ray eventually wants to support more replicas?

reps/2026-04-23-active-passive-head.md
YoyinZyc: Sorry, somehow missed this comment in the issue conversation. Added the kuberay design in the latest version.

reps/2026-04-23-active-passive-head.md
YoyinZyc: Added.

reps/2026-04-23-active-passive-head.md
YoyinZyc: done

reps/2026-04-23-active-passive-head.md
YoyinZyc: Added more detail about the existing service and worker discovery.

reps/2026-04-23-active-passive-head.md
YoyinZyc: yes, the autoscaler gets the state from gcs continuously to make scaling decision. So it can fetch the leadership state from gcs as well. It should do any scaling work until the gcs is promoted.

reps/2026-04-23-active-passive-head.md
YoyinZyc: Done

reps/2026-04-23-active-passive-head.md
YoyinZyc: added a sentence for it.

reps/2026-04-23-active-passive-head.md
YoyinZyc: What similar metrics are you referring to? These are k8s APIServer metrics, how could the lease client export them?

reps/2026-04-23-active-passive-head.md
YoyinZyc: Are you referring to scenario D which is the same process GCS has today?

reps/2026-04-23-active-passive-head.md
andrewsykim: > But using replicas is more kubernetes native for controller to reconcile and can be extended if ray eventually wants to support more replicas?

I think we can just introduce `replicas` field in the future when we can support > 2 replicas. Putting replicas outside of `gcsFaultToleranceOptions` implies you can scale to N replicas without enabling GCS persistence, which is not the case. Putting the mode in `gcsFaultToleranceOptions` makes it clearer that you can only enable active-passive replica mode when GCS persistence is also enabled

reps/2026-04-23-active-passive-head.md
YoyinZyc: @gangli113 helped adding more details about autoscaler, let us know if there is any other concerns. Thanks!

reps/2026-04-23-active-passive-head.md
edoakes: I mean our lease client should also export metrics like the latency for its requests to the API server and a failed request count

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,227 @@
...
+      - **Startup**: Upon boot, the local autoscaler targets the local GCS and periodically sends a GetClusterResourceState gRPC request to check if the cluster resource state in GCS is ready. The cluster resource state should only be ready when the GCS is in active mode.
+      - **Active State**: Autoscaler gets the cluster resource state from GCS and continues with the normal autoscaling logic including updating KubeRay CRDs, reporting the autoscaling state back to GCS.
+      - **Passive State**: Autoscaler gets the cluster resource state from GCS, which indicates the cluster resource state is not ready. The autoscaler skips the rest of the logic.
+  - **Dashboard**: the local Dashboard process is started by KubeRay pointing to the local GCS. Because it queries the passive local GCS, the dashboard UI initially displays a single-node cluster (only itself) with no active jobs or actors. Since the standby pod is Not Ready, its dashboard API is completely unreachable by external clients with the current k8s service setup. 
+  - **Job API**: with right k8s service configuration, the `ray job submit` request should not be able to reach the passive head.
```
edoakes: We should not rely on this as the only mechanism. Users may submit job requests in a different way (e.g., exec'ing into the pod and running the job submit command there)

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,227 @@
...
+  - **Serve Controller**: The Ray Serve Controller is a standard Ray actor. Since GCS in standby mode has no active workers and no running jobs, no Serve Controller is ever spawned on the standby head.
```
edoakes, line 288: It doesn't matter if the serve controller isn't run on the standby head, the serve API endpoints in the dashboard on a passive head could still make requests to the active head

Most likely I think we need a shared bit for API server components to check if they are the active or passive and fail requests when passive. This can be used for serve API & jobs API

edoakes: We could even abstract it into a utility/decorator that performs the check & rejects requests with an appropriate status code

YoyinZyc: I may need more clarification on this. what do you mean about "serve API endpoints in the dashboard on a passive head could still make requests to the active head"? 

And you are suggesting that whenever there is an unexpected request hit Job API or Serve API, it should ask its local gcs to check if it is leader or passive. If it is passive, reject the request. Am I understanding correctly?

YoyinZyc: also in terms of the decorator, are you saying something on gcs side like a rpc or API for leadership status checking?

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,325 @@
...
+- **Passive Node (Standby)**: Operates in a restricted passive mode. It runs the GCS gRPC server in a read-only state (to satisfy health checks) but defers loading cluster state and isolates itself from external traffic routing. It continuously polls the lease status.
+
+For Ray on Kubernetes, these two pods will be scheduled on different nodes by leveraging `podAntiAffinity` configuration to prevent a single node failure from taking down both the active and passive heads.
+
+The **Promotion Protocol** is designed to be non-blocking and safe:
+1. **Background Polling**: A dedicated thread continuously polls the lease status via the Kubernetes API, completely isolating blocking network I/O from the main ASIO event loop.
+2. **Fail-Fast on Leadership Loss**: Continues to poll the lease state post-promotion. If lease ownership lapses or renewal requests cross the expiration deadline, the leader immediately self-terminates (`RAY_LOG(FATAL)`) to protect cluster state consistency and avoid split-brain scenarios.
```
MengjinYan, line 86: Just to confirm, in this case, all the components on the head node will be terminated and then restarted, is that correct?

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,325 @@
...
+## Compatibility, Deprecation, and Migration Plan
+
+- **Head Node Components Compatibility**: 
+  To ensure a highly available Active-Passive architecture, all background processes running on the Ray Head Node must operate safely in a multi-head environment.
+
+  **1. Single-container co-located components**:
+  These components are running in the same container as different subprocesses. The gcs termination caused by lease lost will terminate other processes.
+  - **GCS Server**
+    Safe on Standby. Global tracking managers are instantiated empty; state hydration is deferred until promotion. The rpc server can only answer read requests in passive mode. All mutation requests should not be accepted and the rejection should be handled gracefully on the client side. (With the proper traffic routing, the requests will be sent to the active head only. But it is another layer of protection to avoid any potential split brain issue.)
```
MengjinYan: > The rpc server can only answer read requests in passive mode

Nit: It could be implementation detail, but I think it might be better just to identity the endpoints that needs to be called and allow them instead of open all the read endpoints.

reps/2026-04-23-active-passive-head.md
```diff
@@ -0,0 +1,325 @@
...
+      - **Passive State**: The Raylet continuously sends heartbeats to the local GCS to keep its node liveness. 
+      - **Promotion**: During promotion, the Raylet should register itself and  physically writes it to Redis.
+  - **Dashboard**: the local Dashboard process is started by KubeRay pointing to the local GCS. Because it queries the passive local GCS, the dashboard UI initially displays a single-node cluster (only itself) with no active jobs or actors. Since the standby pod is Not Ready, its dashboard API is completely unreachable by external clients with the current k8s service setup. 
+  - **Job API**: with right k8s service configuration, the `ray job submit` request should not be able to reach the passive head.
+  - **Serve Controller**: The Ray Serve Controller is a standard Ray actor. Since GCS in standby mode has no active workers and no running jobs, no Serve Controller is ever spawned on the standby head.
+  - **Dashboard Agent**: Runs locally and reports metrics to the co-located passive GCS. Upon promotion, it continues reporting to the now-promoted local GCS. Ready to initialize runtime environments if new jobs are submitted post-failover.
```
MengjinYan: nit: We might also want to disable event generation in the passive mode. There might not be many events generated but disabling it as a safe guard

reps/2026-04-23-active-passive-head.md
YoyinZyc: There are basically two containers running on head pod; For components like raylet which are in the same container with gcs, will instantly be killed together once gcs fails. For autoscaler which is running as a sidecar, it has its own lifecycle. It will watch the gcs state, if the gcs is passive, it should safely suppress its scaling logic.

Also if the node permanently evicts the previously head pod, everything will be recreated on another pod.

reps/2026-04-23-active-passive-head.md
YoyinZyc: I think the event generation is controlled by some envs(like `RAY_enable_ray_event` or `RAY_DASHBOARD_AGGREGATOR_AGENT_PUBLISH_EVENTS_TO_GCS `). Is it possible the re-enable the event generation at run time after promotion? Or are you referring to other approach to disable the event generation?

reps/2026-04-23-active-passive-head.md
YoyinZyc: Ah I see, added two more metrics for lease renew latency and failure count.

reps/2026-04-23-active-passive-head.md
YoyinZyc: At least `GetClusterID` and healthcheck rpc have to be enabled on passive. I may uncover more during the implementation.

reps/2026-04-23-active-passive-head.md
MengjinYan: I wasn't referring to any specific mechanism but more of a high level safe guard. 

You are right. The configurations are all specified at the start time. We probably need to add an additional config in the aggregator agent to control whether to emit events at the runtime.

YoyinZyc: Then the most intuitive approach is to ask the dashboard agent to watch and poll the leadership status periodically (in low frequency, 5s) and emit event post-promotion only.

reps/2026-04-23-active-passive-head.md
YoyinZyc: Ok changed to using `EnableActivePassiveHead` under `gcsFaultToleranceOptions` to gate the enablement of the feature.

reps/2026-04-23-active-passive-head.md
edoakes: > you are suggesting that whenever there is an unexpected request hit Job API or Serve API, it should ask its local gcs to check if it is leader or passive. If it is passive, reject the request. Am I understanding correctly?

Yes that's right.

> also in terms of the decorator, are you saying something on gcs side like a rpc or API for leadership status checking?

I am saying we could have a shared utility or python decorator in the dashboard logic that automatically performs this check before accepting any requests

edoakes (approved):
Looks good to me as a starting point for this feature; I'm sure there will be some more low level details to iron out as we go.

PR #69, opened 2026-07-13
REP: Ray Sandboxing with gVisor

andrewsykim:
This REP outlines a proposal to support a native experimental library for managing sandbox environments from Ray.

### DIFF
--- a/reps/2026-07-14-ray-sandboxing.md
+++ b/reps/2026-07-14-ray-sandboxing.md
@@ -6,7 +6,7 @@ Ray is emerging as the de facto orchestrator for Reinforcement Learning (RL) fra
 
 Currently, running untrusted code directly on Ray worker nodes poses severe security and operational risks, including arbitrary host access, data leakage, and cluster instability. To avoid these risks, frameworks often delegate execution to external sandbox services or third-party APIs. However, this externalized approach fragments the Ray ecosystem with ad-hoc abstractions and increases orchestration complexity.
 
-This proposal addresses these challenges by introducing an (experimental) Ray Sandbox library to programmatically manage isolated sandbox environments. The `ray.experimental.sandbox` library will provide a common denominator abstraction for secure sandboxing with unified resource scheduling in Ray. The initial implementation will focus solely on a gVisor-based backend (`runsc`) operating directly on Ray worker nodes for fast sandbox startup, strong kernel isolation, and dense bin packing of execution environments.
+This proposal addresses these challenges by introducing an (experimental) Ray Sandbox library to programmatically manage isolated sandbox environments. The library will provide a common denominator abstraction and building blocks for secure sandboxing with unified resource scheduling in Ray. The initial implementation will focus solely on a gVisor-based backend (`runsc`) operating directly on Ray worker nodes for fast sandbox startup, strong kernel isolation, and dense bin packing of execution environments.
 
 ### Should this change be within `ray` or outside?
 
@@ -28,7 +28,10 @@ Within `ray` as an experimental library (`ray.experimental.sandbox`).
 
 ### Overview
 
-`ray.experimental.sandbox` introduces an abstraction layer between Ray workloads (tasks, actors, or driver scripts) and underlying container or micro-VM runtimes for sandbox environments.
+Ray will introduce the following experimental APIs to provide the following building blocks for sandboxing within Ray:
+1. A `ray.experimental.SandboxRuntime` API - this is a common interface for managing local sandbox environemnts. For the initial version, we will only support gVisor as the sandbox runtime. 
+2. A `ray.experimental.Sandbox` Ray actor that wraps `SandboxRuntime` and provides a high-level API for creating and managing sandbox environemnts. The main purpose of this API is to manage scheduling and lifecycle of sandboxes using familiar Ray Core APIs.
+3. A Modal-like library (`ray.experimental.sandbox`) that manages `ray.experimental.Sandbox` actors for sandboxes.
 
 ```
 +-------------------------------------------------------------------+
@@ -39,19 +42,23 @@ Within `ray` as an experimental library (`ray.experimental.sandbox`).
                                   v
 +-------------------------------------------------------------------+
 |                      ray.experimental.sandbox                     |
-|  +---------------------+  +-----------------+  +-----------------+|
-|  |   Sandbox           |  |  SandboxPool    |  |  ExecResult     ||
-|  +---------------------+  +-----------------+  +-----------------+|
+|                   (High-Level Modal-Like Library)                 |
 +-------------------------------------------------------------------+
                                   |
                                   v
 +-------------------------------------------------------------------+
-|                          SandboxEnv Interface                     |
+|                     ray.experimental.Sandbox                      |
+|             (Ray Actor wrapping `SandboxRuntime`)                 |
 +-------------------------------------------------------------------+
                                   |
                                   v
 +-------------------------------------------------------------------+
-|                         GVisorSandboxEnv                          |
+|                     SandboxRuntime Interface                      |
++-------------------------------------------------------------------+
+                                  |
+                                  v
++-------------------------------------------------------------------+
+|                        GVisorSandboxRuntime                       |
 |            (Manages host `runsc` spec & lifecycle)                 |
 +-------------------------------------------------------------------+
                                   |
@@ -65,9 +72,119 @@ Within `ray` as an experimental library (`ray.experimental.sandbox`).
 +-------------------------------------------------------------------+
 ```
 
-### Python API Design
+### API Design
 
-The `ray.experimental.sandbox` library will provide an intuitive, high-level abstraction for creating and managing isolated sandbox environments. The initial implementation focuses solely on gVisor (`runsc`), providing node-local container sandboxing optimized for fast startup times (sub-100ms) and high-density bin packing of concurrent execution tasks.
+#### ray.experimental.SandboxRuntime
+
+This is the lowest-level API for sandboxing and should only be used by higher-level Ray libraries (see below) or power users who need fine-grained control over the sandbox environment.
+
+Below is the interface for `SandboxRuntime`. Ray will only support a gVisor backend for the initial release.
+```python
+class SandboxRuntime(ABC):
+    """Low-level interface for managing local sandbox runtime environments."""
+
+    @abstractmethod
+    def create(self, config: SandboxConfig) -> str:
+        """Provision the sandbox instance and return unique instance ID."""
+        pass
+
+    @abstractmethod
+    def exec(self, instance_id: str, command: str, timeout: int = None, env: dict = None) -> ExecutionResult:
+        """Execute a command inside the specified sandbox."""
+        pass
+
+    @abstractmethod
+    def upload_file(self, instance_id: str, local_path: str, remote_path: str) -> None:
+        """Copy local file into the sandbox."""
+        pass
+
+    @abstractmethod
+    def download_file(self, instance_id: str, remote_path: str, local_path: str) -> None:
+        """Copy file from the sandbox to local."""
+        pass
+
+    @abstractmethod
+    def delete(self, instance_id: str) -> None:
+        """Clean up and terminate the sandbox instance."""
+        pass
+```
+
+#### ray.experimental.Sandbox
+
+`ray.experimental.Sandbox` is a Ray actor that wraps `SandboxRuntime` and provides a high-level API for creating and managing sandbox environments. The main purpose of this actor is to manage scheduling and lifecycle of sandboxes using familiar Ray Core APIs.
+
+Below is the interface for `Sandbox` actor:
+```python
+@ray.remote
+class Sandbox():
+    """Ray actor interface for managing scheduling and lifecycle of an isolated sandbox."""
+    
+    def __init__(self, runtime: SandboxRuntime):
+        self.runtime = runtime
+
+    def exec(self, command: str, timeout: int = None, env: dict = None) -> ExecutionResult:
+        """Execute a command inside the sandbox."""
+        pass
+    def upload_file(self, local_path: str, remote_path: str) -> None:
+        """Copy local file into the sandbox."""
+        pass
+
+    @abstractmethod
+    def download_file(self, remote_path: str, local_path: str) -> None:
+        """Copy file from the sandbox to local."""
+        pass
+
+    @abstractmethod
+    def delete(self) -> None:
+        """Clean up and terminate the sandbox instance."""
+        pass
+```
+
+
+#### ray.experimental.sandbox
+
+The `ray.experimental.sandbox` library will provide a high-level abstraction for creating and managing isolated sandbox environments.
+Under the hood, it will use a `ray.experimental.Sandbox` actor for scheduling and resource assignment and `SandboxRuntime` to manage the sandbox runtime
+using gVisor. For the initial version of this API, we will provide a simple `create` and `create_async` API that returns a `SandboxHandle`. Below are usage example:
+
+```python
+def create(
+    config: Optional[SandboxConfig] = None,
+    **kwargs,
+) -> SandboxHandle:
+    """Create a sandbox environment."""
+    pass
+
+
+def create_async(
+    config: Optional[SandboxConfig] = None,
+    **kwargs,
+) -> "ray.util.async_func[SandboxHandle]":
+    """Create a sandbox environment asynchronously."""
+    pass
+
+
+class SandboxHandle():
+    """High-level handle interface for interacting with a Ray sandbox."""
+
+    def exec(self, command: str, timeout: int = None, env: dict = None) -> ExecutionResult:
+        """Execute a command inside the sandbox."""
+        pass
+
+    def upload_file(self, local_path: str, remote_path: str) -> None:
+        """Copy local file into the sandbox."""
+        pass
+
+    def download_file(self, remote_path: str, local_path: str) -> None:
+        """Copy file from the sandbox to local."""
+        pass
+
+    def delete(self) -> None:
+        """Clean up and terminate the sandbox instance."""
+        pass
+```
+
+### Examples
 
 #### Example: Basic Sandbox Creation and Command Execution
 
@@ -112,45 +229,54 @@ with sandbox.create() as sb:
     sb.download_file(remote_path="/tmp/output.json", local_path="results.json")
 ```
 
----
 
-### SandboxEnv Interface
-
-All sandbox implementations will implement a generic `SandboxEnv` interface:
+#### Example: Create a `ray.experimental.Sandbox` actor
 
 ```python
-class SandboxEnv(ABC):
-    @abstractmethod
+import ray
+
+sandbox = ray.experimental.Sandbox.options(
+    num_cpus=1,
+    memory=1e9
+).remote(
+    timeout=1800
+)
+
+sandbox.write_file.remote("/workspace/main.py", "print('Hello from Ray Sandbox!')")
+result = ray.get(sandbox_client.exec.remote("python3 /workspace/main.py"))
+```
+
+#### Example: Create a custom actor using ray.experimental.SandboxRuntime
+
+```python
+import ray
+from ray.experimental.sandbox_runtime import SandboxRuntime
+
+@ray.remote
+class CustomSandbox():
+    def __init__(self):
+      self.runtime = SandboxRuntime()
+      self.instance_id = None
+
     def create(self, config: SandboxConfig) -> str:
         """Provision the sandbox instance and return unique instance ID."""
+        # custom create logic using self.runtime.create()
         pass
 
-    @abstractmethod
     def exec(self, instance_id: str, command: str, timeout: int = None, env: dict = None) -> ExecutionResult:
         """Execute a command inside the specified sandbox."""
+        # custom exec logic using self.runtime.exec()
         pass
 
-    @abstractmethod
-    def upload_file(self, instance_id: str, local_path: str, remote_path: str) -> None:
-        """Copy local file into the sandbox."""
-        pass
-
-    @abstractmethod
-    def download_file(self, instance_id: str, remote_path: str, local_path: str) -> None:
-        """Copy file out of the sandbox to local filesystem."""
-        pass
-
-    @abstractmethod
-    def terminate(self, instance_id: str) -> None:
-        """Destroy the sandbox instance and release underlying resources."""
-        pass
+sandbox = CustomSandbox.remote()
+result = ray.get(sandbox.exec.remote("echo 'Hello from Ray Sandbox!'"))
 ```
 
 ---
 
 ### Sandboxing with gVisor
 
-The `gvisor` sandbox environment implementation (`GVisorSandboxEnv`) uses gVisor's OCI runtime (`runsc`) to execute untrusted code in lightweight, kernel-isolated sandboxes directly on Ray worker nodes.
+The `gvisor` sandbox environment implementation (`GVisorSandboxRuntime`) uses gVisor's OCI runtime (`runsc`) to execute untrusted code in lightweight, kernel-isolated sandboxes directly on Ray worker nodes.
 
 #### Motivation: Fast Startup & Dense Bin Packing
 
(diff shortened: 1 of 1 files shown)


reps/2026-07-14-ray-sandboxing.md
```diff
@@ -0,0 +1,147 @@
...
+# dummy script
+script = """import time
+if __name__ == "__main__":
+    time.sleep(1)
+"""
+path = "/tmp/test.py"
+
+# initialize the actor as a "sandboxed" actor
+sandboxed_actor = SandboxedActor.options(
+    sandbox_env=SandboxEnv(backend="docker", image_uri="rayproject/ray:2.56.0")
```
rueian: I wonder if we should allow users to run any user-defined actor on a sandbox, or should we just expose a sandbox api for them to interact with the sandbox?

A sandbox api could be like
```
sandbox = Sandbox(restore_from=snapshot)

sandbox.exec(...)
sandbox.upload(...)
sandbox.download(...)
sandbox.snapshot(...)
```

reps/2026-07-14-ray-sandboxing.md
```diff
@@ -0,0 +1,147 @@
...
+).remote(script, path)
+
+# execute_code runs in the sandbox 
+result = ray.get(sandboxed_actor.execute_code.remote())
+```
+
+## Sandboxed Actors
+
+When a Ray actor specifies a SandboxEnv, the Raylet will invoke the default worker 
+(ray._private.workers.default_worker.py) inside the sandbox environment to run the actor code.
```
SongGuyang: We can not guarantee the default worker could run into the user-defined images. Actually,  most of the sandbox images are from open source, and we also can not install the python environment and pip packages before start the sandbox in consideration of efficiency. So, one feasible approach, actor or task run outside the sandbox environment, and use "exec" to execute user command?

SongGuyang:
By the way, can we design for two scenarios:
- Deploy sandbox inside the cluster of RL. We can use the Ray API.
- Deploy sandbox outside the cluster of RL. Maybe we need to involve Ray Serve?

andrewsykim:
> Deploy sandbox outside the cluster of RL. Maybe we need to involve Ray Serve?

@SongGuyang can you share more details for what you had in mind here. We are already seeing Ray users running sandbox environments outside the Ray cluster but orchestrating sandbox creation / execution from inside the Ray cluster during RL. I'm not sure what we could change in Ray to make this easier as the sandbox lifecycle is entirely outside the control of Ray and it's up to users to integrate with out-of-cluster sandboxes.

reps/2026-07-14-ray-sandboxing.md
andrewsykim: > We can not guarantee the default worker could run into the user-defined images.

Yes, in this approach, the sandbox env would need to have whatever Ray dependencies are required to run the default worker process such that actors can run there.

> So, one feasible approach, actor or task run outside the sandbox environment, and use "exec" to execute user command?

Users can already exec commands to external sandboxes from another Ray actor. Are you suggesting Ray should have native support for specific sandboxing solutions so it can exec commands on behalf of users?

pcmoritz:
@andrewsykim Thanks a lot for putting this together! Some food for though: Starting actual Ray actors inside the sandbox might not be a great idea for two reasons: (i) as @SongGuyang pointed out, many sandbox images won't have Ray installed and in general it would be great to not have assumptions on the sandbox images and (ii) the sandboxes should be isolated and exposing the Ray API inside the sandbox will make it hard to do strong isolation (e.g. even returning pickled results from inside the actor and unpickling on the caller).

One alternative design is to re-use the Ray Control Plane / Infra for scheduling the sandboxes to make it convenient for users to use the sandboxes (and integrating with e.g. gvisor or other infra), but then not run actors inside the sandbox. Note that we could still expose the sandbox as an actor to users, and just run a proxy actor outside of the sandbox.

andrewsykim:
@pcmoritz thanks for the feedback. 

> One alternative design is to re-use the Ray Control Plane / Infra for scheduling the sandboxes to make it convenient for users to use the sandboxes (and integrating with e.g. gvisor or other infra), but then not run actors inside the sandbox. Note that we could still expose the sandbox as an actor to users, and just run a proxy actor outside of the sandbox.

This is an interesting alternative and doesn't feel too different from what is already proposed. The main difference is using the proposed "sandbox env agent" to proxy actor calls into the sandbox instead of using it to spin up a default worker there. But I'm curious if executing actor methods through an external proxy actor (without running a Ray worker inside the sandbox) is actually doable. Do you happen to know if that's supported?"

reps/2026-07-14-ray-sandboxing.md
andrewsykim: This is what users are doing already -- calling an external API to provision sandboxes and proxy commands to them. The main difference would be Ray providing some common API to abstract away implementation details across multiple sandbox implementations. I'm not sure how much value that adds on its own. @SongGuyang is this what you had in mind in your other comment? @pcmoritz  curious about your thoughts here as well.

andrewsykim:
> (i) as @SongGuyang pointed out, many sandbox images won't have Ray installed and in general it would be great to not have assumptions on the sandbox images and (ii) the sandboxes should be isolated and exposing the Ray API inside the sandbox will make it hard to do strong isolation (e.g. even returning pickled results from inside the actor and unpickling on the caller).

Re i), yes having Ray installed in the sandbox env will be required. I don't think there's a way around this requirement (assuming we want Ray Actors running in sandboxes with little to no change).  

Re ii), I agree that this approach will not accomplish the same isolation guarentees as hosting sandboxes completely outside of Ray clusters. However, I think it's doable to put some guardrails in place that make the isolation good eonugh for a lot of RL use-cases without sacrificing the utility of native sandboxing. For example, we will probably allow returning results from the sandboxed actors back to the caller as you mentioned, but we would not grant sandboxed actors access to the object store or network access to any other Raylets

pcmoritz:
Having Ray installed in the sandbox seems like a too heavy requirement from all the sandboxing / environments I have seen. Instead it would be most useful to expose running a (bash / sh) command in the sandbox instead of a full Ray task, together with reading the stdout (together with other APIs like uploading / downloading files from the sandbox -- similar to what @rueian suggests). Modal sandboxes (https://modal.com/products/sandboxes) are very popular, we could borrow the API from them.

If we have such a sandboxing implementation for different backends (e.g. raw k8s / gvisor / etc.) with automatic scheduling and integration with labels / resources on a Ray cluster, that might be more than enough value actually and exactly what people want. Given that, maybe we should also discard the proxy actor idea, since if people actually want a proxy actor, they could always instantiate it themselves (and then the actor creates the sandbox).

andrewsykim:
> Having Ray installed in the sandbox seems like a too heavy requirement from all the sandboxing / environments I have seen

We wouldn't need all Ray dependencies for this to work, just enough to run CoreWorker process. Do you think even that minimal footprint is too heavy? Is the concern mainly about image/dependency size, or the toil on users having to customize their sandbox environments to include it?

> If we have such a sandboxing implementation for different backends (e.g. raw k8s / gvisor / etc.) with automatic scheduling and integration with labels / resources on a Ray cluster, that might be more than enough value actually and exactly what people want. Given that, maybe we should also discard the proxy actor idea, since if people actually want a proxy actor, they could always instantiate it themselves (and then the actor creates the sandbox).

In this model, Ray provides a native sandboxing library rather than extending the Actor API to run inside sandboxes. I don't feel like this is fundamentally different from what users are doing today (calling external APIs/libraries to manage sandboxes). The primary benefit would be Ray offering an opinionated abstraction with default implementations out of the box. @pcmoritz are you suggesting that alone provides enough value? If so, happy to adapt the REP to propose a native sandboxing library (as opposed to extending Ray Actors).

pcmoritz:
I think being able to run existing images without modifications is important (there are many such images e.g. for certain benchmarks and SWE like tasks that people want to just use out of the box without modifying), so any assumption we make about the image will be problematic and the best way to circumvent such assumptions is to use just executing commands as the interface to the sandbox / image.

While I can't guarantee that this will be widely used, based on how often sandboxing is coming up in user / customer conversations, it seems likely that it will be popular (as long as it is broadly compatible with people's infrastructure e.g. support Ray running on any K8s cluster without strong assumptions on the cluster). Currently the managed sandboxing solutions are pretty popular since getting everything right is a little tricky, but if we have a good solution that works well out of the box it seems likely that it will be adopted since the managed solutions are typically way more expensive than they should be.

The best way to find out is probably to build an MVP and circulate it with users, that way we can also iterate on the design and API.

andrewsykim:
thanks for the feedback, I will update the REP to reflect the new direction (native library for sandboxing). Will also try to build a POC in parallel to get a better idea of technical details and API design.

andrewsykim:
Updated the REP to reflect the new direction of introducing a native library in Ray for sandboxing. Also opened a draft PR with a rough prototype of this working with Kubernetes Pods for sandboxing https://github.com/ray-project/ray/pull/64964

andrewsykim:
After some offline discussions with @richardliaw and Pradeep, I changed the proposal to now include both a gVisor and kubernetes backend for sandbox environments. The rationale for starting with gVisor is that many sandboxing use-cases require really fast startup and dense packing, which kubernetes will not be able to support out of the box. gVisor running inside Ray containers is a viable solution with sufficient isolation that can meet the two requirements. We will still support the Kubernetes option for use-cases that may not require really fast startup and dense packing.   
 
I also updated the POC PR to include a working gVisor backend for reference https://github.com/ray-project/ray/pull/64964

robertnishihara (commented):
Thanks for opening this!

A few questions (for my understanding).
1. **Touch points with Ray:** Now that we're no longer running Ray actors inside the sandbox, does the `sandbox` class implementation involve any "Ray" stuff? Or can it be conceptualized purely as a Python utility that spins up a container in a subprocess and communicates with it? Aside from `SandboxPool`, which parts of this PR are intertwined with Ray versus could be implemented as a standalone library?
2. **SandboxPool:** I'd lean toward removing the `SandboxPool` at least initially because it seems like something that is best explicitly controlled by the user.
    - I imagine that in the common case, you will have 1 actor managing ~100 sandboxes, is that right? If so, then the application logic that runs in the actor and interfaces with the sandbox will need to be multithreaded or async and might get kind of complex. So it might make sense to leave the `SandboxPool` implementation in the application logic so users have more control.
    - Given how quickly we can spin up sandboxes, how common is it to reuse the sandbox?
    - Some magic is happening when you pass a `SandboxPool` into a task in order to colocate the task with the `SandboxPool` actors. This scheduling policy is something that users may want control over.
3. **gVisor versus Kubernetes:** Can you elaborate on the pros / cons of the `gVisor` version versus the `Kubernetes` version?
    - How portable are they and what requirements / permissions do they require from the underlying infra or Kubernetes cluster? 
    - Lack of support for custom container images seems like a likely showstopper. What are the obstacles there?
    - Is it correct that in the **gVisor** case the sandbox is created locally on the same machine, whereas in the **Kubernetes** case, the sandbox may be created anywhere in the cluster? This seems like a sizable behavior difference. Is one preferable?
4. **Resources:** When a Ray actor creates subprocesses, normally the resource bookkeeping is done at the level of the parent actor. E.g., if I create an actor that starts an SGLang subprocess that uses a GPU, the parent actor should request a GPU. This is important because the subprocess resource requirements need to be taken into account for placement of the parent actor and need to be reserved in a gang-scheduling fashion along with the parent actor's resource requirements. All of this is to say that I'm not sure it makes sense to do resource bookkeeping when sandboxes are created / destroyed, the parent actor should do it instead.

andrewsykim:
> 1. Touch points with Ray: Now that we're no longer running Ray actors inside the sandbox, does the sandbox class implementation involve any "Ray" stuff? Or can it be conceptualized purely as a Python utility that spins up a container in a subprocess and communicates with it? Aside from SandboxPool, which parts of this PR are intertwined with Ray versus could be implemented as a standalone library?

Yes, in the current state it is more or less a Python library that uses gVisor to spin up isolated processes. However, it will be interwined with Ray for scheduling and resource allocation. In other words, resource assignment for sandboxes will use resources available in the Ray cluster, meaning users that use Ray for RL orchestration can also use Ray for sandbox orchestration.  This also implies that Ray's autoscaling capabilities extends to sandboxes automatically.

> 2. SandboxPool: I'd lean toward removing the SandboxPool at least initially because it seems like something that is best explicitly controlled by the user.

I think I agree with you. Let's remove SandboxPool for now, we can always revisit and add this capability in the future based on user feedback. 

> 3. gVisor versus Kubernetes: Can you elaborate on the pros / cons of the gVisor version versus the Kubernetes version?

Your assessment is correct that gVisor backend creates local containers on the same machine, where-as kubernetes would create Pods on a Kubernetes cluster (could be same or different k8s cluster than the Ray cluster). The main benefit of the Kubernetes backend is flexibility to fully customize sandbox environments. But thinking about it further, using Kubernetes pods for sandboxes would mean we can't use Ray resources to schedule sandboxes and we break the "unified scheduler" for RL and sandboxing that we would get with gVisor. Also, the Kubernetes approach, while really flexible, is unlikely to meet requirements for fast startup unless we have some warm pooling capability (like SandboxPool). This approach also requires granting the Raylet access to create Pods which may be not desirable in some cases for security reasons. With these things considered, I propose we build the first experimental release of sandboxing only using gVisor and revisit the kubernetes implementation in the future. What do you think?

> Resources: When a Ray actor creates subprocesses, normally the resource bookkeeping is done at the level of the parent actor. E.g., if I create an actor that starts an SGLang subprocess that uses a GPU, the parent actor should request a GPU. This is important because the subprocess resource requirements need to be taken into account for placement of the parent actor and need to be reserved in a gang-scheduling fashion along with the parent actor's resource requirements. All of this is to say that I'm not sure it makes sense to do resource bookkeeping when sandboxes are created / destroyed, the parent actor should do it instead.

Are you implying that a Ray actor that calls the sandbox library should only be able to use resources for sandboxes up to what is assigned to the actor? e.g. an actor with "CPU: 4" can only create 4 sandboxes with 1 CPU each?  My initial thinking is that sandboxes can be scheduled anywhere on the Ray cluster and doesn't need to be co-located to the actor that creates the sandbox. It's more akin to a Ray actor being able to create another Ray actor. I think this pattern is also more common in RL where the RL framework would create hundreds or thousands of sandboxes from 1 orchestrating actor. Let me know if I'm misunderstanding your question.

robertnishihara:
> My initial thinking is that sandboxes can be scheduled anywhere on the Ray cluster and doesn't need to be co-located to the actor that creates the sandbox.

Thinking about this a bit more, I think someone using a sandbox "service" will think about just two components.
1. The client that sends commands to the sandbox.
2. The sandbox itself.

These definitely do not need to be colocated on the same machine.

However, I think we should conceptualize the primitives we're providing here in Ray not as a "sandbox service" but rather "primitives that can be used to build a sandbox service (possibly scoped to within your application)".

If you're building a sandbox service (instead of using one out of the box), you need to solve for some additional things
1. Sandbox scheduling / placement
2. Quick sandbox spin-up
3. Forwarding commands from clients to sandboxes

In the context of Ray, it makes sense to reuse Ray's actor scheduling mechanism for (1). For (2), it makes sense to start sandboxes as subprocesses but to have multiple sandboxes per "manager actor" to avoid actor startup costs. For (3), I think it makes sense for clients to talk to the manager actor and for the manager actor to forward the command to the sandbox subprocess.

All of this is clunky if you are expecting a sandbox API, but powerful if you are building a sandbox service.

So what do we need to provide?
1. Nothing for the client. For now, that can just be Python logic that calls methods on the manager actor.
2. For the manager actor, we should provide a reference implementation, but no API.
     - It should be a multi-threaded or async Ray actor that uses the sandbox API from this REP to spin up actors and forward commands to them.
     - I think the reference implementation should also do all of the resource bookkeeping and would encode things like the maximum number of sandboxes it can run concurrently.
     - There is a large-ish space of interesting designs for this actor, and I suspect this is something that users will want control over, e.g., if there are no available resources, should we return an error or queue the sandbox creation?
4. For the sandbox itself, this REP provides the API for creating sandboxes as subprocesses and forwarding commands to them.

andrewsykim:
@robertnishihara  I think the prototype I implemented in https://github.com/ray-project/ray/pull/64964 is very close to what you are suggesting with some variations. When you create a Sandbox with some resource requirement, it will first schedule a regular Ray actor with the resource requiremnets of the sandbox (_SandboxActor) that then invokes `runsc run` to run the sandbox. So in this implementation we are using Ray for scheduling / placement of the sandbox itself. When a client proxies commands to the sandbox, it will first send the command to the remote Ray actor (likely on a different node), which will then invoke `runsc exec` on the local sandbox.

The one thing that is missing in that PR is a reference implementation of the "manager actor" and I agree it is something we can include in Ray as a reference implementation and leave it up to users on how to use it. But I think it is also acceptable if users want to call the sandbox API directly from their clients instead of a "manager actor".

pcmoritz:
On the resource question: There also needs to be a mechanism to communicate memory and cpu limits to the sandbox, these are different from Ray resource request, since they will be strongly enforced (if the sandbox supports it). It is worth thinking about the right API here (how do other sandbox providers expose this?). It is natural to make these arguments of `sandbox.create` like you do, but is it e.g. a good idea to use strings for the memory limits or is a different representation better?

pcmoritz:
Maybe we should just follow the convention in https://modal.com/docs/sdk/py/latest/Sandbox#create for this and e.g. support memory limits as int in MiB -- the full (request, limit) API could be supported via a modal API compatible API adapter that spins up the managing actor with the right requests and then exposes the limit on the sandbox.

andrewsykim:
Noted to revisit the API for resourcing. I think this is something we can hash out in the implementation PR or during experimental phase of the library.

andrewsykim:
> There also needs to be a mechanism to communicate memory and cpu limits to the sandbox, these are different from Ray resource request, since they will be strongly enforced (if the sandbox supports it)

gVisor supports specifying resources via the OCI bundle you pass down to `runsc`. In the current prototype, the resources in the OCI bundle match that of the scheduled actor. This will likely leverage the same mechanisms for "writable cgroups" that are already used for Ray's resource isolation feature.

robertnishihara:
> The one thing that is missing in that PR is a reference implementation of the "manager actor" and I agree it is something we can include in Ray as a reference implementation and leave it up to users on how to use it. But I think it is also acceptable if users want to call the sandbox API directly from their clients instead of a "manager actor".

@andrewsykim does that mean that there is 1 actor per sandbox? Wouldn't that be too many actors? I'm assuming the sandboxes can be lighter weight (and faster to start) than an actor (which is maybe incorrect), and so users would want to have many sandboxes per actor.

andrewsykim:
Yes, at least in the current prototype, we schedule 1 Ray actor per sandbox. The purpose of the actor is to use Ray primitives for scheduling, resource assignment, and distributed communication from the client or "manager actor" to the sandbox to execute commands.

I am not sure if this is too many actors, even with 1 actor per sandbox, we can still achieve hundreds to thousands of sandboxes per node (depending on the node size). I think the other workstreams for improving actor scalability will help as well. In the future we can maybe optimize this further and run 1 actor per node to manage all local sandboxes and resource assignment, but maybe it's premature right now?

robertnishihara:
Ok, here's a snapshot of my current thoughts.

**Target users / abstraction:** Like before, I think we should optimize for "providing primitives for building a sandbox service" as opposed to "a sandbox service" itself. This generally means: (1) optimize more for power users who are comfortable writing a lot of application logic, (2) provide control instead of doing things under the hood. Examples of things users will want to control are scheduling, policies around autoscaling or pausing / resuming / migrating sandboxes.

I was previously concerned that 1 actor per sandbox would be too much overhead and that in order to do many sandboxes per actor, we would need to expose a different abstraction. However, now I am less concerned about the actor overhead / startup time, since I think the sandboxes will generally be more heavyweight / slower (e.g., for agentic tasks). So I think 1 actor to 1 sandbox makes sense.

**API:** For the API, I can think of two main options: (1) mimic an existing sandbox API like Modal, (2) make it more Ray native. They are not completely incompatible, but I would bias toward making it more Ray native. Why? A lot of the benefit of running on a Ray cluster is the extra control we can expose (e.g., over scheduling / placement), which normal sandbox services are not solving for. For example, I want to be able to create the sandbox inside of a placement group. I want to control placement with labels, etc. We should reuse Ray's existing APIs and familiar patterns for this.

I'm imagining someone building an RL application that creates training workers, inference workers, sandboxes, etc. It's natural / intuitive (to me) to be able to use the same APIs / tools to control placement of all of these in the Ray cluster.

Even more in this direction, I think we should just make the entire thing an actor so we can use actor scheduling capabilities and not have to mirror the actor API in the sandbox API?

```python
sandbox_client = ray.experimental.Sandbox.options(
    num_cpus=1,
    memory=1e9
).remote(
    backend="gvisor",
    timeout=1800
)

sandbox_client.write_file.remote("/workspace/main.py", "print('Hello from Ray Sandbox!')")
result = ray.get(sandbox_client.exec.remote("python3 /workspace/main.py"))
```

**Suggested next steps:**
- Let me know if you agree with these points. If so, could you reflect them in the REP?
- For the purposes of approving the REP, I suggest you enumerate a minimal set of APIs / doc strings (for the `Sandbox` actor methods) that we want to start with. We will certainly add more as we go and modify / iterate on them as we learn more.

andrewsykim:
I really like the `ray.experimental.Sandbox` actor approach that you suggested, it's close to the initial version of the REP and I think it's a very natural / intuitive approach that Ray developers will be familiar with. I also agree with Philip's point in another thread for needing lower level interface for power users that want full control of sandbox lifecycle.

So sounds like we can actually introduce multiple APIs to start with:
1) `ray.experimental.sandbox.runtime` -> this is the lower level, local interface that translates to gVisor calls for sandboxes. It's the user's responsibility to manage actors that call this interface for gVisor sandboxes.
2) `ray.experimental.Sandbox` ->  a Ray actor that calls ray.experimental.sandbox.runtime. Allows for using Ray Core API for sandbox scheduling 
3) `ray.experimental.sandbox` -> a Modal-like API for Sandboxing, which would create the `ray.experimental.Sandbox` actor per sandbox (which calls `ray.experimental.sandbox.runtime` for gVisor sandboxes).

I'm getting the impression that we're converging towards starting with both 1) and 2).
The first is for power users that want full control of sandbox lifecycle and perform custom logic like health checking or managing multiple sandboxes from a single actor. The second will be a good starting point for everyone else that wants to extend Ray API for sandboxing.
We can revisit 3) in the future.

If we're aligned on this approach, I can update the REP.

pcmoritz:
Thanks for updating the APIs! On a practical note, we should probably put everything into a common package like `ray.experimental.sandbox` which will be migrated to `ray.sandbox` once the API stabilizes to group things together. As the public API, this makes the most sense to me (to avoid mixing up `sandbox` and `ray.sandbox`):

`ray.sandbox.runtime` contains `SandboxRuntime` and all the implementations
`ray.sandbox` contains `SandboxActor`, which is the actor and `SandboxHandle` which is the handle
`ray.sandbox` contains `create` and `create_async` which is the main (?) user facing constructor and creates the SandboxHandle

And in the future there will be a modal compatible API in `ray.sandbox.modal` so people can do `from ray.sandbox import modal` and just use their existing code.

One thing that might make sense for the actual PR is to study if it is necessary to expose both the SandboxActor as well as the SandboxHandle as public APIs, or whether we can expose one thing there (e.g. a single RemoteSandbox class).

andrewsykim:
Good suggestions. I clarified package naming for GA and added a follow-up item to add a Modal compatible API for Ray sandboxing.

reps/2026-07-14-ray-sandboxing.md
```diff
@@ -0,0 +1,385 @@
...
+                                  |
+                                  v
++-------------------------------------------------------------------+
+|                      ray.experimental.sandbox                     |
+|                   (High-Level Modal-Like Library)                 |
++-------------------------------------------------------------------+
+                                  |
+                                  v
++-------------------------------------------------------------------+
+|                     ray.experimental.Sandbox                      |
```
pcmoritz: This needs to be updated to `ray.experimental.sandbox.Sandbox`

reps/2026-07-14-ray-sandboxing.md
```diff
@@ -0,0 +1,385 @@
...
+|   +-----------------------+       +-----------------------+       |
+|   | gVisor Sandbox 1      |       | gVisor Sandbox 2      |       |
+|   | (Isolated Execution)  |       | (Isolated Execution)  |       |
+|   +-----------------------+       +-----------------------+       |
++-------------------------------------------------------------------+
+```
+
+### API Design
+
+#### ray.experimental.SandboxRuntime
```
pcmoritz: Update to `ray.experimental.sandbox.runtime.SandboxRuntime` to be consistent

reps/2026-07-14-ray-sandboxing.md
pcmoritz: (there is a few other places below where it needs updating)

andrewsykim: ack will fix

pcmoritz (approved):

robertnishihara (approved):

PR #70, opened 2026-07-20
REP: Ray Batch Actor Scheduling

andrewsykim:


2026-07-20-ray-batch-actor-scheduling.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+### Implementation Details
+
+Under the hood, the batch context manager will signal the Core Worker to buffer actor creation requests instead of sending them to the GCS immediately. This allows the actor scheduler to optimize actor placement for co-location and locality, reducing overhead and improving overall scheduling throughput.
+
+To achieve this, we extend the actor creation workflow as follows:
+
+* **ray.batch() API**: The `ray.batch()` context manager signals to the Core Worker (e.g. Cython CoreWorkerProcess.GetCoreWorker().EnterActorBatch()) to enter batching mode for the current thread. Upon exit, it calls ExitActorBatch().
+* **Core Worker Actor Buffering**: Inside CoreWorker::CreateActor, if "batch mode" is enabled, the TaskSpecification for actors are pushed into a thread local buffer.
```
Yicheng-Lu-llll, line 54: Brainstorming here: what happens if actors need dependencies? Some actors' dependencies get resolved quickly, while others resolve very slowly (waiting for an upstream task to produce some object). 

This dependency requirement might be unusual for the RL case, but if it does happen, then for the actors still waiting on dependency resolution, we can only register them but cannot submit. For the ones whose dependencies are already resolved, we can submit. 

I'm wondering how we should handle this. Would we end up submitting one by one? It seems like we can only batch the Register phase but cannot control the Submission Execution, so my sense is that we should keep the Submission Execution as is.

2026-07-20-ray-batch-actor-scheduling.md
```diff
@@ -0,0 +1,82 @@
...
+    * ActorIDs are generated synchronously (as usual via the worker context deterministic generator) and returned to Python immediately, so the user has valid ActorHandles.
+    * When ExitActorBatch() is called, it flushes the buffer by invoking an RPC to the GCS server asynchronously to register all the actors in the buffer in one atomic operation (see RegisterActorBatchRequest below).
+* **GCS RPC Extension (RegisterActorBatch)**: A new RPC RegisterActorBatchRequest will be added to gcs_service.proto.
+    * GcsActorManager implements HandleRegisterActorBatch, which iterates through the batched tasks and invokes RegisterActor for each task locally in the GCS.
+    * Once all tasks are successfully registered in the backend storage, a single RPC reply is returned to the Core Worker.
```
Yicheng-Lu-llll, line 59: If I remember correctly, for named actors, we don't return the actor handle unless the register response comes back. But for normal actors, we don't wait.

My personal thinking is that the reason register phase exists, and we need to wait synchronously is we need to handle this case: if an actor is still waiting on dependency resolution and we do not register that actor first, then GCS does not know about this actor. But the actor handle may have already been given to others, and if the owner dies, those who hold the actor handles will wait forever.

So it makes sense to me that we should synchronously wait for all register responses. But we need to understand why we don't wait for normal actors. Is this for performance reasons? Will we cause a regression here?

Yicheng-Lu-llll (commented):
Thanks! left some nits.

2026-07-20-ray-batch-actor-scheduling.md
Yicheng-Lu-llll: or opportunistically batching

2026-07-20-ray-batch-actor-scheduling.md
```diff
@@ -0,0 +1,82 @@
...
+* **Submission Execution**: After the GCS responds to the RegisterActorBatch RPC, the Core Worker loops over the batch and calls actor_task_submitter_->SubmitActorCreationTask(task_spec) to push the tasks to the placement group / scheduling queue.
```
Yicheng-Lu-llll, line 60: Could we also clarify the behavior when some of the actors get scheduled but the rest fail due to insufficient resources? Though I think the behavior would be the same.

2026-07-20-ray-batch-actor-scheduling.md
```diff
@@ -0,0 +1,82 @@
+# Summary
+
+### General Motivation
+
+Ray currently processes actor creation requests one by one. For workloads that spawn hundreds or thousands of actors (e.g., large-scale RL), this sequential creation can become a bottleneck because each actor registration requires a separate RPC call to the GCS server, and the scheduling of actors may be suboptimal when evaluated on a per-actor basis.
```
Yicheng-Lu-llll, line 5: it's also worth mentioning that, particularly for named actors, we need to wait for the register RPC to come back before we can start the next one. Since the current call has to return first, this serialization ends up slowing everything down. Normal actors can do this concurrently. I would expect Batch really helps here.

2026-07-20-ray-batch-actor-scheduling.md
```diff
@@ -0,0 +1,82 @@
...
+@ray.remote
+class MyActor:
+    def __init__(self):
+        pass
+
+ray.init()
+
+# create 1000 actors in a batch
+with ray.batch():
+    actors = [MyActor.remote() for _ in range(1000)]
```
MengjinYan, line 44: I'm wondering if there should be a explicit API to add the actor to the batch and delaying returning the actor handle until the batches are registered, instead of assuming all the actors created under `ray.batch()` will be added to the batch and get the actor handles right away. 

The reason is that, currently Ray assumes that once we have the actor handle, we are able to call functions of the actor to submit actor tasks. And it is also possible that the user then call `ray.get()` to fetch the object reference returned from the actor task. This will cause the current implementation to hang because the `ray.get()` function will block until the actor task is executed, while the actor can only be registered after existing the context manager's block. 

And I think having an explicit API to add to batch and batch register to get the handle can be one of the solutions.

PR #75, opened 2026-09-10
REP: Driver resume for RayJob via a durable progress ledger

jhasm:
Follow-on work item 1 of [REP-64](https://github.com/ray-project/enhancements/blob/main/reps/2026-02-23-gcs-embedded-storage.md) — *"driver resilience during head pod restart"* — which named it as needed and out of its own scope. REP-64 is mine; this is the middle of the three pieces it framed (GCS persistence → driver resilience → application checkpointing).

📄 **[Rendered REP](https://github.com/jhasm/ray-enhancements/blob/rep-rayjob-driver-resume/reps/2026-09-10-rayjob-driver-resume.md)** — the diagrams render there.

### The problem

A RayJob's driver is a single point of failure with no durable state. With REP-64's embedded storage a head restart no longer kills the *cluster* — but it still kills the *job*, because the record of which units finished lived in the driver's heap. The job restarts from unit 1 and recomputes work whose results are already durable.

### The proposal

A disposable driver, a detached named coordinator actor that owns the work, and a progress ledger in `internal_kv` under `gcs_storage=rocksdb`. No Redis, no external database, no new operational dependency.

### What I'd like decided

1. **Placement.** The REP asks for this **in-tree** (a small `ray.util` module) even though the pattern needs **zero core patches on 2.57+**. §*Should this change be within `ray` or outside?* argues why feasibility-outside is not an argument for living outside. A staged fallback is offered if you would rather phase the risk.
2. **Three small core changes**, none a new subsystem: a documented `internal_kv` durability contract (plus per-namespace quota/TTL); a fix for [#55996](https://github.com/ray-project/ray/issues/55996); and the `_internal_kv_put` return-value inversion, which today returns `True` when the key *already existed*.
3. **Scope.** The REP is deliberately as detailed about what this does *not* help — NCCL training (Ray Train checkpointing already covers it, and this adds nothing), fine-grained work above a measured rate band, arbitrary driver code, exactly-once semantics — as about what it does.

### Evidence

Every claim was validated against **unmodified Ray 2.58.0 before the REP was written**: 32 claims, each with its refutation condition committed to git *before* the run that settled it, 35 runs, negative controls throughout, and 5 runs retained as void or partial with their diagnoses.

Three load-bearing claims were **refuted** and the design repaired — fencing stops writes but not deletes; the same race recurs across epochs; `session_name` is a storage-path identity, not a cluster identity. They are written up in §*What we got wrong* rather than quietly dropped, because they are what a reimplementation from the summary alone will get wrong too.

Full bundle: [ray-project/ray#66065](https://github.com/ray-project/ray/pull/66065) (draft, not for merge) — claim ledger, pre-registered experiment cards, all runs, and a stdlib crash-interleaving model checker that runs its full sweep in 51 s on one core.

### Stewardship

Proposing @edoakes as shepherd — shepherd of both REP-64 and REP-65, which this composes with. Reviewers of REP-64 @rueian, @andrewsykim, @MengjinYan.

### DIFF
--- a/reps/2026-09-10-rayjob-driver-resume.md
+++ b/reps/2026-09-10-rayjob-driver-resume.md
@@ -70,7 +70,7 @@ is by construction an ordinary client process.
 |---|---|
 | **Long-running sharded batch jobs** — a driver enumerating shards; resume skips the completed ones | **Distributed training (NCCL).** The collective group breaks on head failure and does not reconnect; training resumes from its application checkpoint regardless. Ray Train checkpointing covers this and we add nothing to it |
 | **Multi-stage orchestration pipelines**, where each stage is expensive and externally durable | **Fine-grained, high-throughput work.** Measured, not assumed: `N=10³` admits no usable configuration above ~100 units/s; `N=10⁶` tolerates 10⁴/s. Millisecond units should not use this |
-| **Spot / preemptible fleets**, where driver loss is routine rather than exceptional | **Arbitrary, uncooperative driver code.** A hard non-goal. This does not checkpoint a Python process; work must be restructured into re-enumerable idempotent units |
+| **Spot / preemptible fleets**, where driver loss is routine rather than exceptional | **Arbitrary, uncooperative driver code.** A hard non-goal. This does not checkpoint a Python process; work must be restructured into re-enumerable idempotent units — see the [appendix](#appendix-writing-a-driver-that-can-resume) for what that means concretely |
 | | **Exactly-once semantics.** Not offered. The contract is at-least-once: a crash between execution and commit re-executes at most `W` units |
 
 **Environment and scale**, per the REP checklist:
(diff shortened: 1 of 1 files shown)


edoakes:
@jhasm do you have a specific workload in mind that this would be used for? We currently implement checkpoint/recovery logic in a few of the Ray libraries (Data, Train, Serve) following somewhat of a similar pattern, but I'm not sure that the proposal here would do much to help them or unify the implementations. Documenting the `internal_kv` persistence contract & fixing minor behaviors there sound like good independent changes to me FWIW.

jhasm:
Yes — multi-hour fan-out batch and multi-stage pipeline jobs run as RayJobs on KubeRay with GCS FT. Tens of thousands of independent shards whose outputs are already durable on object storage, on clusters where head restarts and submitter-pod eviction are routine (node maintenance, spot capacity) rather than exceptional. Today an eviction at shard 4,900 of 5,000 re-runs the entrypoint from line one; the information needed to avoid that is a few kilobytes.

On unification: agreed, and I should not have left that ambiguous — the REP does not unify Data/Train/Serve recovery and proposes no library change. Their recovery lives *inside* a library-owned execution loop and in every case assumes the driver is alive: Train resumes from an application checkpoint but a lost driver ends the run, Data recovers task/actor failures within one `Dataset` execution using a driver-resident plan, and Serve survives driver loss but has no notion of committed progress because deployments are not jobs.

The gap is the layer above all three: user-written orchestration code in a RayJob entrypoint — the `for shard in shards: ray.get(...)` loop, or the stage-by-stage pipeline that *calls* Data or Train. That progress lives in the driver's heap and nothing in Ray persists it. Serve's pattern (detached actor + `internal_kv`) is the closest existing analogue; this is essentially that pattern made available to job drivers, with the durability contract it depends on written down. If the libraries later want a common substrate the ledger is a plausible one, but that is not a claim I am making here.

Pushed 159c15d making both of those explicit in the REP rather than leaving them implied.

On the `internal_kv` pieces — I agree they are good independent changes and I would rather they land that way than be gated on this. To be concrete, all three stand alone: the durability/quota/TTL contract (existing durable users — Serve, Jobs, the dashboard — already depend on undocumented behaviour), the `DEADLINE_EXCEEDED` wedge in ray-project/ray#55996 (a hang bug regardless of this proposal), and the `_internal_kv_put(overwrite=False)` return inversion, which returns `True` when the key already existed and is a correctness trap for every compare-and-set caller. Happy to split those out as separate issues/PRs now and let the ledger pattern be judged separately on whether the orchestration-layer gap is worth an in-tree answer.

edoakes:
I would propose:

- Split out the independent changes into their own issues and/or PRs
- Then implement FT for your workload as a validation/test case for this pattern
- Based on the above, we can decide if/how to expose this as a common utility in-tree

jhasm:
That plan works for me — done on the first step, and it's the right order.

**1. Independent changes split out.** All three are now tracked on their own, none gated on this REP:

- ray-project/ray#66113 — document the `internal_kv` durability contract under `gcs_storage=rocksdb` (does an ack imply the write survives a GCS restart?), plus per-namespace quota and TTL. Both additive and defaulting to off, so no behaviour change.
- ray-project/ray#55996 — `DEADLINE_EXCEEDED` is not retried by the GCS client and wedges the caller. Pre-existing, filed earlier.
- ray-project/ray#66114 — `_internal_kv_put(..., overwrite=False)` returns `True` when the key *already existed*, i.e. `True` means "I did not write". Filed with a repro and a compatibility-preserving path: fix the docstring now, add an explicitly-named `_internal_kv_put_if_absent` returning "did I win", leave the existing return value alone.

**2. Validation on a real workload.** Agreed, and I'd rather this be the deciding input than anything in the document. The case I'll run it against is the fan-out batch job described above — the concrete question being whether a driver lost at shard ~4,900 of 5,000 resumes with zero redundant committed units, and whether the ledger's key count stays flat in `N`. I'll report what it does rather than what I expect it to do, including if the answer is that the pattern needs per-workload shaping and doesn't generalise.

**3. In-tree decision deferred to after (2).** Pushed a933881, which adds an *Agreed sequencing* section and reframes the in-tree module (tier 3) as conditional on the outcome of step 2 rather than an immediate ask. The previous framing implied more than it should have.

One thing I'd keep visible during the deferral: the out-of-tree window is exactly when the subtle parts get copied incorrectly — per-incarnation instance ids that aren't actor-derived, epoch fencing, and the read-back after a compare-and-set that #66114 currently makes necessary. Those are written down in the non-optional-requirements appendix so the copies at least have something to check against.

Happy to keep this PR open as the record for (2), or close it and reopen with results if you'd rather not carry an open REP through the validation period — your call.

jhasm:
> @jhasm do you have a specific workload in mind that this would be used for? We currently implement checkpoint/recovery logic in a few of the Ray libraries (Data, Train, Serve) following somewhat of a similar pattern, but I'm not sure that the proposal here would do much to help them or unify the implementations. Documenting the `internal_kv` persistence contract & fixing minor behaviors there sound like good independent changes to me FWIW.

@edoakes  makes sense. I agree checkpointing is a good solution for resuming work in general, but the challenge with application-level checkpointing is the adoption, because it requires work from the application owners to invoke checkpointing and in some cases pay the latency cost. 
The proposal in this REP is kind of like system level checkpointing for the driver in collaboration with the driver program owner (deterministic order and idempotent units). But I also agree this has limited impact because it is not fine grained enough for training, data processing, etc. workloads (driver level, instead of checkpointing inside the workload).

I am curious if the team has any thoughts or plans around general system-level checkpointing. We have internally been thinking about such a solution as well, due to friction in adoption of application-level checkpointing, but do not have anything concrete yet.

